PR #11, opened 2018-09-13
Turnovers from RU Localization

dkatson:
Hi there,
As we agreed on the Pilot project, I've converted 3 Turnovers from Ru localization into extension "Turnovers Pack". This Extension could be installed on any localization (but translations are needed). 
I found this functionality very useful for Accountants. That's why started from that.


JesperSchulz:
That is great! Thanks a lot! I'm looking forward to dive into it. Two questions though:
1) Which code do I need to remove from the RU BaseApp? Is it identical?
2) Does this code not come with any tests? Most likely we have tests in the RU test suite, which also need to be extracted.
Thanks again. Now I'm curious to see the required workload on our side :-)

dkatson:
Hi @JesperSchulz . Thanks for the feedback.
1. 

Original Object|Destination Object|Was Changed
------------ | -------------|------------ 
page 12405|page 69595|Yes
page 12406|page 69596|Yes
page 12407|page 69597|Yes

These objects where changed because we can't extend tablerelation property with extention. 
In RU localization in these objects user can set filter on Source Type and Source No. To reach that in RU localization flow fields in table 15 (and some others) where added with filtering by Source Type
image

But in others localizations there are no such filters
image

So my logic was to create new Flow Fields in Tables 15, 18 and 23 - clones of original, but with source type filter. This resulted in 3 extension tables objects. And I modified code in Turnover Pages to connect to my new fields, not original.

This gives good chances to use this extension in all localizations, including RU.

2. I did not created any tests, cos there are no processing business logic, only view and filter data.

JesperSchulz:
Thanks for the detailed explanation! Does that mean, that I can delete PAG12405-7 from the Russian localization, once the extension is checked in? The extension would replace the functionality? This is a brilliant pilot, as the complexity seems to be rather limited. But I'd like to make sure I don't remove any functionality by incident ;-) I hope I get time to process your PR next week!

dkatson:
Yes, absolutely correct. New pages replace PAG12405-7 from the Russian localization with the same functionality. 
I took code from the NAV2018 image. So maybe something could be changed by Microsoft in the original pages in the nearest future, before official release but I don't think so :)
I tried to import BCSandbox with RU localization, to get the latest code, but got an error. Freddy helped me with that, so i will take the latest build and check if we have everything here.

JesperSchulz:
No rush :-) I need to look into the process on how we get this code into our repository. As this is a pilot, we have no precedence, so there's a bit of work in getting this off the ground. This may take a while, but we're working on it! Thanks a lot for your contribution!

AndreiPanko:
@dkatson 
Hi Dmitry, 
Thank you for your effort at extracting these objects into extensions.
I have to reject  your pull request as suggested functionality cannot replace RU features:
* We can not replace RU pages 12405, 12406 and 12407 with this extension, because this extension doesn't support other features in RU localization, such as Agreements and Correspondence.
* We can not add this extension without removing RU version, because users will be confused with duplicate functionality.


But I want to confirm that you are absolutely right that these objects don't belong to RU layer but to W1. We are aware about request to move this feature to W1. If you want as to prioritize this request, please submit your suggestion at https://experience.dynamics.com/ideas/list/?forum=e288ef32-82ed-e611-8101-5065f38b21f1 


Thanks,
Andrei Panko
Program manager

dkatson:
Hi Andrew!
Thanks for your feedback! 

I know that. And thought that Agreements and Correspondence will/should be extracted in separate extensions, that will be depended on Turnovers extension.

So
- Turnovers: W1
- Agreements: RU , depended on Turnovers
- Correspondence: RU, dependent on Turnovers and Agreements.

I think, this is possible. And this is the way we need to move further. 

What do you think?

Dmitry

AndreiPanko:
@dkatson 
We do consider componentization via extensions and we are working on the right level of details to make sure it is manageable.

PR #909, opened 2019-01-22
Update COD1871.C5LedTransMigrator.al

Hornsyld:
Adds the description from the C5-entry to the Gen. Journal Line instead of using the Account No.'s name description since it's validated in.


Hornsyld:
I dont like this change.. We don't want to description being validated from Account Description!
All the Ledgers come from another system and have been given a description from that system.. So if you validate description from account description you will loose all the description that have been typed into the old system.

Hornsyld:
Can you change 
            if DescriptionText <> '' then
to 
            if DocumentNo <> 'C5MIGRATE' then
Thanks.

GiorgosPa:
Hi @Hornsyld,
 
It is the same, empty description is only passed for the aggregated transaction where DocumentNo is 'C5MIGRATE' for individual transactions  C5LedTrans.Txt is passed as description.

Thank you for your suggestions.

PR #1247, opened 2019-02-14
Added events to Invoice and Credit Memo on insert line and before output

RasmusTidselbak:
It would be great, if we can manipulate the file, during creation. As well as being able to send the generated files to a webservice instead of just downloading it.

### DIFF
--- a/OIOUBL/src/Management/COD13646.Management.al
+++ b/OIOUBL/src/Management/COD13646.Management.al
@@ -20,7 +20,7 @@ codeunit 13646 "OIOUBL-Management"
 
         OIOUBLFileEvents.FileCreated(SourceFile);
 
-        if FileManagement.CanRunDotNetOnClient() then begin
+        if not ExportFileFromEvent(SourceFile) and FileManagement.CanRunDotNetOnClient() then begin
             FilePath := FileManagement.DownloadTempFile(SourceFile);
             FileManagement.CopyClientFile(FilePath, STRSUBSTNO('%1\%2', FolderPath, FileName), true);
         end else
@@ -54,5 +54,20 @@ codeunit 13646 "OIOUBL-Management"
           (DocumentSendingProfile."Electronic Format" = 'OIOUBL') OR
           (DocumentSendingProfile."E-Mail Format" = 'OIOUBL') OR
           (DocumentSendingProfile."Disk Format" = 'OIOUBL'));
+    end;
+
+    local procedure ExportFileFromEvent(SourceFile: Text[1024]) IsExported: Boolean;
+    var
+        OutputBlob: Record TempBlob temporary;
+        FileManagement: Codeunit "File Management";
+    begin
+        FileManagement.BLOBImportFromServerFile(OutputBlob, SourceFile);
+        OnBeforeExportFile(OutputBlob, IsExported);
+    end;
+
+
+    [IntegrationEvent(false, false)]
+    local procedure OnBeforeExportFile(var OutputBlob: Record TempBlob; var IsExported: Boolean);
+    begin
     end;
 }
--- a/OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
+++ b/OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
@@ -371,19 +371,19 @@ codeunit 13637 "OIOUBL-Export Sales Cr. Memo"
 
 
 
-    [BusinessEvent(false)]
+    [IntegrationEvent(false, false)]
     local procedure OnBeforeInsertCrMemoLine(var SalesCrMemoLine: Record "Sales Cr.Memo Line"; var XMLCurrNode: XmlElement; var IsHandled: Boolean);
     begin
     end;
 
 
 
-    [BusinessEvent(false)]
+    [IntegrationEvent(false, false)]
     local procedure OnAfterInsertCrMemoLine(var SalesCrMemoLine: Record "Sales Cr.Memo Line"; var XMLCurrNode: XmlElement);
     begin
     end;
 
-    [BusinessEvent(false)]
+    [IntegrationEvent(false, false)]
     local procedure OnBeforeExportFile(var OutputBlob: Record TempBlob; var IsExported: Boolean);
     begin
     end;
--- a/OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
+++ b/OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
@@ -176,7 +176,6 @@ codeunit 13637 "OIOUBL-Export Sales Cr. Memo"
         TotalAmount: Decimal;
         TotalInvDiscountAmount: Decimal;
         TotalTaxAmount: Decimal;
-        IsHandled: Boolean;
         IsExported: Boolean;
         OutputFile: File;
         FileOutstream: Outstream;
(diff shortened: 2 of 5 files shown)


RasmusTidselbak:
I'm missing a not

OIOUBL/src/SalesInvoice/COD13636.ExportSalesInvoice.al
```diff
@@ -434,4 +448,20 @@ codeunit 13636 "OIOUBL-Export Sales Invoice"
         end;
     end;
 
+    [BusinessEvent(false)]
```
MarcHansenMicrosoft: Please use integration events

OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
```diff
@@ -353,4 +369,23 @@ codeunit 13637 "OIOUBL-Export Sales Cr. Memo"
         end;
     end;
 
+
+
+    [BusinessEvent(false)]
```
MarcHansenMicrosoft: Please use integration event

MarcHansenMicrosoft:
Why is it you want to skip lines?
Will this give a problem with totals and more?
I think we should mirror to solution on all types.

Please also add test at the end of the review

OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
```diff
@@ -305,16 +308,29 @@ codeunit 13637 "OIOUBL-Export Sales Cr. Memo"
 
         // CreditMemo->CreditMemoLine
         repeat
-            SalesCrMemoLine.TESTFIELD(Description);
+            OnBeforeInsertCrMemoLine(SalesCrMemoLine, XMLCurrNode, IsHandled);
```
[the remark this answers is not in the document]
RasmusTidselbak: Can you elaborate?

RasmusTidselbak:
> Why is it you want to skip lines?
> Will this give a problem with totals and more?

I'm not interested in skipping lines completely, I would like to let extensions generate lines on their own, with other XML elements than the implemented. Therefore I've included the IsHandled, like described in the [Handled-pattern](https://community.dynamics.com/nav/b/navmarkbrummel/archive/2015/11/25/the-handled-pattern).

OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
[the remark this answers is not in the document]
MarcHansenMicrosoft: I would move:
            OnBeforeInsertCrMemoLine(SalesCrMemoLine, XMLCurrNode, IsHandled);

            if not IsHandled then begin
                SalesCrMemoLine.TESTFIELD(Description);
                ExcludeVAT(SalesCrMemoLine, SalesCrMemoHeader."Prices Including VAT");
                InsertCrMemoLine(XMLCurrNode, SalesCrMemoHeader, SalesCrMemoLine, CurrencyCode);
            end;

            OnAfterInsertCrMemoLine(SalesCrMemoLine, XMLCurrNode);

into a function called AddCrMemoLineToDocument and rename the events to match.

MarcHansenMicrosoft:
> Why is it you want to skip lines?
> Will this give a problem with totals and more?
> 
> I'm not interested in skipping lines completely, I would like to let extensions generate lines on their own, with other XML elements than the implemented. Therefore I've included the IsHandled, like described in the  @Handled-pattern.
@RasmusTidselbak 
OK so it is more a substitution.
Thanks

OIOUBL/src/SalesInvoice/COD13636.ExportSalesInvoice.al
RasmusTidselbak: Done

RasmusTidselbak:
I just realised that the app will fail, if the file is not exported.
I'm working on a solution.

RasmusTidselbak:
> Please also add test at the end of the review
@MarcHansenMicrosoft 
How do you wan't the tests to be created. Since I can't hook into events when both the test and the app is in the same project.

MarcHansenMicrosoft:
It should not be a problem adding event subscribers in the same project but remember to make then manual binding.
If this is not working please tell me we might log a bug on it then.

OIOUBL/src/Management/COD13646.Management.al
```diff
@@ -55,4 +55,19 @@ codeunit 13646 "OIOUBL-Management"
           (DocumentSendingProfile."E-Mail Format" = 'OIOUBL') OR
           (DocumentSendingProfile."Disk Format" = 'OIOUBL'));
     end;
+
```
MarcHansenMicrosoft: Use the existing OIOUBLFileEvents.FileCreated, no need to create a new event.

OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
```diff
@@ -303,14 +307,15 @@ codeunit 13637 "OIOUBL-Export Sales Cr. Memo"
             until SalesCrMemoLine2.NEXT() = 0;
         OIOUBLXMLGenerator.InsertLegalMonetaryTotal(XMLCurrNode, TaxableAmount, TaxAmount, TotalAmount, TotalInvDiscountAmount, CurrencyCode);
 
+        OnAfterAddCrMemoHeaderToDocument(XMLCurrNode, SalesCrMemoHeader);
```
MarcHansenMicrosoft: This is a odd location is it needed?

OIOUBL/src/SalesInvoice/COD13636.ExportSalesInvoice.al
```diff
@@ -374,15 +377,13 @@ codeunit 13636 "OIOUBL-Export Sales Invoice"
             until SalesInvLine2.NEXT() = 0;
 
         OIOUBLXMLGenerator.InsertLegalMonetaryTotal(XMLCurrNode, TaxableAmount, TaxAmount, TotalAmount, TotalInvDiscountAmount, CurrencyCode);
-
+        OnAfterAddInvoiceHeaderToDocument(XMLCurrNode, SalesInvoiceHeader);
```
MarcHansenMicrosoft: Same commen tas credit memo

OIOUBL/src/Management/COD13646.Management.al
RasmusTidselbak: Yes, that could be a solution, but I wan't to be able to send the file directly from the event and only from the event.
By using the OIOUBLFileEvents.FileCreated, the user would still be asked to download the file or the file would be stored on the disk.  

Fx. if I were to create an extension which send the file directly to a service, the user doesn't want the file exported to neither the server nor download the file, since it has already been sent by the system.

OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
RasmusTidselbak: In case I wan't to add extra elements or attributes to the header of the document.

OIOUBL/src/SalesInvoice/COD13636.ExportSalesInvoice.al
RasmusTidselbak: Same comment: "In case I wan't to add extra elements or attributes to the header of the document."

OIOUBL/src/Management/COD13646.Management.al
MarcHansenMicrosoft: OK but then refacture the event with a IsHandled

MarcHansenMicrosoft:
What happend to the test?

OIOUBL/src/Management/COD13646.Management.al
RasmusTidselbak: Will do

RasmusTidselbak:
> What happend to the test?

I'm working on it.
Remember that I'm working on this in my spare time.

OIOUBL/src/SalesInvoice/COD13636.ExportSalesInvoice.al
MarcHansenMicrosoft: OK

OIOUBL/src/SalesCreditMemo/COD13637.ExportSalesCreditMemo.al
MarcHansenMicrosoft: ok

MarcHansenMicrosoft:
Sorry did not wan't to rush on you. I was just not sure if you still had issues with adding test subscribers.

clausmacali:
Hi Marc When you are it making changes
OIOUBL Extension Exports the XML to Server File and after that based on the settings it either downloads / sends the XML File to Document Exchange Service.
 
Creating a File on the Server is not allowed in the Cloud Version, How are you planning to enhance this extension (Replace File to Streams)? As we are working on an Integration we cannot use files, we are replacing them with Streams. So we need blob fields instead of File Path in few functions, so that it can use the Standard Flow. Can you please let us know your plans on this? Thanks

MarcHansenMicrosoft:
No current plans to change OIOUBL file handling.
It works in it's current version as the extension is internal.

It could be that @RasmusTidselbak would change the eventing to send a blob or stream instead.

OIOUBL/src/Management/COD13646.Management.al
```diff
@@ -20,7 +20,7 @@ codeunit 13646 "OIOUBL-Management"
 
         OIOUBLFileEvents.FileCreated(SourceFile);
 
-        if FileManagement.CanRunDotNetOnClient() then begin
+        if not ExportFileFromEvent(SourceFile) and FileManagement.CanRunDotNetOnClient() then begin
```
RasmusTidselbak: @MarcHansenMicrosoft Can you help me out here, I'm out of ideas on how I can test that the file is not downloaded, when I return IsExported=true from the ExportFileFromEvent function.

MarcHansenMicrosoft: @RasmusTidselbak sorry for the slow response time.

First I do not think you should test for what the program should not do.

And I also think you should restructure the code to make it more clear that first you call the publisher to handle the save if it, then you check if it can save using com and last you just upload to the client.

MarcHansenMicrosoft:
The pull request has been stalled and still requires some work.
@RasmusTidselbak if you want to and have the time you are welcome to reopen and continue.

PR #3014, opened 2019-06-27
Unable to build without this update

gunnargestsson:
Fix the path to the Javascript file


JesperSchulz:
Very nice catch. I did see the js folder but decided not to push it, as it seems to be in the wrong place. I'll make sure we clean that up! Thanks a lot!

JesperSchulz:
Now it compiles, but it doesn't seem to work right. I'll have someone look at it. At least, you should be unblocked for now!

PR #4189, opened 2019-09-11
Marc amc banking365 fundamentals v1

AMCJN:
For adding more test to AMC


AddOns/AMCBanking365Fundamentals/app/.vscode/rad.json
```diff
@@ -0,0 +1 @@
+﻿{"Added":[],"Modified":[],"Removed":[]}
```
darjoo: Any reason we need the .vscode folder and files?

PR #4252, opened 2019-09-13
Update BREAKINGCHANGES.md

ChristopherDedominici:
Add error and solution for 'User Plan' table and 'Plan' table

### DIFF
--- a/BREAKINGCHANGES.md
+++ b/BREAKINGCHANGES.md
@@ -40,6 +40,17 @@ We’re working hard to make this a comprehensive list, but there’s always a c
  
 ---
 
+## Azure AD Plan Module
+**Error**: _'User Plan' is inaccessible due to its protection level_     
+
+**Solution**: Access the table through the facade APIs, `codeunit 9016 "Azure AD Plan"`, or using the `query 774 "Users in Plans"`.
+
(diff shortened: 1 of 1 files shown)


BREAKINGCHANGES.md
```diff
@@ -40,6 +40,17 @@ We’re working hard to make this a comprehensive list, but there’s always a c
...
 ---
 
+## Azure AD Plan Module
+**Error**: _'User Plan' is inaccessible due to its protection level_     
+
+**Solution**: Access the table through the facade APIs, `codeunit 9016 "Azure AD Plan"`, or using the `query 774 "Users in Plans"`.
+
+**Error**: _'Plan' is inaccessible due to its protection level_     
+
+**Solution**: Access the table through the facade APIs, `codeunit 9016 "Azure AD Plan"`, or using the `query 775 Plan`.
```
darjoo: I think you might want to specific for OnPrem customers, they can access the table through facade APIs and for SaaS use the query. Or just make them use the query. I don't know what's the best solution.

jwikman: Or maybe add something about that for CU1, "Azure AD Plan" will be available for SaaS ;-)

PR #4420, opened 2019-10-02
BLOB Storage Module Extended by From an To FieldRef

Walwode:
incl. Tests


JesperSchulz:
Code being reviewed and processed internally.

PR #4541, opened 2019-10-15
Added Reset function to the assisted setup

AndreasGloeckner:
Discussed this feature in Yammer with Soumya Dutta:

https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=314088954175488&search_origin=global&scoring=linear1Y-prankie-group-private-higher&match=any-exact&search_sort=relevance&page=1&search=assisted%20setup.

We need this function in test scenarios to make sure that the wizard's complete status is reseted before opening the assisted setup.
If the assisted setup is not reseted a confirm will appear. But if its not completed, the confirm will not appear.

Because there is no "optional confirmhandler" the reset function is the easiest way to solve this behavior.

### DIFF
--- a/Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
+++ b/Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
@@ -93,9 +93,9 @@ codeunit 3725 "Assisted Setup"
     /// <param name="ExtensionID">The app ID of the extension to which the setup belongs.</param>
     /// <param name="PageID">The ID of the page to open when the user clicks the setup.</param>
     /// <returns>True if the assisted setup is uncomplete now.</returns>
-    procedure Reset(ExtensionID: Guid; PageID: Integer)
+    procedure Reset(PageID: Integer)
     begin
-        AssistedSetupImpl.Reset(ExtensionID, PageID);
+        AssistedSetupImpl.Reset(PageID);
     end;
 
     /// <summary>Issues the call to execute the setup.</summary>
--- a/Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
+++ b/Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
@@ -89,6 +89,15 @@ codeunit 3725 "Assisted Setup"
         AssistedSetupImpl.Complete(ExtensionID, PageID);
     end;
 
+    /// <summary>Sets the status of the assisted setup to uncomplete.</summary>
+    /// <param name="ExtensionID">The app ID of the extension to which the setup belongs.</param>
+    /// <param name="PageID">The ID of the page to open when the user clicks the setup.</param>
+    /// <returns>True if the assisted setup is uncomplete now.</returns>
+    procedure Reset(ExtensionID: Guid; PageID: Integer)
+    begin
+        AssistedSetupImpl.Reset(ExtensionID, PageID);
+    end;
+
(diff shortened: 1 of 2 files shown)


Modules/System/Assisted Setup/src/AssistedSetupImpl.Codeunit.al
```diff
@@ -99,6 +99,18 @@ codeunit 1813 "Assisted Setup Impl."
         AssistedSetup.Modify(true);
     end;
 
+    procedure Reset(ExtensionID: Guid; PageID: Integer)
```
JesperSchulz: What is the ExtensionID used for here? I fail to find the spot where that parameter used. Doesn't seem to be part of the Primary Key on the Assisted Setup table. Maybe we can omit it?

JesperSchulz: Just spoke to Dutta. I understand that we had to keep this parameter even though it is not needed, as removing it would be a breaking change. I've created an internal bug to overload the methods without the unused parameter and to deprecate the ones which have the ExtensionID. In this case, I would get rid of the parameter right away. There is no need to introduce a new function which we deprecate right away.

AndreasGloeckner: Do you then introduce a function with Reset(PageID:Integer) or will you even omit that function?

Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
```diff
@@ -89,6 +89,15 @@ codeunit 3725 "Assisted Setup"
         AssistedSetupImpl.Complete(ExtensionID, PageID);
     end;
 
+    /// <summary>Sets the status of the assisted setup to uncomplete.</summary>
+    /// <param name="ExtensionID">The app ID of the extension to which the setup belongs.</param>
+    /// <param name="PageID">The ID of the page to open when the user clicks the setup.</param>
+    /// <returns>True if the assisted setup is uncomplete now.</returns>
```
DuttaSoumya: The function does not seem like it it returning anything. Please remove comment.

Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
```diff
@@ -89,6 +89,15 @@ codeunit 3725 "Assisted Setup"
         AssistedSetupImpl.Complete(ExtensionID, PageID);
     end;
 
+    /// <summary>Sets the status of the assisted setup to uncomplete.</summary>
```
DuttaSoumya: Minor typo: Change "uncomplete" to "incomplete" please.

Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
```diff
@@ -89,6 +89,15 @@ codeunit 3725 "Assisted Setup"
...
+    /// <param name="ExtensionID">The app ID of the extension to which the setup belongs.</param>
```
DuttaSoumya: Remove the comment about ExtensionID

AndreasGloeckner: done

bc-ghost:

bc-ghost:

bc-ghost:

bc-ghost:

PR #5252, opened 2019-12-11
Add Parent Language ID

markbrummel:
Added a function to get Parent Language ID from Windows Culture Info

### DIFF
--- a/Modules/System/Language/src/Language.Codeunit.al
+++ b/Modules/System/Language/src/Language.Codeunit.al
@@ -153,6 +153,17 @@ codeunit 43 Language
     end;
 
     /// <summary>
+    /// Returns an alternative language based on Windows Culture Info.
+    /// </summary>
+    /// <param name="LanguageId">Exit parameter that holds the chosen language ID.</param>
+    procedure GetParentLanguageId(LanguageId: Integer): Integer
+    var
+        LanguageImpl: Codeunit "Language Impl.";
+    begin
+        exit(LanguageImpl.GetParentLanguageId((LanguageId)));
+    end;
+    
+    /// <summary>
     /// Integration event, emitted from <see cref="GetUserLanguageCode"/>.
     /// Subscribe to this event to change the default behavior by changing the provided parameter(s).
     /// </summary>
--- a/Modules/System/Language/src/Language.Codeunit.al
+++ b/Modules/System/Language/src/Language.Codeunit.al
@@ -153,7 +153,7 @@ codeunit 43 Language
     end;
 
     /// <summary>
-    /// Returns an alternative language based on Windows Culture Info.
+    /// Returns the parent LanguageId based on Windows Culture Info.
     /// </summary>
     /// <param name="LanguageId">Exit parameter that holds the chosen language ID.</param>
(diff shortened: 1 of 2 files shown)


Modules/System/Language/src/Language.Codeunit.al
```diff
@@ -152,6 +152,17 @@ codeunit 43 Language
         LanguageImpl.LookupWindowsLanguageId(LanguageId);
     end;
 
+    /// <summary>
+    /// Returns an alternative language based on Windows Culture Info.
```
JesperSchulz: It's not an alternative language, it's the culture identifier of the parent language, isn't it?

Modules/System/Language/src/LanguageImpl.Codeunit.al
```diff
@@ -149,6 +149,13 @@ codeunit 54 "Language Impl."
         if PAGE.RunModal(PAGE::"Windows Languages", WindowsLanguage) = ACTION::LookupOK then
             LanguageId := WindowsLanguage."Language ID";
     end;
+        
+    procedure GetParentLanguageId(LanguageId: Integer): Integer
+    var
+        CultureInfo: DotNet CultureInfo;
+    begin
+        exit(CultureInfo.CultureInfo(LanguageId).Parent.LCID);
```
JesperSchulz: Do we need a null check here? Can the parent be null, leading to an exception when trying to get the LCID?

GiorgosPa: It can also be a TryFunction and let the consumer decide

Modules/System/Language/src/Language.Codeunit.al
markbrummel: This is what is on MSDN. Maybe we should copy/paste.
Gets the CultureInfo that represents the parent culture of the current CultureInfo.

Modules/System/Language/src/LanguageImpl.Codeunit.al
markbrummel: I like the idea of the TryFunction. DotNet will throw an error but not one that is easy to read.

If I add a TryFunction, does that complicate the pull request because translations are required for the error message?

markbrummel: Alternatively we can return the orriginal integer when the tryfunction fails.

Modules/System/Language/src/Language.Codeunit.al
markbrummel: Changed

Modules/System/Language/src/LanguageImpl.Codeunit.al
markbrummel: Implemented TryFunction. If there is no parent, it returns itsself.

JesperSchulz:
I made two more minor changes:
- added dependency to DotNet Aliases, since you are now using _DotNet CultureInfo_
- added () to end of method calls Parent and LCID, as those are cop rules we have to follow. Other than that, the PR goes in as you submitted. Thank you for your contribution.

bc-ghost:

bc-ghost:

bc-ghost:

PR #5299, opened 2019-12-16
Added procedure to remove Assisted Setup

Avanti05:
Procedures to remove deprecated Assisted Setups. 

If the Pages the Assisted Setup record refers to are removed, running the assisted setup will result in a runtime error.

### DIFF
--- a/Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
+++ b/Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
@@ -112,9 +112,9 @@ codeunit 3725 "Assisted Setup"
     /// <summary>Removes an Assisted Setup from a given extension so it will no longer be shown in the list.</summary>
     /// <param name="ExtensionID">The app ID of the extension to which the setup belongs.</param>
     /// <param name="PageID">The ID of the page to be removed.</param>
-    procedure Remove(ExtensionID: Guid; PageID: Integer)
+    procedure Remove(PageID: Integer)
     begin
-        AssistedSetupImpl.Remove(ExtensionID, PageID);
+        AssistedSetupImpl.Remove(PageID);
     end;
 
 
--- a/Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
+++ b/Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
@@ -109,6 +109,15 @@ codeunit 3725 "Assisted Setup"
         AssistedSetupImpl.Open(AssistedSetupGroup);
     end;
 
+    /// <summary>Removes an Assisted Setup from a given extension so it will no longer be shown in the list.</summary>
+    /// <param name="ExtensionID">The app ID of the extension to which the setup belongs.</param>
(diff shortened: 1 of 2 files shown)


JesperSchulz:
Just for info: We've worked on the underlying root cause, which is that invoking Page.RUN on a page which doesn't exists resulted in a runtime error. That behavior will change soon.

Modules/System/Assisted Setup/src/AssistedSetup.Codeunit.al
```diff
@@ -109,6 +109,15 @@ codeunit 3725 "Assisted Setup"
         AssistedSetupImpl.Open(AssistedSetupGroup);
     end;
 
+    /// <summary>Removes an Assisted Setup from a given extension so it will no longer be shown in the list.</summary>
+    /// <param name="ExtensionID">The app ID of the extension to which the setup belongs.</param>
+    /// <param name="PageID">The ID of the page to be removed.</param>
+    procedure Remove(ExtensionID: Guid; PageID: Integer)
```
mazhelez: `ExtensionID` parameter isn't used.

Avanti05: No, I used the Complete() and Run() procedures as template. I considered doing a check on the Assisted Setup record to ensure that you only delete your own.

Avanti05: I have updated the pull request and omitted the ExtensionId parameter.

mazhelez: We have removed this parameter from the mentioned functioned as they are not used :) (The changes are not yet uptaken in this repo).
So I advise you to do the same.

Avanti05: I will update :) Thank you for the notice.

darjoo:
Hi @Avanti05,
My apologies for the slow progress over the holiday period.
I was recently informed that Assisted Setup was updated with the ability to remove Assisted Setups when a user to tries to run it. Please take a look at the [code](https://github.com/microsoft/ALAppExtensions/blob/c955a27246d30ee61496155b53f0d6c9dc98e566/Modules/System/Assisted%20Setup/src/AssistedSetupImpl.Codeunit.al#L152).

As I was implementing the tests for this procedure, I stumbled upon the fact that this procedure might not be the best. From how I see it, the partner would require another extension which is created specifically to remove the Assisted Setup and I do not think that is the right decision to make. Is this how you see it being used as well, or in another manner?

-Darrick

Avanti05:
Hi Darrick,

Actually we have been refacotring and restructuring some of our assisted setups, which resulted in some of them being joined meaning some have been deprecated. In that regard we would like to be able to remove the deprecated Assisted Setup when updating, in order not to confuse the the users concerning which assisted setup they need to run. 
We have considered adding a procedure to reset an Assisted Setup, in case new options are added which require user attention. This will of course be adde in a seperate pull request. 

Best regards 
Martin

bc-ghost:

PR #5598, opened 2020-01-14
Invalid enum for test marker and obsolete authentication switch

sirhc101:
Implemented two changes in ELSTER / ERiC Add-On:

1. At least in the current version of ELSTER/ERiC it's necessary to send all data in USTVA format (Sales VAT Adv. Notifcation) authenticated. send-NoSig is deprecated and no longer supported for this data format.
2. For sending productive data the created XML is currently creating a XML Node called "Testmerker" with value 000000000. Referring to schema documentation for ELSTER version 11 (V11_V2.23) this is no valid enumeration. Based on the schema documentation this Node is optional and only for testing purpose.

### DIFF
--- a/AddOns/Elster/app/src/Pages/PAG11016.SalesVATAdvNotifCard.al
+++ b/AddOns/Elster/app/src/Pages/PAG11016.SalesVATAdvNotifCard.al
@@ -116,11 +116,6 @@ page 11016 "Sales VAT Adv. Notif. Card"
                     ApplicationArea = Basic, Suite;
                     ToolTip = 'Specifies the VAT statement template name used to calculate the amounts and assign them to the key figures required by the tax authorities.';
                 }
-                field("Use Authentication"; "Use Authentication")
-                {
-                    ApplicationArea = Basic, Suite;
-                    ToolTip = 'Specifies whether to submit signed data to the tax authorities.';
-                }
             }
             group(Communication)
             {
--- a/AddOns/Elster/app/src/Reports/REP11016.CreateXMLFileVATAdvNotif.al
+++ b/AddOns/Elster/app/src/Reports/REP11016.CreateXMLFileVATAdvNotif.al
@@ -273,12 +273,8 @@ report 11016 "Create XML-File VAT Adv.Notif."
             exit;
         if not AddElement(XmlRootElem, XmlElemNew, 'DatenArt', 'UStVA', XmlNameSpace) then
             exit;
-        if UseAuthentication then begin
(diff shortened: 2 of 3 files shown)


JesperSchulz:
Hi Christoph! As soon as you've signed the CLA, we'll start processing your PR :-)

AddOns/Elster/app/src/Pages/PAG11016.SalesVATAdvNotifCard.al
```diff
@@ -116,11 +116,6 @@ page 11016 "Sales VAT Adv. Notif. Card"
                     ApplicationArea = Basic, Suite;
                     ToolTip = 'Specifies the VAT statement template name used to calculate the amounts and assign them to the key figures required by the tax authorities.';
                 }
-                field("Use Authentication"; "Use Authentication")
```
JesperSchulz: Breaking Change. Field should get deprecated.

sirhc101: Sorry for delay. 
Do you mean that I should not have removed the "Use Authentication" field from the page?
Well I did this because of this compiler error message:
> Field 'Use Authentication' is removed. Reason: Authentication is required for electronic transmission.AL(AL0433)

I initially tried to keep it at the page and simply add an ObsoleteReason.
However, I changed the ObsoleteState to Pending and modifed the ObsoleteReason. Further I removed the Use Authentication field from Tests. I think more changes are not necessary referring to my changes.

sorenfriisalexandersen:
Christoph, could you send a link to the specs?
We would like to have this validated by a couple of more partners. This is a general approach we take with Open Source. 
Thanks.

sirhc101:
@sorenfriisalexandersen 
Well, thats not so easy to be honest. You need a developer account at ELSTER website [see here](https://www.elster.de/elsterweb/infoseite/entwickler) to get the interface documentation. The documentation is large because of all the stuff which is supported by ELSTER ERiC, but here is a quote from the API reference documentation:
image

Sorry it's in german language. The documentation doesn't exist in englisch language.

JesperSchulz:
This PR is "on hold" until we've gotten confirmation that the changes are ok and necessary. The first part we're having a few issues getting a definite answer on, the latter part seems not to be the case (just yet). We'll pick it up again, once the details have been ironed out.

JesperSchulz:
Doing a New Year's clean-up here on GitHub. We were never able to get to the bottom of this PR. I will close the PR under the assumption, that these changes either aren't needed or have been addressed otherwise. If I am mistaken and this change still is needed after all, I apologize and would kindly request that you reopen this PR. I will then make another attempt to get it processed.

Issue #2043, opened 2019-04-09
Passwords encrypted in C/AL with Rfc2898DeriveBytes 


angela2389:
We are currently using Rfc2898DeriveBytes in C/AL to encrypt passwords/data (in combination with other encryption). In AL this functionality is missing, therefore causing that we cannot check if passwords are correct when users fill in the password to access data files. We cannot decrypt the passwords in C/AL either to convert them to a different type in AL. Users will therefore be unable to access encrypted data files. Thereby, we would like to continue using Rfc2898DeriveBytes to properly secure our files.

Can you please add the Rfc2898DeriveBytes functionality in a Codeunit we can access?

PredragMaricic:
Hi @angela2389,
Thanks for your proposal. Looks like a valid one, but it is also a new requirement. When it comes to such proposals, we also need to evaluate importance for all partners. We will do triage this week and inform you about the decision.

PredragMaricic:
Hi @angela2389,,

We created and opensourced Encryption Management module, so you might want to take a look and suggest the code to provide support for Rfc2898DeriveBytes. Then we can review it and add it to the module if possible.
https://github.com/microsoft/ALAppExtensions/tree/master/Modules/Cryptography%20Management

angela2389:
Hi @PredragMaricic ,

Is it possible to implement it like this?

```
   procedure FuncCreateHashWithRfc2898DeriveBytes(Password: Text; InputText: Text; DivideHashBy: Integer ) Hash: Text
    var
        ByteArray: DotNet Array; // System.Array
        Convert: DotNet Convert; // System.Convert
        Encoding: DotNet Encoding; // System.Text.Encoding
        Rfc2898DeriveBytes: DotNet Rfc2898DeriveBytes; //System.Security.Cryptography.Rfc2898DeriveBytes
        SHA1Managed: DotNet SHA1Managed; // System.Security.Cryptography.SHA1Managed
    begin
        if InputText = '' then
          exit;

        Rfc2898DeriveBytes := Rfc2898DeriveBytes.Rfc2898DeriveBytes(Password,Encoding.ASCII.GetBytes(InputText)); 
        SHA1Managed := SHA1Managed.SHA1Managed;
        ByteArray := SHA1Managed.ComputeHash(Rfc2898DeriveBytes.GetBytes(SHA1Managed.HashSize / DivideHashBy));              
        Hash := Convert.ToBase64String(ByteArray);    

        Clear(ByteArray);                              
[...]

PR #5985, opened 2020-02-07
Rfc2898derivebytes codeunit

angela2389:
As per ticket https://github.com/microsoft/ALAppExtensions/issues/2043 hereby the suggested code

### DIFF
--- a/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
+++ b/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
@@ -2,7 +2,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
-codeunit 50101 "Rfc2898DeriveBytes"
+codeunit 132576 "Rfc2898DeriveBytes"
 {
     procedure FuncCreateHashWithRfc2898DeriveBytes(Password: Text; InputText: Text; PsuedoRandomNumber: Integer; FillCharacter: Integer) Hash: Text
     var
--- a/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
+++ b/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
@@ -4,28 +4,19 @@
 // ------------------------------------------------------------------------------------------------
 codeunit 132576 "Rfc2898DeriveBytes"
 {
-    procedure FuncCreateHashWithRfc2898DeriveBytes(Password: Text; InputText: Text; PsuedoRandomNumber: Integer; FillCharacter: Integer) Hash: Text
+    Subtype = Test;
+
+    [Test]
+    procedure TestDeriveBytes()
     var
-        ByteArray: DotNet Array;
-        Convert: DotNet Convert;
-        Encoding: DotNet Encoding;
-        Rfc2898DeriveBytes: DotNet Rfc2898DeriveBytes;
-        SHA1Managed: DotNet SHA1Managed;
+        CodRfc2898DeriveBytes: Codeunit Rfc2898DeriveBytes;
+        EncryptedText: Text;
+        ExpectedEncryptedText: Label 'RgIsn9T5fqPK8bsXjzlmWqinRxg=';
     begin
-        if InputText = '' then
-            exit;
-
-        //Implement password-based key derivation functionality, PBKDF2, by using a pseudo-random number generator based on HMACSHA1.
-        Rfc2898DeriveBytes := Rfc2898DeriveBytes.Rfc2898DeriveBytes(Password, Encoding.ASCII.GetBytes(PadStr(InputText, FillCharacter)));
-
-        //Return a Base64 encoded string of the SHA1 hash of the first X bytes (X = PsuedoRandomNumber) returned from the generator.
-        SHA1Managed := SHA1Managed.SHA1Managed;
-        ByteArray := SHA1Managed.ComputeHash(Rfc2898DeriveBytes.GetBytes(PsuedoRandomNumber));
-        Hash := Convert.ToBase64String(ByteArray);
-
-        //Clear all used variables
-        Clear(ByteArray);
-        Clear(SHA1Managed);
-        Clear(Rfc2898DeriveBytes);
+        // Encrypt Text 
+        EncryptedText := CodRfc2898DeriveBytes.FuncCreateHashWithRfc2898DeriveBytes('Test', 'Test123', 23, 10);
+        // Verify Result 
+        IF (EncryptedText <> ExpectedEncryptedText) THEN
+            ERROR('Failed to encrypt text');
     end;
 }
--- a/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
+++ b/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
@@ -11,10 +11,10 @@ codeunit 132576 "Rfc2898DeriveBytes Test"
     var
         CodRfc2898DeriveBytes: Codeunit Rfc2898DeriveBytes;
         EncryptedText: Text;
-        ExpectedEncryptedText: Label 'RgIsn9T5fqPK8bsXjzlmWqinRxg=';
+        ExpectedEncryptedText: Label 'fzBagIx5+suJx7TQ/wLi98ZQHik=';
     begin
         // Encrypt Text 
-        EncryptedText := CodRfc2898DeriveBytes.HashRfc2898DeriveBytes('Test', 'Test123', 23, 10);
+        EncryptedText := CodRfc2898DeriveBytes.HashRfc2898DeriveBytes('Test', 'Test1234', 23);
         // Verify Result 
         IF (EncryptedText <> ExpectedEncryptedText) THEN
             ERROR('Failed to encrypt text');
--- a/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
+++ b/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
@@ -2,7 +2,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
-codeunit 132576 "Rfc2898DeriveBytes Test"
+
+codeunit 11111567 "Test Rfc2898DeriveBytes"
 {
     Subtype = Test;
 
@@ -11,12 +12,19 @@ codeunit 132576 "Rfc2898DeriveBytes Test"
     var
         CodRfc2898DeriveBytes: Codeunit Rfc2898DeriveBytes;
         EncryptedText: Text;
-        ExpectedEncryptedText: Label 'fzBagIx5+suJx7TQ/wLi98ZQHik=';
+        ExpectedEncryptedText: Text;
+        LibraryAssert: Codeunit "Library Assert";
     begin
-        // Encrypt Text 
+        // [GIVEN] With Encryption Key 
+        ExpectedEncryptedText := GetRfc2898DeriveBytesEncryptedText();
+        // [WHEN] Encrypt Text 
         EncryptedText := CodRfc2898DeriveBytes.HashRfc2898DeriveBytes('Test', 'Test1234', 23);
-        // Verify Result 
-        IF (EncryptedText <> ExpectedEncryptedText) THEN
-            ERROR('Failed to encrypt text');
+        // [THEN] Verify Result 
+        LibraryAssert.AreEqual(ExpectedEncryptedText, EncryptedText, 'Failed to encrypt text with Rfc2898DeriveBytes');
+    end;
+
+    local procedure GetRfc2898DeriveBytesEncryptedText(): Text
+    begin
+        exit('fzBagIx5+suJx7TQ/wLi98ZQHik=');
     end;
 }
--- a/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
+++ b/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
@@ -2,7 +2,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
-codeunit 132576 "Rfc2898DeriveBytes"
+codeunit 132576 "Rfc2898DeriveBytes Test"
 {
     Subtype = Test;
 
@@ -14,7 +14,7 @@ codeunit 132576 "Rfc2898DeriveBytes"
         ExpectedEncryptedText: Label 'RgIsn9T5fqPK8bsXjzlmWqinRxg=';
     begin
         // Encrypt Text 
-        EncryptedText := CodRfc2898DeriveBytes.FuncCreateHashWithRfc2898DeriveBytes('Test', 'Test123', 23, 10);
+        EncryptedText := CodRfc2898DeriveBytes.HashRfc2898DeriveBytes('Test', 'Test123', 23, 10);
         // Verify Result 
         IF (EncryptedText <> ExpectedEncryptedText) THEN
             ERROR('Failed to encrypt text');
--- a/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
+++ b/Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
@@ -18,7 +18,7 @@ codeunit 11111567 "Test Rfc2898DeriveBytes"
         // [GIVEN] With Encryption Key 
         ExpectedEncryptedText := GetRfc2898DeriveBytesEncryptedText();
         // [WHEN] Encrypt Text 
-        EncryptedText := CodRfc2898DeriveBytes.HashRfc2898DeriveBytes('Test', 'Test1234', 23);
+        EncryptedText := CodRfc2898DeriveBytes.HashRfc2898DeriveBytes('Test', 'Test1234', 23, 1);
         // [THEN] Verify Result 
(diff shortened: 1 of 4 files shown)


PaulIvan:
@angela2389 we are missing tests and documentation we cannot accept this change otherwise.
Please have a look at these tests for the Rijndael Encryption for inspiration :) 

https://github.com/microsoft/ALAppExtensions/blob/master/Modules/System%20Tests/Cryptography%20Management/src/RijndaelCryptographyTest.Codeunit.al

Also we need short function documentation have a look here : https://github.com/microsoft/ALAppExtensions/blob/master/Modules/System/Cryptography%20Management/README.md

angela2389:
Thanks @PaulIvan, I have added a test codeunit and included documentation now.

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,31 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+codeunit 50101 "Rfc2898DeriveBytes"
```
darjoo: The codeunit itself needs to be documented like in https://github.com/microsoft/ALAppExtensions/blob/c07cb0dc989fdd022a9dc18eb8db22a190b5fefb/Modules/System/Cryptography%20Management/src/RijndaelCryptography.Codeunit.al#L6-L8

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,31 @@
...
+{
+    procedure FuncCreateHashWithRfc2898DeriveBytes(Password: Text; InputText: Text; PsuedoRandomNumber: Integer; FillCharacter: Integer) Hash: Text
```
darjoo: The function has to be documented as well. Please take a look at https://github.com/microsoft/ALAppExtensions/blob/c07cb0dc989fdd022a9dc18eb8db22a190b5fefb/Modules/System/Cryptography%20Management/src/RijndaelCryptography.Codeunit.al#L56-L62

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,31 @@
...
+        Convert: DotNet Convert;
+        Encoding: DotNet Encoding;
+        Rfc2898DeriveBytes: DotNet Rfc2898DeriveBytes;
+        SHA1Managed: DotNet SHA1Managed;
+    begin
+        if InputText = '' then
+            exit;
+
+        //Implement password-based key derivation functionality, PBKDF2, by using a pseudo-random number generator based on HMACSHA1.
+        Rfc2898DeriveBytes := Rfc2898DeriveBytes.Rfc2898DeriveBytes(Password, Encoding.ASCII.GetBytes(PadStr(InputText, FillCharacter)));
```
darjoo: Is this InputText variable the salt?

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,31 @@
...
+
+        //Return a Base64 encoded string of the SHA1 hash of the first X bytes (X = PsuedoRandomNumber) returned from the generator.
+        SHA1Managed := SHA1Managed.SHA1Managed;
+        ByteArray := SHA1Managed.ComputeHash(Rfc2898DeriveBytes.GetBytes(PsuedoRandomNumber));
+        Hash := Convert.ToBase64String(ByteArray);
+
+        //Clear all used variables
```
darjoo: Unnecessary to clear local variables at the end of the function.

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
darjoo: Make the codeunit Access = Public;

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
darjoo: Rename function to HashRfc2898DeriveBytes

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,31 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+codeunit 50101 "Rfc2898DeriveBytes"
+{
+    procedure FuncCreateHashWithRfc2898DeriveBytes(Password: Text; InputText: Text; PsuedoRandomNumber: Integer; FillCharacter: Integer) Hash: Text
+    var
```
darjoo: Move the implementation code into https://github.com/microsoft/ALAppExtensions/blob/master/Modules/System/Cryptography%20Management/src/CryptographyManagementImpl.Codeunit.al

This codeunit acts as a facade layer in between, calling that new function.

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
angela2389: done!

angela2389: done!

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
angela2389: done!

angela2389:
Thanks @darjoo, can you recheck?

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,41 @@
...
+
+    procedure HashRfc2898DeriveBytes(Password: Text; Salt: Text; DesiredSaltLength: Integer; PsuedoRandomNumber: Integer) Hash: Text
+    var
+        ByteArray: DotNet Array;
+        Convert: DotNet Convert;
+        Encoding: DotNet Encoding;
+        Rfc2898DeriveBytes: DotNet Rfc2898DeriveBytes;
+        SHA1Managed: DotNet SHA1Managed;
+    begin
+        if InputText = '' then
```
darjoo: Should be Salt as well otherwise this won't compile :)

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
darjoo: Facade layers do not contain implementations, as they're the APIs being called from other extensions.
While the Impl layer, "Cryptography Management Impl." contains the implementation of said APIs.

This function should only contain one variable for the impl codeunit (CryptographyManagementImpl) and one line CryptographyManagementImpl.HashRfc2898DeriveBytes(….).
While the CryptographyManagementImpl.HashRfc2898DeriveBytes function contains all the logic we have here right now.

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,41 @@
...
+    Access = Public;
+
+    /// <summary>
+    /// Initializes a new instance of the Rfc2898DeriveBytes class providing the Password, Salt, Saltlength and Psuedorandomnumber.
+    /// </summary>
+    /// <param name="Password">Represents the password to be used to initialize a new instance of Rfc2898DeriveBytes</param>
+    /// <param name="Salt">Represents the salt to be used to initialize a new instance of Rfc2898DeriveBytes</param>
+    /// <param name="DesiredSaltLength">Represents the desired saltlength, if the Salt is longer or shorter it will change to the specified length</param>
+    /// <param name="PsuedoRandomNumber">Represents the PsuedoRandomNumber, which is used to transform Rfc2898DeriveByes with SHA1Managed into a ByteArray</param>
+
```
darjoo: Extra space to be removed

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,41 @@
...
+    procedure HashRfc2898DeriveBytes(Password: Text; Salt: Text; DesiredSaltLength: Integer; PsuedoRandomNumber: Integer) Hash: Text
```
darjoo: Change PseudoRandomNumber to NoOfBytes, makes more sense from what I read from https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rfc2898derivebytes.getbytes?view=netframework-4.8.

darjoo: After discussing with my colleagues, it does not make sense to have DesiredSaltLength as part of the API. We should not be modifying the salt in this function. 

The PadStr(Salt...) below should be carried out before calling this function.

angela2389: Oke, I added the PadStr in because if the Salt is too short it will give an error. It's ofcourse fine for us to check it beforehand. I will remove it.

Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
```diff
@@ -0,0 +1,22 @@
...
+    procedure TestDeriveBytes()
+    var
+        CodRfc2898DeriveBytes: Codeunit Rfc2898DeriveBytes;
+        EncryptedText: Text;
+        ExpectedEncryptedText: Label 'RgIsn9T5fqPK8bsXjzlmWqinRxg=';
+    begin
+        // Encrypt Text 
+        EncryptedText := CodRfc2898DeriveBytes.HashRfc2898DeriveBytes('Test', 'Test123', 23, 10);
+        // Verify Result 
+        IF (EncryptedText <> ExpectedEncryptedText) THEN
```
mazhelez: `if`, `then`, `Error`*

Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
```diff
@@ -0,0 +1,22 @@
...
+            ERROR('Failed to encrypt text');
```
mazhelez: Please, user [Library Assert](https://github.com/microsoft/ALAppExtensions/tree/master/Modules/DevTools/TestFramework/TestLibraries/Assert) :)

Also, I recommend following the Given-When-Then template (see [other tests](https://github.com/microsoft/ALAppExtensions/tree/master/Modules/DevTools/TestFramework/TestLibraries/Assert))

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
mazhelez: Having `DesiredSaltLength` doesn't really prevent an error from occurring. 

You need to handle the `ArgumentException` in a way that it does not pops out.

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,41 @@
...
+        Convert: DotNet Convert;
+        Encoding: DotNet Encoding;
+        Rfc2898DeriveBytes: DotNet Rfc2898DeriveBytes;
+        SHA1Managed: DotNet SHA1Managed;
+    begin
+        if InputText = '' then
+            exit;
+
+        //Implement password-based key derivation functionality, PBKDF2, by using a pseudo-random number generator based on HMACSHA1.
+        Rfc2898DeriveBytes := Rfc2898DeriveBytes.Rfc2898DeriveBytes(Password, Encoding.ASCII.GetBytes(PadStr(Salt, DesiredSaltLength)));
```
mazhelez: Any particular reason to use ASCII?

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,41 @@
...
+
+        //Return a Base64 encoded string of the SHA1 hash of the first X bytes (X = PsuedoRandomNumber) returned from the generator.
+        SHA1Managed := SHA1Managed.SHA1Managed;
```
mazhelez: From the [official documentation of SHA1Managed](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha1managed.-ctor?view=netframework-4.8)

_Due to collision problems with SHA1, Microsoft recommends a security model based on SHA256 or better._

Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
```diff
@@ -0,0 +1,22 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+codeunit 132576 "Rfc2898DeriveBytes Test"
```
mazhelez: Extra line between the license header and the codeunit definition is needed.

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
angela2389: Hi Mazhelez, for SHA1 and ASCII: We implemented this method within C/AL and secured many files with this. If customers updated from C/AL and AL, they will lose access to these files if we cannot use the same encryption method.

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
```diff
@@ -0,0 +1,26 @@
...
+{
+    Access = Public;
+
+    /// <summary>
+    /// Initializes a new instance of the Rfc2898DeriveBytes class providing the Password, Salt and NoOfBytes.
+    /// </summary>
+    /// <param name="Password">Represents the password to be used to initialize a new instance of Rfc2898DeriveBytes</param>
+    /// <param name="Salt">Represents the salt to be used to initialize a new instance of Rfc2898DeriveBytes</param>
+    /// <param name="NoOfBytes">Represents the NoOfBytes, which is used to transform Rfc2898DeriveByes with SHA1Managed into a ByteArray</param>
+    procedure HashRfc2898DeriveBytes(Password: Text; Salt: Text; NoOfBytes: Integer) Hash: Text
```
darjoo: Sorry, one more change to the name.
HashRfc2898DeriveBytesWithSha1Managed

After speaking with my colleagues, we will create a separate function which allows the user to choose between the different hashing algorithms. Nothing that needs to be done on your side except the name change here and in CryptographyManagementImpl.Codeunit.al.

Modules/System/Cryptography Management/src/CryptographyManagementImpl.Codeunit.al
```diff
@@ -590,4 +590,24 @@ codeunit 1279 "Cryptography Management Impl."
         if IsNull(RijndaelProvider) then
             InitRijndaelProvider();
     end;
+
+    procedure HashRfc2898DeriveBytes(Password: Text; Salt: Text; NoOfBytes: Integer)
```
darjoo: add parameter HashAlgorithmType: Option MD5,SHA1,SHA256,SHA384,SHA512

Modules/System/Cryptography Management/src/CryptographyManagementImpl.Codeunit.al
```diff
@@ -590,4 +590,24 @@ codeunit 1279 "Cryptography Management Impl."
...
+    begin
+        if Salt = '' then
+            exit;
+
+        //Implement password-based key derivation functionality, PBKDF2, by using a pseudo-random number generator based on HMACSHA1.
+        Rfc2898DeriveBytes := Rfc2898DeriveBytes.Rfc2898DeriveBytes(Password, Encoding.ASCII.GetBytes(Salt));
+
+        //Return a Base64 encoded string of the SHA1 hash of the first X bytes (X = NoOfBytes) returned from the generator.
+        SHA1Managed := SHA1Managed.SHA1Managed;
+        ByteArray := SHA1Managed.ComputeHash(Rfc2898DeriveBytes.GetBytes(NoOfBytes));
```
darjoo: Remove SHA1Managed variable

if not TryGenerateHash(ByteArray, Rfc2898DeriveBytes.GetBytes(NoOfBytes), Format(HashAlgorithmType)) then
    Error(GetLastErrorText());

Modules/System/Cryptography Management/src/Rfc2898DeriveBytes.Codeunit.al
darjoo: add parameter HashAlgorithmType: Option MD5,SHA1,SHA256,SHA384,SHA512

Modules/System Tests/Cryptography Management/src/Rfc2898DeriveBytesTest.Codeunit.al
```diff
@@ -0,0 +1,30 @@
...
+    var
+        CodRfc2898DeriveBytes: Codeunit Rfc2898DeriveBytes;
+        EncryptedText: Text;
+        ExpectedEncryptedText: Text;
+        LibraryAssert: Codeunit "Library Assert";
+    begin
+        // [GIVEN] With Encryption Key 
+        ExpectedEncryptedText := GetRfc2898DeriveBytesEncryptedText();
+        // [WHEN] Encrypt Text 
+        EncryptedText := CodRfc2898DeriveBytes.HashRfc2898DeriveBytes('Test', 'Test1234', 23);
```
darjoo: Update call to HashRfc2898DeriveBytes, to use option SHA1.

bc-ghost:

bc-ghost:

bc-ghost:

PR #6060, opened 2020-02-13
Set LastErrorText as Error Message Preview

SimonTwilfer:
Setting LastErrorText as Error Message Preview in Test Method Line
instead of the Error Code which is already used for the Error Code.


JesperSchulz:
Nice one. I'll put that into our next major release for April :-) Thanks for fixing!

bc-ghost:

PR #6238, opened 2020-02-26
Set TestRunnerDisplayName after Lookup Test Runner

SimonTwilfer:
After selecting a Test Runner in the AL Test Tool the displayed name is not updated.


bc-ghost:

PR #6808, opened 2020-04-19
Publisher OnBeforeGetCustomizedCueStyleOption in Codeunit 9702

miljance:
With this change it is possible to dynamically decide on the Cue Style using enum "Cues And KPIs Style" but regardless of the the decimal Cue Value. 

Explanation: The Cue should be shown in red if there are some rejected entries behind the Document Count.

Example: The Cue shows Number 10 indicating that there are 10 Documents involved. Out of those 10 any or all of those 10 could be critical or marked as rejected and needs checking out first. So only in this case the Cue should have a Style 'Unfavorable'.

### DIFF
--- a/Modules/System/Cues and KPIs/src/CuesAndKPIs.Codeunit.al
+++ b/Modules/System/Cues and KPIs/src/CuesAndKPIs.Codeunit.al
@@ -100,6 +100,7 @@ codeunit 9701 "Cues And KPIs"
     /// </summary>
     /// <param name="TableID">The ID of the table where the cue is defined.</param>
     /// <param name="FieldNo">The ID of the field which the cue is based on.</param>
+    /// <param name="CueValue">Cue value parameter that can be used to determine cue style.</param>
     /// <param name="CueStyle">Exit parameter that holds newly determined cue style based on custom prerequisites.</param>
     /// <param name="Resolved">A boolean value that describes whether or not the cue style has been determined.</param>
     [IntegrationEvent(false, false)]
--- a/Modules/System/Cues and KPIs/src/CuesAndKPIs.Codeunit.al
+++ b/Modules/System/Cues and KPIs/src/CuesAndKPIs.Codeunit.al
@@ -94,8 +94,16 @@ codeunit 9701 "Cues And KPIs"
     begin
     end;
 
+    /// <summary>
+    /// Event that allows definition of cue style for a cue using style enum without the usage of a cue setup table.
+    /// Subscribe to this event if you want to define a cue style for a cue using custom prerequisites.
+    /// </summary>
+    /// <param name="TableID">The ID of the table where the cue is defined.</param>
+    /// <param name="FieldNo">The ID of the field which the cue is based on.</param>
+    /// <param name="CueStyle">Exit parameter that holds newly determined cue style based on custom prerequisites.</param>
+    /// <param name="Resolved">A boolean value that describes whether or not the cue style has been determined.</param>
     [IntegrationEvent(false, false)]
-    procedure OnBeforeGetCustomizedCueStyleOption(TableId: Integer; FieldNo: Integer; CueValue: Decimal; var Style: Enum "Cues And KPIs Style"; var Handled: Boolean)
+    procedure OnBeforeGetCustomizedCueStyleOption(TableID: Integer; FieldNo: Integer; CueValue: Decimal; var CueStyle: Enum "Cues And KPIs Style"; var Resolved: Boolean)
     begin
     end;
 }
--- a/Modules/System/Cues and KPIs/src/CuesAndKPIs.Codeunit.al
+++ b/Modules/System/Cues and KPIs/src/CuesAndKPIs.Codeunit.al
@@ -93,4 +93,9 @@ codeunit 9701 "Cues And KPIs"
     internal procedure OnConvertStyleToStyleText(CueStyle: Enum "Cues And KPIs Style"; var Result: Text; var Resolved: Boolean)
     begin
     end;
+
+    [IntegrationEvent(false, false)]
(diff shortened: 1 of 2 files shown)


Modules/System/Cues and KPIs/src/CuesAndKPIsImpl.Codeunit.al
```diff
@@ -259,4 +264,9 @@ codeunit 9702 "Cues And KPIs Impl."
               CueSetup.FieldCaption("Threshold 2"),
               CueSetup.FieldCaption("Threshold 1"));
     end;
+
+    [IntegrationEvent(false, false)]
```
mazhelez, line 268: Having an integration event in the implementation would not work as expected.
Implementation codeunits, like this one, are internal so you won't be able to subscribe for this event.

To solve that, just move the event in the facade codeunit and call it here.

See [this example](https://github.com/microsoft/ALAppExtensions/blob/a499f0afcde09e0d836398a5d1abc2890b6ad7fb/Modules/System/Language/src/LanguageImpl.Codeunit.al#L19).

miljance: Thank you for clarification. I have overlooked the internal property. I made necessary change. I hope that the change fits the prerequisites now.

Modules/System/Cues and KPIs/src/CuesAndKPIs.Codeunit.al
```diff
@@ -93,4 +93,9 @@ codeunit 9701 "Cues And KPIs"
     internal procedure OnConvertStyleToStyleText(CueStyle: Enum "Cues And KPIs Style"; var Result: Text; var Resolved: Boolean)
     begin
     end;
+
+    [IntegrationEvent(false, false)]
+    procedure OnBeforeGetCustomizedCueStyleOption(TableId: Integer; FieldNo: Integer; CueValue: Decimal; var Style: Enum "Cues And KPIs Style"; var Handled: Boolean)
```
mazhelez, line 98: Missing documentation 😃

miljance: I hope it fits now

Modules/System/Cues and KPIs/src/CuesAndKPIs.Codeunit.al
```diff
@@ -93,4 +93,17 @@ codeunit 9701 "Cues And KPIs"
...
+    /// <summary>
+    /// Event that allows definition of cue style for a cue using style enum without the usage of a cue setup table.
+    /// Subscribe to this event if you want to define a cue style for a cue using custom prerequisites.
+    /// </summary>
+    /// <param name="TableID">The ID of the table where the cue is defined.</param>
+    /// <param name="FieldNo">The ID of the field which the cue is based on.</param>
+    /// <param name="CueStyle">Exit parameter that holds newly determined cue style based on custom prerequisites.</param>
+    /// <param name="Resolved">A boolean value that describes whether or not the cue style has been determined.</param>
+    [IntegrationEvent(false, false)]
+    procedure OnBeforeGetCustomizedCueStyleOption(TableID: Integer; FieldNo: Integer; CueValue: Decimal; var CueStyle: Enum "Cues And KPIs Style"; var Resolved: Boolean)
```
mazhelez, line 106: `CueValue` should be documented as well.

Modules/System/Cues and KPIs/src/CuesAndKPIsImpl.Codeunit.al
```diff
@@ -121,7 +121,12 @@ codeunit 9702 "Cues And KPIs Impl."
     local procedure GetCustomizedCueStyleOption(TableId: Integer; FieldNo: Integer; CueValue: Decimal): Integer
     var
         CueSetup: Record "Cue Setup";
+        CueStyle: Enum "Cues And KPIs Style";
+        Resolved: Boolean;
     begin
+        CuesAndKPIs.OnBeforeGetCustomizedCueStyleOption(TableID, FieldNo, CueValue, CueStyle, Resolved);
+        if Resolved then
```
mazhelez, line 128: The pattern is somewhere called `Handled`. Not that I mind the `Resolved` version, but what made you change your mind?

Modules/System/Cues and KPIs/src/CuesAndKPIsImpl.Codeunit.al
```diff
@@ -121,7 +121,12 @@ codeunit 9702 "Cues And KPIs Impl."
...
+            exit(Style);
```
mazhelez, line 129: `CueStyle` maybe? 🤔

Modules/System/Cues and KPIs/src/CuesAndKPIs.Codeunit.al
miljance: Thanks, I missed the spot.

Modules/System/Cues and KPIs/src/CuesAndKPIsImpl.Codeunit.al
miljance: I do not have anything against naming it "Handled". I saw other publisher in Codeunit 9701 called OnConvertStyleToStyleText and the term "Resolved" and it sounded better than Handled at least in the case for this purpose in this context.

Modules/System/Cues and KPIs/src/CuesAndKPIsImpl.Codeunit.al
miljance: Indeed. I was too quick with the changes. Corrected.

bc-ghost:

bc-ghost:

PR #7245, opened 2020-05-28
Add logic to give better Error Message

daansaveyn:
When the TryValidateAgainstSchema function was used in combination with the GetLastErrorText functionality, the error message was unclear. The actual validation error was not displayed. 

For example:

image

These code modifications ensure the actual validation error is shown:

image

### DIFF
--- a/Modules/System/XML Validation/src/XmlValidationImpl.Codeunit.al
+++ b/Modules/System/XML Validation/src/XmlValidationImpl.Codeunit.al
@@ -11,7 +11,7 @@ codeunit 6241 "Xml Validation Impl."
         InvalidXmlErr: Label 'The XML definition is invalid.';
         InvalidSchemaErr: Label 'The schema definition is not valid XML.';
         [WithEvents]
-        XmlValidatingReader: dotnet XmlValidatingReader;
+        XmlSchemaSet: dotnet XmlSchemaSet;
 
     procedure ValidateAgainstSchema(Xml: Text; XmlSchema: Text; Namespace: Text)
     var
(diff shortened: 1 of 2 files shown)


JesperSchulz:
@ajkauffmann, as this module originates from your PR, would you mind reviewing this change? :-) If you approve, I will continue processing the PR.

ajkauffmann:
Sorry... had a crazy week. Will look into it asap. Have a few questions I believe.

ajkauffmann:
@JesperSchulz the PR is using XmlValidatingReader class. According to documentation, this class is obsolete, starting with .Net Framework 2.0. Instead, XmlReader should be used.

Source: https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmlvalidatingreader

daansaveyn:
> @JesperSchulz the PR is using XmlValidatingReader class. According to documentation, this class is obsolete, starting with .Net Framework 2.0. Instead, XmlReader should be used.
> 
> Source: https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmlvalidatingreader

Thank you for the feedback. I must have overlooked this. I'll check the code and see what I can do.

daansaveyn:
I've just pushed a new commit. 

There's however a small change in behavior: if the xmlns attribute is not specified (for example xmlns="http://www.w3.org/2001/XMLSchema") xml validation will not fail in case of an invalid XML file. This was not the case previously, but seems to be .net behavior. Currently, this is not a problem for us, but it might be for others.

JesperSchulz:
@ajkauffmann, do you have any comments to the recent changes? We should really get some traction on wrapping this PR up.

ajkauffmann:
My apologies, I wasn't aware that this was waiting for me.

I agree with @daansaveyn that the error message should not include the text 

`A call to System.Xml.XmlDocument.Validate failed with this message: `

But I believe this can be achieved with less effort. The real message is in the inner exception, so we only need to catch the error object and then throw the error again with the message from the inner exception. I will create a pull request for that, which only contains a few lines.

GiorgosPa:
Thank you for your suggestion, we will proceed with [Pull Request 10175](https://github.com/microsoft/ALAppExtensions/pull/10175)

PR #7876, opened 2020-07-15
Xml Writer

angela2389:
Hi Microsoft team,

Can you please review this Pull Request? Hopefully this will be the last one :)

### DIFF
--- a/Modules/System/XmlWriter/src/XmlWriter.Codeunit.al
+++ b/Modules/System/XmlWriter/src/XmlWriter.Codeunit.al
@@ -1,0 +1,90 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+/// <summary>
+// Provides helper functions for System.Xml.XmlWriter
+/// </summary>
+
+codeunit 1385 "XmlWriter"
+{
+    Access = Public;
+
+    var
+        XmlWriterImpl: Codeunit "XmlWriter Impl.";
+
+    /// <summary>
+    /// Creates the XmlWriter Document
+    /// </summary>
+    procedure XmlWriterCreateDocument()
+    begin
+        XmlWriterImpl.XmlWriterCreateDocument();
+    end;
+
+    /// <summary>
+    /// When overridden in a derived class, writes the specified start tag.
+    /// </summary>
+    /// <param name="Prefix">The namespace prefix of the element.</param>
+    /// <param name="LocalName">The local name of the element.</param>
+    /// <param name="NameSpace">The namespace URI to associate with the element. If this namespace is already in scope and has an associated prefix, the writer automatically writes that prefix also.</param>
+    procedure XmlWriterStartElement(Prefix: Text; LocalName: Text; NameSpace: Text)
+    begin
+        XmlWriterImpl.XmlWriterStartElement(Prefix, LocalName, NameSpace);
+    end;
+
+    /// <summary>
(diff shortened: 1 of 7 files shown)


JesperSchulz:
Before we start looking into your PR, could you elaborate on where our [native XML Document implementation](https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/xmldocument/xmldocument-data-type) doesn't suffice? Instead of creating a competing XML writer, could we add to our native implementation to suite your needs?

angela2389:
Hi @JesperSchulz,

I do use XML Documents and it works well for small files, but in case of big XML files we have a lot of performance issues. Customers get impatient if the loading time goes over 1 minute. Using Xml Writer vs Xml Document makes a huge difference.

Please see the results of @AndersMad (we are not in the same company) on https://github.com/microsoft/AL/issues/999 and Business Central Idea: https://experience.dynamics.com/ideas/idea/?ideaid=f07b7746-6c80-e911-80e7-0003ff68bdb5

JesperSchulz:
Got it, thanks! Let's see what we can do here :-)

Modules/System/XmlWriter/src/XmlWriterImpl.Codeunit.al
```diff
@@ -0,0 +1,61 @@
...
+    end;
+
+    procedure XmlWriterElementString(LocalName: Text; ElementValue: Text)
+    begin
+        XmlTextWriter.WriteElementString(LocalName, ElementValue);
+    end;
+
+    procedure XmlWriterEndElement()
+    begin
+        XmlTextWriter.WriteEndElement;
```
JesperSchulz, line 36: Let's keep the original function names. If the method was called WriteEndElement, let's call our AL function WriteEndElement.

Modules/System/XmlWriter/src/XmlWriterImpl.Codeunit.al
```diff
@@ -0,0 +1,61 @@
...
+    end;
+
+    procedure XmlWriterAddAttribute(ElementName: Text; ElementValue: Text; Prefix: Text)
+    begin
+        if Prefix <> '' then
```
JesperSchulz, line 41: Let's not put logic, where none is needed. Remember, that we also have function-overloads in al now!

Modules/System/XmlWriter/src/XmlWriter.Codeunit.al
```diff
@@ -0,0 +1,90 @@
...
+/// <summary>
+// Provides helper functions for System.Xml.XmlWriter
+/// </summary>
+
+codeunit 1385 "XmlWriter"
+{
+    Access = Public;
+
+    var
+        XmlWriterImpl: Codeunit "XmlWriter Impl.";
```
JesperSchulz, line 15: I wonder, if we even need an Impl. codeunit for this one? Right now, the facade calls the impl. codeunit with a one-liner. The impl. codeunit the again calls the .NET method with another one-liner - seems a little over-wrapped to me :-)

angela2389: If it's fine for you to do without I will remove it :), I'm just trying to follow the template of the other modules.

angela2389: Actually, the XmlWriterCreateDocument function has four lines now..

JesperSchulz:
We will process your request within the month of September :-)

JesperSchulz (approved):

bc-ghost:

bc-ghost:
Thanks for submitting this issue. We will publish a fix for it in the next minor update for release 17.

bc-ghost:

PR #8264, opened 2020-08-10
Update COD13623.MigrateToExtensionV2.al

RasmusTidselbak:
Fixed an issue where GLN is not included in upgrades.


ChristopherDedominici:
By design: the field "GLN" is missing because we don't provide new code that is using obsoleted fields.

JesperSchulz:
Closing PR based on the following input:
_By design: the field "GLN" is missing because we don't provide new code that is using obsoleted fields._

PR #8657, opened 2020-09-15
Any.DecimalInRange fixed

NKarolak:
Sometimes `Any.DecimalInRange` returned a decimal number, whose No. of decimal places did not equal the parameter.
This now has been fixed in accordance with @nikolakukrika. It should be ported afterwards to library random.

The variable name `pow` has been chosen in accordance to `procedure DecimalInRange(MinValue: Decimal; MaxValue: Decimal; DecimalPlaces: Integer): Decimal`.

I am unsure, but it looks like `procedure DecimalInRange(MinValue: Decimal; MaxValue: Decimal; DecimalPlaces: Integer): Decimal` might need a fix as well.

Please refer to https://www.yammer.com/dynamicsnavdev/threads/857565538000896 for further information.

### DIFF
--- a/Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
+++ b/Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
@@ -44,8 +44,17 @@ codeunit 130500 "Any"
     /// <param name="DecimalPlaces">Number of decimal places</param>
     /// <returns>Pseudo-random decimal value</returns>
     procedure DecimalInRange(MaxValue: Integer; DecimalPlaces: Integer): Decimal
+    var
+        x: Integer;
+        Pow: Integer;
     begin
-        exit(IntegerInRange(MaxValue * Power(10, DecimalPlaces)) / Power(10, DecimalPlaces));
+        Pow := Power(10, DecimalPlaces);
+
+        x := IntegerInRange(MaxValue * Pow);
+        if x mod 10 = 0 then
+            x += IntegerInRange(1, 9);
+
+        exit(x / Pow);
     end;
 
     /// <summary>


Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
```diff
@@ -44,8 +44,17 @@ codeunit 130500 "Any"
     /// <param name="DecimalPlaces">Number of decimal places</param>
     /// <returns>Pseudo-random decimal value</returns>
     procedure DecimalInRange(MaxValue: Integer; DecimalPlaces: Integer): Decimal
+    var
+        x: Integer;
+        Pow: Integer;
```
StefanMaron, line 49: I would rename Pow to NumberWithLengthEqualsDecimalPlaces. This is much more understandable for future code readers

Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
```diff
@@ -44,8 +44,17 @@ codeunit 130500 "Any"
     /// <param name="DecimalPlaces">Number of decimal places</param>
     /// <returns>Pseudo-random decimal value</returns>
     procedure DecimalInRange(MaxValue: Integer; DecimalPlaces: Integer): Decimal
+    var
+        x: Integer;
```
StefanMaron, line 48: I would rename x to RandomIntegerInRange. This is much more understandable for future code readers

Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
NKarolak: If you decide to rename Pow here, you'll need to update it also in 

- `procedure DecimalInRange(MinValue: Decimal; MaxValue: Decimal; DecimalPlaces: Integer): Decimal`
- `procedure DecimalInRange(MaxValue: Integer; DecimalPlaces: Integer): Decimal`

nikolakukrika: It is too long, Pow or Power is intuitive enough. Let's keep the change small.

Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
nikolakukrika: PseudoRandomInteger? It is not random it is PseudoRandom. InRange is just implementation detail, thus I would not reflect it in the name. It is shorter as well.

Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
nikolakukrika: I would use the very descriptive name if the method longer, for very a small method like this one, Pow or Power will do.

Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
NKarolak: @nikolakukrika 
>  It is not random it is PseudoRandom.

Just for my information; could you please explain the difference?

nikolakukrika: If you call Library Any or Library Random and ask for randIntInRange 5 times and get 11, 25, 1, 2, 4
When this transaction ends and you call the same function again, you will get the same numbers - 11, 25, 1, 2, 4. 

With true random, you would get a completely different set of numbers.

This behavior is essential for the tests, otherwise we would not be able to reproduce failures on a local box. Pseudorandom is achieved by always setting a seed for the Any library and in a weird way in Library Random : ). I will implement clearing the seed after every test method for library any, so the tests in a codeunit and the tests individually will always get the same numbers.

bc-ghost:

JesperSchulz:
Our ghost doesn't seem to merge the PR before it closes it. Reopening :-)

JesperSchulz (approved):

bc-ghost:
Thanks for submitting this issue. We will publish a fix for it in the next minor update for release 17.

Issue #9558, opened 2020-10-22
[Extensibility Request] table 8888 "Email Outbox" - Add Getter Functions


pri-kise:
Could you please add some Getter Functions to table "Email Outbox". and the Table "Sent Email"?

```al
procedure GetMessageId(var MessageId: Guid)
begin
    MessageId:= Rec."Account Id";
end;
procedure GetAccountId(var AccountId: Guid)
begin
    AccountId := Rec."Account Id";
end;
procedure GetUserSecurityId(var UserSecurityId: Guid)
begin
    UserSecurityId:= Rec."User Security Id";
end;
procedure GetConnector():Enum "Email Connector"
begin
    exit(Rec.Connector);
end;
procedure GetEmailStatus(): Enum "Email Status"
begin
    exit(Rec.Status);
end;
procedure GetDescription(): Text[2048]
begin
    exit(Rec.Description);
end;
procedure GetErrorMessage(): Text[2048]
begin
    exit(Rec."Error Message");
end;
procedure GetDateQueued(): DateTime
begin
    exit(Rec."Date Queued");
end;
procedure GetDateFailed(): DateTime
begin
    exit(Rec."Date Failed");
end;
procedure GetDateQueued(): DateTime
begin
    exit(Rec."Date Queued");
end;
procedure GetSendFrom(): Text[250]
begin
    exit(Rec."Send From");
end;
```

dtkb:
What is the use of such getters on a record, where the fields are already publicly accessible?

pri-kise:
I'm using the latest insider preview for BC17 (System.Application: 17.1.17933.0)  where those fields are set Access=Internal. 
https://github.com/microsoft/ALAppExtensions/blob/2ce8069cae5e9ac3929998b57d6b2785a59e4572/Modules/System/Email/src/Email/Outbox/EmailOutbox.Table.al#L22
The table is set to Access=Public (https://github.com/microsoft/ALAppExtensions/blob/2ce8069cae5e9ac3929998b57d6b2785a59e4572/Modules/System/Email/src/Email/Outbox/EmailOutbox.Table.al#L9), but the fields are accesibly for me
![image](https://user-images.githubusercontent.com/56249171/96961675-7365c000-1505-11eb-8799-199cd8041542.png)

If the Access=Internal is removed this would be fine, too.

mazhelez:
Hey @pri-kise 

[...]

Issue #9527, opened 2020-10-22
codeunit 8888 "Email Dispatcher" - OnAfterUpdateOutboxStatus


pri-kise:
Can we please get the following event?
```al
local procedure UpdateOutboxStatus(var EmailOutbox: Record "Email Outbox"; Status: Enum "Email Status")
begin
    EmailOutbox.Status := Status;
    EmailOutbox.Modify();
    Commit();
    OnAfterUpdateOutboxStatus(EmailOutbox. Status);//NEW
end;

```

```al
[IntegrationEvent(false, false)]
local procedure  OnAfterUpdateOutboxStatus(EmailOutbox: Record "Email Outbox"; NewStatus: Enum "Email Status")
begin
end;
```

pri-kise:
@StanPesotskiy  I dont't know if this event is even possible if the Codeunit is set to internal. Maybe the implementation Codeunit has to call the Event placed into another Codeunit with Access=Public....?

darjoo:
@pri-kise Could you please let me know why you need this event? What is the scenario you're trying to achieve?

pri-kise:
We want to send optional notifications to the user if the senidng of an e-mail outbox entry fails.
For this reason we need an event when the status is beeing updated. An since the property Status is to internal I request the new Status, too.

darjoo:
I'll go ahead and put in an event that is triggered when the email fails to send.
It'll be something along the lines of `OnAfterEmailFailedToSend(Rec: Record 'Email Outbox')`

JesperSchulz:
Sounds like this PR is linked together with #9613, which we're currently discussing. While we're doing so, this on should be set on hold as well. Let's get the design of this right.

GiorgosPa:
@pri-kise thank you for your suggestion, we have implemented it a little bit differently. I hope it suits your needs.

```
    /// <summary>
    /// Integration event that notifies senders about whether their email was successfully sent in the background.
    /// </summary>
    /// <param name="MessageId">The ID of the email in the queue.</param>
    /// <param name="Status">True if the message was successfully sent.</param>
    [IntegrationEvent(false, false)]
    internal procedure OnAfterSendEmail(MessageId: Guid; Status: Boolean)
    begin
    end;
```

Issue #9005, opened 2020-10-01
SentEmail.Table : Include 'Source Record ID' field to improve tracability


fvet:
Previously we had built our own 'Sent Emails' feature, that logged all outgoing mails, including (links to) attachments.
We're considering removing our custom feature and replace it with the upcoming 'Sent Emails' feature. 

One downside is the storage of the attachments as BLOB. We already store the attachment in separate tables, so the blob approach will duplicate the database size. (unless we forcily remove the records with 'Retention Policies') This is not inevitable.

The other thing we currently lack is a link between the initiating record and the 'Sent Emails'. We had integrated some logic to transfer the 'Source Record Id' (whenever possible, not for grouped mailing) in the 'Sent Emails' table. This allowed usefull navigation options both from the Sent Email to the source record (e.g. Sales Invoice) or to have better insights on the source record (Sales Invoice) if and what attachments have been sent by mail. (often you might want to check / debug if all attachments have been sent correctly, no attachments are missing, the PDF is correct, ...) Or in case a customer refers to a mail, you have all data on hand immediately. (This only works as long as the Sent Emails are retained)

**Ask:**
- Is it possible to improve the 'Email' framework to include a 'Source Record ID' (since it's currently not open to extend), so we can restore this logic when working with the new 'Email Feature'?

JesperSchulz:
That is absolutely possible! By end of week, I expect us to open source the source code for our new email module and add-on apps. We would love to receive a pull request from you, which adds the capabilities you think are missing. For now, I will leave this issue open. Let's reconnect about this, once our new email solution is ready for public - then we can see, if this is something we can collaborate on, or if it is something, we'll take in as a "suggestion for improvement". No doubt, that it's a great idea.

mazhelez:
Hey @fvet 

There has been some changes in the extensibility model of the email module. Do you mind having a second look?
Now, _Sent Email_ is extensible to maybe you can build the needed solution as an extension.

JesperSchulz:
I am happy to inform you, that the link between the initiating record and the 'Sent Email' is being established in the upcoming major release of Business Central (version 18 in April). If the new implementation doesn't fully meet your requirements, please create a new issue for us - or even better, create a pull request for us :-)

pri-kise:
Thanks. Yeah. @JesperSchulz  is this already available in the insider builds? I will check it out and really looking forward to this feature :)

fvet:
[...]

PR #9613, opened 2020-10-27
Add Getter Functions for E-Mail Outbox / E-Mail Sent

pri-kise:
Add Getter Functions as described in issue #9558

### DIFF
--- a/Modules/System/Email/src/Email/Outbox/EmailOutbox.Table.al
+++ b/Modules/System/Email/src/Email/Outbox/EmailOutbox.Table.al
@@ -113,34 +113,30 @@ table 8888 "Email Outbox"
     /// <summary>
     /// Returns the "Message Id" of the Email Outbox Entry
     /// </summary>
-    /// <param name="MessageId"></param>
-    procedure GetMessageId(var MessageId: Guid)
+    procedure GetMessageId(): Guid
     begin
-        MessageId := Rec."Message Id";
+        exit(Rec."Message Id");
     end;
 
     /// <summary>
     /// Returns the "Account Id" of the Email Outbox Entry
     /// </summary>
-    /// <param name="AccountId"></param>
-    procedure GetAccountId(var AccountId: Guid)
+    procedure GetAccountId(): Guid
     begin
-        AccountId := Rec."Account Id";
+        exit(Rec."Account Id");
     end;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


Modules/System/Email/src/Email/Outbox/EmailOutbox.Table.al
```diff
@@ -109,4 +109,54 @@ table 8888 "Email Outbox"
...
+    begin
+        exit(Rec.Connector);
+    end;
+
+    procedure GetEmailStatus(): Enum "Email Status"
+    begin
+        exit(Rec.Status);
+    end;
+
+    procedure GetDescription(): Text[2048]
```
mazhelez, line 138: I'd suggest dropping the size of the result to avoid breaking changes if we decide to change it.

Modules/System/Email/src/Email/Outbox/EmailOutbox.Table.al
```diff
@@ -109,4 +109,54 @@ table 8888 "Email Outbox"
         {
         }
     }
+
+    procedure GetMessageId(var MessageId: Guid)
```
mazhelez, line 113: Isn't `Guid` an allowed return type? I believe it is.

mazhelez (changes requested):
Don't forget to briefly document the new getters :)

Refer to [AL code documentation](https://github.com/microsoft/ALAppExtensions/wiki/AL-Code-Documentation).

Modules/System/Email/src/Email/Outbox/EmailOutbox.Table.al
pri-kise: Yes, you were right. When I first wrote those getters i didn't got Guid with the Intellisense, but now it looks fine.

JesperSchulz:
PR also linked to #9527.

pri-kise:
@JesperSchulz I haven't thought of the privacy mechanism yet, but that increases the complexity of such a small feature.
I could add a check that the every getter checks if the current user is EmailAdmin or the creator of the E-Mail.

I undestand that you are pausing the PR right now, but at the moment I'm not willing to invest more time, since I don't know if it's worth it.

This simple user access concept may lead to problems with issue #9005, too. Since it might happen that another user (no email admin)  wants to check which mails have been sent already.

Issue #9681, opened 2020-10-31
AL Test Tool: "Update Test Methods" only on a selected/filtered number of test codeunits in the AL Test Tool


lvanvugt:
As reported on yammer (https://www.yammer.com/dynamicsnavdev/threads/925303939506176) I was wondering:

`How do I get to work **Update Test Methods** only on a selected/filtered number of test codeunits in the AL Test Tool?`

It seems to always process all test codeunits in a suite, which is a nuisance if I only want a single or a couple of test codeunits to be updated.

I would like to suggest the following change:

`"Update Test Methods" should only operate on the test codeunits in the filtered view. If there is no filter it will operate on all test codeunits.`

As per the suggestion of @nikolakukrika I will implement it myself and provide a pull request.

### Note
In (CAL) Test Tool this is done only for the selected test codeunit (line).

lvanvugt:
Hi @StanPesotskiy, (long-time-no-see ;-). Why is this closed and set to feature-request?

StanPesotskiy:
Hi Luc! It was the result of a clean up. I marked it as a Feature-request but requests with such tags do not go to our backlog.
But your request is not a regular one as you suggest a solution. I would expect that you get some review comments  instead of a thread here. Do you expect some activity from MS here?

lvanvugt:
Sure, I do expect activity. This thing is bugging me. ;-)

JesperSchulz:
Following up on this active issue. What exactly is the request here? I'm slightly confused :-)

lvanvugt:
The issue is that when I execute the **Update Test Methods** is updates _all_ test codeunits irrespective of having only selected one or put a filter on the suite. This is very annoying when I am working on tests and want to first work only on a subset of methods and later add the rest of the methods for one codeunit. If I recall right this is how the CAL Test Tool did do it.

JesperSchulz:
Got it. I just needed to make sure I understand, because of the "i will implement it myself" vs. feature request. We'll see what we can do!

lvanvugt:
Well, I did implement it and provided a pull request for that. Don't know what came of it eventually.

JesperSchulz:
We're going to get this fixed from our end :-) Shouldn't be too hard, so let's just get it done.

JesperSchulz:
Closing issue as "Won't Fix", as agreed upon with @nikolakukrika.

PR #9682, opened 2020-10-31
"Update Test Methods" works on filtered test codeunits #9681

lvanvugt:


Modules/DevTools/TestFramework/TestRunner/src/TestSuiteMgt.Codeunit.al
```diff
@@ -267,7 +267,6 @@ codeunit 130456 "Test Suite Mgt."
         TestRunnerGetMethods: Codeunit "Test Runner - Get Methods";
     begin
         BackupTestMethodLine.Copy(TestMethodLine);
-        TestMethodLine.Reset();
```
nikolakukrika, line 270: Hi Luc,

This will not work. We have a bug already: If you put a filter to the individual test method and select UpdateTestMethods action, it will load all methods except that method. That method will disappear.

Also the problem is in the behavior, I would not expect that only filtered list is updated.

Could we do this implementation:
Check if there is a single codeunit, if yes, then update all
If there is multiple, prompt the user to select if they only want to update all (default) or selected? (Something similar to Run Tests action) 
Then we can keep this code as it is and add new path to update selected methods codeunit.

nikolakukrika (changes requested):
See the comment, could we update it to the prompt to make it more discoverable?

Modules/DevTools/TestFramework/TestRunner/src/TestSuiteMgt.Codeunit.al
lvanvugt: I am totally fine with your suggestion. Thanx for the effort.

Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTLogEntry.Table.al
```diff
@@ -5,6 +5,7 @@
 table 149002 "BCPT Log Entry"
 {
     DataClassification = SystemMetadata;
+    DrillDownPageId = "BCPT Log Entries";
```
nikolakukrika, line 8: Is this change related?

lvanvugt: No, sorry, this should not be in the PR. See #9719. As described there I wanted to change another PR for that, but GitHub seems not to allow me. Or even worse: it incorporates it in the existing PR. Any tips on how to separate PRs are more than welcome.

stkillen: Hi @lvanvugt you should look into branches. I responded to your issue.

JesperSchulz:
This is odd! Somehow this PR was merged together with the other one! Something in your branch setup was not quite right it seems!

JesperSchulz:
Based on the code comments though, it seems like this PR should have gotten abandoned anyway :-) I hence won't reopen it. If I'm mistaken, please create a new PR and excuse my mistake!

lvanvugt:
I probably am to to source to all troubles here ;-). Need to learn this.

PR #9874, opened 2020-11-17
New procedure ToRecord in Codeunit "Temp Blob"

alapenas:
Hi,

We already have in the BLOB Storage module procedures as FromRecord(), FromRecordRef(), FromFieldRef(), but only 2 reverse procedure ToRecordRef() and ToFieldRef().

I have also updated other Modules with the new procedure.
Readme and TestCodeunit were updated too.

### DIFF
--- a/Apps/GB/UKMakingTaxDigital/app/src/Connection/MTDFraudPreventionMgt.Codeunit.al
+++ b/Apps/GB/UKMakingTaxDigital/app/src/Connection/MTDFraudPreventionMgt.Codeunit.al
@@ -97,7 +97,6 @@ codeunit 10541 "MTD Fraud Prevention Mgt."
     local procedure CollectDataForClient(var DueDateTime: DateTime; JsonFieldNo: Integer)
     var
         TempBlob: Codeunit "Temp Blob";
-        RecordRef: RecordRef;
         Expired: Boolean;
         OutStream: OutStream;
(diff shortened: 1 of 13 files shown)


IhorHandziuk:
Hi @alapenas,

Thank you for your contribution. Unfortunately, in AL language records cannot be passed as parameters of type `var Variant` (it's a long-standing limitation), this is the reason we did not implement the `ToRecord` method when developing the BLOB Storage module.

If you try to compile any of the first party extensions with your change, you will get an error similar to:
`error AL0133: Argument 1: cannot convert from 'Record Customer' to 'var Variant'`

It is, however, possible to assign a record to a variable of type `Variant` and then make the call. For example, the code:
```
    procedure Run();
    var
        Customer: Record Customer;
        CustomerVariant: Variant;
    begin
        CustomerVariant := Customer;
        ToRecord(CustomerVariant);
        Customer := CustomerVariant;
    end;

    procedure ToRecord(var RecordVariant: Variant)
    begin
        // do something with RecordVariant
    end;
```
will compile and work as expected, but such a construct will require the `ToRecord` call to be surrounded by the 2 assignments and has no advantage over using `ToRecordRef` similarly surrounded by 2 calls (`GetTable` and `SetTable`)

alapenas:
@IhorHandziuk thank you for you reply.

PR #10055, opened 2020-11-30
Get all upgrade tag and set skipped field

Bertverbeek4PS:
As discussed with Nikola Kukrika
We would like to get all the upgrade tags from a company and the database to compare if a upgrade tag is in the table or not.
Also we are going to use this not be part of the Automated Data Upgrade but in a seperate page.
That is why we want to use the skipped field.

### DIFF
--- a/Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
+++ b/Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
@@ -51,7 +51,7 @@ codeunit 9999 "Upgrade Tag"
     /// <summary>
     /// Sets the upgrade tag to skipped.
     /// </summary>
-    /// <param name="ExistingTag">Tag code to get the Tag</param>
+    /// <param name="ExistingTag">Tag code to set the skipped field</param>
     procedure SetSkippedUpgrade(ExistingTag: Code[250])
     begin
         UpgradeTagImpl.SetSkippedUpgrade(ExistingTag);
@@ -60,7 +60,8 @@ codeunit 9999 "Upgrade Tag"
     /// <summary>
     /// Sets the upgrade tag to skipped.
     /// </summary>
-    /// <param name="ExistingTag">Tag code to get the Tag</param>
+    /// <param name="ExistingTag">Tag code to set the skipped field</param>
+    /// <param name="TagCompanyName">Name of the company to check existance of tag</param>
     procedure SetSkippedUpgrade(ExistingTag: Code[250]; TagCompanyName: Code[30])
     begin
         UpgradeTagImpl.SetSkippedUpgrade(ExistingTag, TagCompanyName);
--- a/Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
+++ b/Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
@@ -49,6 +49,24 @@ codeunit 9999 "Upgrade Tag"
     end;
 
     /// <summary>
+    /// Sets the upgrade tag to skipped.
+    /// </summary>
+    /// <param name="ExistingTag">Tag code to get the Tag</param>
+    procedure SetSkippedUpgrade(ExistingTag: Code[250])
+    begin
+        UpgradeTagImpl.SetSkippedUpgrade(ExistingTag);
+    end;
+
+    /// <summary>
+    /// Sets the upgrade tag to skipped.
+    /// </summary>
+    /// <param name="ExistingTag">Tag code to get the Tag</param>
+    procedure SetSkippedUpgrade(ExistingTag: Code[250]; TagCompanyName: Code[30])
+    begin
+        UpgradeTagImpl.SetSkippedUpgrade(ExistingTag, TagCompanyName);
+    end;
+
+    /// <summary>
     /// This method should be used to set all upgrade tags in a new company. 
     /// The method is called from codeunit 2 - Company Initialize.
     /// </summary>
@@ -74,7 +92,7 @@ codeunit 9999 "Upgrade Tag"
     /// List of upgrade tags that should be inserted if they do not exist.
     /// </param>
     [IntegrationEvent(false, false)]
-    internal procedure OnGetPerCompanyUpgradeTags(var PerCompanyUpgradeTags: List of [Code[250]])
+    procedure OnGetPerCompanyUpgradeTags(var PerCompanyUpgradeTags: List of [Code[250]])
     begin
     end;
 
@@ -85,7 +103,7 @@ codeunit 9999 "Upgrade Tag"
     /// List of upgrade tags that should be inserted if they do not exist.
     /// </param>
     [IntegrationEvent(false, false)]
-    internal procedure OnGetPerDatabaseUpgradeTags(var PerDatabaseUpgradeTags: List of [Code[250]])
+    procedure OnGetPerDatabaseUpgradeTags(var PerDatabaseUpgradeTags: List of [Code[250]])
     begin
     end;
 }
--- a/Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
+++ b/Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
@@ -29,6 +29,30 @@ codeunit 9996 "Upgrade Tag Impl."
         SetUpgradeTagForCompany(NewTag, CopyStr(CompanyName(), 1, MaxStrLen(ConstUpgradeTags.Company)));
     end;
 
+    procedure SetSkippedUpgrade(ExistingTag: Code[250])
+    var
+        ConstUpgradeTags: Record "Upgrade Tags";
+    begin
+        ConstUpgradeTags.Reset;
+        ConstUpgradeTags.SetRange(Tag, ExistingTag);
+        if ConstUpgradeTags.FindSet then
+            repeat
+                ConstUpgradeTags.ModifyAll(ConstUpgradeTags."Skipped Upgrade", true);
+            until ConstUpgradeTags.next = 0;
+    end;
+
+    procedure SetSkippedUpgrade(ExistingTag: Code[250]; TagCompanyName: Code[30])
+    var
+        ConstUpgradeTags: Record "Upgrade Tags";
+    begin
+        ConstUpgradeTags.Reset;
+        ConstUpgradeTags.SetRange(Tag, ExistingTag);
+        ConstUpgradeTags.SetRange(Company, TagCompanyName);
+        if ConstUpgradeTags.FindSet then begin
+            ConstUpgradeTags.ModifyAll(ConstUpgradeTags."Skipped Upgrade", true);
+        end;
+    end;
+
     procedure SetAllUpgradeTags()
     var
         ConstUpgradeTags: Record "Upgrade Tags";


Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
```diff
@@ -74,7 +93,7 @@ codeunit 9999 "Upgrade Tag"
     /// List of upgrade tags that should be inserted if they do not exist.
     /// </param>
     [IntegrationEvent(false, false)]
-    internal procedure OnGetPerCompanyUpgradeTags(var PerCompanyUpgradeTags: List of [Code[250]])
+    procedure OnGetPerCompanyUpgradeTags(var PerCompanyUpgradeTags: List of [Code[250]])
```
nikolakukrika, line 96: Could you add a public method that returns the list of PerCompanyUpgradeTags to avoid exposing the event directly?
Then it will be visible that it is by design - that we want to have this within the interface. We can also control in which flows does this event get triggered.

Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
```diff
@@ -85,7 +104,7 @@ codeunit 9999 "Upgrade Tag"
     /// List of upgrade tags that should be inserted if they do not exist.
     /// </param>
     [IntegrationEvent(false, false)]
-    internal procedure OnGetPerDatabaseUpgradeTags(var PerDatabaseUpgradeTags: List of [Code[250]])
+    procedure OnGetPerDatabaseUpgradeTags(var PerDatabaseUpgradeTags: List of [Code[250]])
```
nikolakukrika, line 107: Same comment here. It would be nice to get a public method that calls the internal implementation.

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,25 @@ codeunit 9996 "Upgrade Tag Impl."
         SetUpgradeTagForCompany(NewTag, CopyStr(CompanyName(), 1, MaxStrLen(ConstUpgradeTags.Company)));
     end;
 
+    procedure SetSkippedUpgrade(ExistingTag: Code[250])
+    var
+        ConstUpgradeTags: Record "Upgrade Tags";
+    begin
+        ConstUpgradeTags.Reset;
```
nikolakukrika, line 36: Naming - I would drop Const, it does not bring value. UpgradeTags is good enough, alternative is ExistingUpgradeTags.

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,25 @@ codeunit 9996 "Upgrade Tag Impl."
...
+        ConstUpgradeTags.SetRange(Tag, ExistingTag);
+        ConstUpgradeTags.ModifyAll(ConstUpgradeTags."Skipped Upgrade", true);
+    end;
+
+    procedure SetSkippedUpgrade(ExistingTag: Code[250]; TagCompanyName: Code[30])
+    var
+        ConstUpgradeTags: Record "Upgrade Tags";
+    begin
+        ConstUpgradeTags.Reset;
```
nikolakukrika, line 45: Same comment here - Const should be removed.

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,25 @@ codeunit 9996 "Upgrade Tag Impl."
         SetUpgradeTagForCompany(NewTag, CopyStr(CompanyName(), 1, MaxStrLen(ConstUpgradeTags.Company)));
     end;
 
+    procedure SetSkippedUpgrade(ExistingTag: Code[250])
+    var
+        ConstUpgradeTags: Record "Upgrade Tags";
+    begin
+        ConstUpgradeTags.Reset;
+        ConstUpgradeTags.SetRange(Tag, ExistingTag);
```
nikolakukrika, line 37: I do not like this, it should operate in the context of the company. Otherwise we are risking that it is marking too many tags as skipped . Thus it should call the other method with CompanyName() so it is running in the given company, the existing logic behaves like this so it is good to align.

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,25 @@ codeunit 9996 "Upgrade Tag Impl."
         SetUpgradeTagForCompany(NewTag, CopyStr(CompanyName(), 1, MaxStrLen(ConstUpgradeTags.Company)));
     end;
 
+    procedure SetSkippedUpgrade(ExistingTag: Code[250])
```
nikolakukrika, line 32: I would add skippedUpgrade parameter so you can set it to true and to false if needed.

Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
```diff
@@ -48,6 +48,25 @@ codeunit 9999 "Upgrade Tag"
         UpgradeTagImpl.SetUpgradeTag(NewTag);
     end;
 
+    /// <summary>
+    /// Sets the upgrade tag to skipped.
+    /// </summary>
+    /// <param name="ExistingTag">Tag code to set the skipped field</param>
+    procedure SetSkippedUpgrade(ExistingTag: Code[250])
```
nikolakukrika, line 55: Consider adding GetSkippedUpgrade(ExistingTag: Code) : Boolean and GetSkippedUpgrade(ExistingTag: Code, CompanyName: Text): Boolean) so you can check..

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,25 @@ codeunit 9996 "Upgrade Tag Impl."
...
+    end;
+
+    procedure SetSkippedUpgrade(ExistingTag: Code[250]; TagCompanyName: Code[30])
+    var
+        ConstUpgradeTags: Record "Upgrade Tags";
+    begin
+        ConstUpgradeTags.Reset;
+        ConstUpgradeTags.SetRange(Tag, ExistingTag);
+        ConstUpgradeTags.SetRange(Company, TagCompanyName);
+        ConstUpgradeTags.ModifyAll(ConstUpgradeTags."Skipped Upgrade", true);
```
nikolakukrika, line 48: Minor (Feel free to reject) - I would recommend get and modify, we should update individual tags. The scenarios for updating all tags are rare.

nikolakukrika (changes requested):

Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
Bertverbeek4PS: procedure GetPerDatabaseUpgradeTags added

Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
Bertverbeek4PS: procedure GetPerCompanyUpgradeTags added

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
Bertverbeek4PS: removed Const

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
Bertverbeek4PS: Removed Const

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
Bertverbeek4PS: Added the parameter SkipUpgrade

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
Bertverbeek4PS: Change to Get en Modify

Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
Bertverbeek4PS: Added boolean

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
Bertverbeek4PS: Added the companyName() and call the other function

Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
```diff
@@ -51,20 +51,22 @@ codeunit 9999 "Upgrade Tag"
...
 
     /// <summary>
     /// Sets the upgrade tag to skipped.
     /// </summary>
-    /// <param name="ExistingTag">Tag code to set the skipped field</param>
+    /// <param name="ExistingTag">Tag code to set the Skipped Upgrade field</param>
     /// <param name="TagCompanyName">Name of the company to check existance of tag</param>
-    procedure SetSkippedUpgrade(ExistingTag: Code[250]; TagCompanyName: Code[30])
+    /// <param name="SkipUpgrade">Sets the Skipped Upgrade field</param>
+    procedure SetSkippedUpgrade(ExistingTag: Code[250]; TagCompanyName: Code[30]; SkipUpgrade: Boolean): Boolean
```
nikolakukrika, line 67: Could we remove this Boolean, it looks confusing? 
It would be better to have a dedicated function to check if the tag was skipped, it would be easier to use the API.

Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
```diff
@@ -86,14 +88,30 @@ codeunit 9999 "Upgrade Tag"
         UpgradeTagImpl.SetAllUpgradeTags(NewCompanyName);
     end;
 
+    /// <summary>
+    /// With this method you get all the upgrade tags by company in a list. 
+    /// </summary>
+    procedure GetPerCompanyUpgradeTags(var PerCompanyUpgradeTags: List of [Code[250]])
```
nikolakukrika, line 94: Double check that all the signatures list the parameters.

Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
Bertverbeek4PS: I have removed the boolean and added the function HasUpgradeTagSkipped
But not sure if SetSkippedUpgrade in Upgrade Tag Impl. must also return boolean then.

Modules/System/Upgrade Tags/src/UpgradeTag.Codeunit.al
Bertverbeek4PS: Add the comment for the parameter. Hope this is what you mean

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,31 @@ codeunit 9996 "Upgrade Tag Impl."
...
+    begin
+        exit(SetSkippedUpgrade(ExistingTag, CompanyName(), SkipUpgrade));
+    end;
+
+    procedure SetSkippedUpgrade(ExistingTag: Code[250]; TagCompanyName: Code[30]; SkipUpgrade: Boolean): Boolean
+    var
+        UpgradeTags: Record "Upgrade Tags";
+    begin
+        if UpgradeTags.Get(ExistingTag, TagCompanyName) then begin
+            UpgradeTags."Skipped Upgrade" := SkipUpgrade;
```
nikolakukrika, line 42: It would be good to check if value is different so we don't fire blank modify statements

Modules/System/Upgrade Tags/.vscode/.alcache/cache_61d28fd385c2d3ef0f4b164b0b92ea85727be4ed33719c27758751358dd70b67.json
```diff
@@ -0,0 +1 @@
...dex":2,"SymbolZipPath":""},"Scope":"Extension"},{"TypeId":10,"Type":"ControlAddIn","Id":"","Name":"Microsoft.Dynamics.Nav.Client.SocialListening","TargetObject":"","TargetObjectType":"ControlAddIn","Publisher":"Microsoft","Application":"System","Version":"17.0.19123.0","CanExecute":false,"CanDesign":false,"CanCreatePage":false,"FsPath":"","EventName":"not_an_event","FieldName":"","SymbolData":{"Path":"c:\\Users\\bverbeek\\Documents\\AL\\ALAppExtensions\\Modules\\System\\Upgrade Tags\\.alpackages\\Microsoft_System_17.0.19123.0.app","Type":"ControlAddIns","Index":3,"SymbolZipPath":""},"Scope":"Extension"},{"TypeId":10,"Type":"ControlAddIn","Id":"","Name":"Microsoft.Dynamics.Nav.Client.VideoPlayer","TargetObject":"","TargetObjectType":"ControlAddIn","Publisher":"Microsoft","Application":"System","Version":"17.0.19123.0","CanExecute":false,"CanDesign":false,"CanCreatePage":false,"FsPath":"","EventName":"not_an_event","FieldName":"","SymbolData":{"Path":"c:\\Users\\bverbeek\\Documents\\AL\\ALAppExtensions\\Modules\\System\\Upgrade Tags\\.alpackages\\Microsoft_System_17.0.19123.0.app","Type":"ControlAddIns","Index":4,"SymbolZipPath":""},"Scope":"Extension"}],"CreatedwithVersion":"0.2.4"}
```
nikolakukrika, line 1: Don't check in these files

Modules/System/Upgrade Tags/.vscode/.alcache/cache_61d28fd385c2d3ef0f4b164b0b92ea85727be4ed33719c27758751358dd70b67events.json
```diff
@@ -0,0 +1 @@
...":"","TargetElementName":"","EventParameters":[{"IsVar":true,"Name":"Version","TypeDefinition":{"Name":"Text[248]","Temporary":false}}],"FieldName":"","SymbolData":null},{"TypeId":3,"Type":"Codeunit","Id":2000000001,"Name":"Version Triggers","TargetObject":"","TargetObjectType":"","Publisher":"Microsoft","Application":"System","Version":"17.0.19123.0","CanExecute":false,"CanDesign":false,"FsPath":"","EventName":"GetReleaseVersion","EventType":"BusinessEvent","EventPublisher":true,"TargetEventName":"","TargetElementName":"","EventParameters":[{"IsVar":true,"Name":"Version","TypeDefinition":{"Name":"Text[50]","Temporary":false}}],"FieldName":"","SymbolData":null},{"TypeId":3,"Type":"Codeunit","Id":2000000001,"Name":"Version Triggers","TargetObject":"","TargetObjectType":"","Publisher":"Microsoft","Application":"System","Version":"17.0.19123.0","CanExecute":false,"CanDesign":false,"FsPath":"","EventName":"GetApplicationBuild","EventType":"BusinessEvent","EventPublisher":true,"TargetEventName":"","TargetElementName":"","EventParameters":[{"IsVar":true,"Name":"Build","TypeDefinition":{"Name":"Text[80]","Temporary":false}}],"FieldName":"","SymbolData":null}],"CreatedwithVersion":"0.2.4"}
```
nikolakukrika, line 1: Don't check in these files

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,31 @@ codeunit 9996 "Upgrade Tag Impl."
...
+            exit(false);
+    end;
+
+    procedure HasUpgradeTagSkipped(ExistingTag: Code[250]; TagCompanyName: Code[30]): Boolean
+    var
+        UpgradeTags: Record "Upgrade Tags";
+    begin
+        if UpgradeTags.Get(ExistingTag, TagCompanyName) then
+            exit(UpgradeTags."Skipped Upgrade");
+    end;
```
nikolakukrika, line 55: Minor - add exit(false) so it is clear that the function should return false and it was not forgotten. It is good to have a clear intent

nikolakukrika (approved):

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
Bertverbeek4PS: Added the condition

Modules/System/Upgrade Tags/.vscode/.alcache/cache_61d28fd385c2d3ef0f4b164b0b92ea85727be4ed33719c27758751358dd70b67.json
Bertverbeek4PS: Sorry did a commit all :(

Modules/System/Upgrade Tags/.vscode/.alcache/cache_61d28fd385c2d3ef0f4b164b0b92ea85727be4ed33719c27758751358dd70b67events.json
Bertverbeek4PS: Sorry did a commit all :(

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
Bertverbeek4PS: Added exit false

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
nikolakukrika, line 42: If is missing here. Did you push?

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
nikolakukrika, line 55: exit false is missing here - did you push?

nikolakukrika (approved):

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,36 @@ codeunit 9996 "Upgrade Tag Impl."
...
+    var
+        UpgradeTags: Record "Upgrade Tags";
+    begin
+        if UpgradeTags.Get(ExistingTag, TagCompanyName) then begin
+            if UpgradeTags."Skipped Upgrade" <> SkipUpgrade then begin
+                UpgradeTags."Skipped Upgrade" := SkipUpgrade;
+                UpgradeTags.Modify();
+                exit(true);
+            end else
+                exit(false);
```
nikolakukrika, line 47: Minor comment here - I would return true in both cases since the value was set. False should be if the record does not exist.  Or we could remove the return value for now since it is not used?

nikolakukrika (approved):

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
Bertverbeek4PS: Removed indeed the return value

nikolakukrika (approved):

JesperSchulz:
Internal code review process started.

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,33 @@ codeunit 9996 "Upgrade Tag Impl."
...
+    procedure SetSkippedUpgrade(ExistingTag: Code[250]; SkipUpgrade: Boolean)
+    begin
+        SetSkippedUpgrade(ExistingTag, CompanyName(), SkipUpgrade);
+    end;
+
+    procedure SetSkippedUpgrade(ExistingTag: Code[250]; TagCompanyName: Code[30]; SkipUpgrade: Boolean)
+    var
+        UpgradeTags: Record "Upgrade Tags";
+    begin
+        if UpgradeTags.Get(ExistingTag, TagCompanyName) then begin
```
JesperSchulz, line 41: Only use begin for compound statements.

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
```diff
@@ -29,6 +29,33 @@ codeunit 9996 "Upgrade Tag Impl."
         SetUpgradeTagForCompany(NewTag, CopyStr(CompanyName(), 1, MaxStrLen(ConstUpgradeTags.Company)));
     end;
 
+    procedure SetSkippedUpgrade(ExistingTag: Code[250]; SkipUpgrade: Boolean)
+    begin
+        SetSkippedUpgrade(ExistingTag, CompanyName(), SkipUpgrade);
```
JesperSchulz, line 34: CompanyName is Text, parameter is Code[30]. Possible overflow.

JesperSchulz (approved):
Approved. I will fix the minor issues in our internal pack we're submitting.

JesperSchulz (changes requested):
Code coverage analysis of build system returned 0%. Tests are missing.

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
JesperSchulz: No need for exit false. False is the default.

Bertverbeek4PS:
Thanks Jesper! Good luck with it.

Bertverbeek4PS:
Sorry Jesper. I want it on a second pull reqeust for a new function SetDatabaseUpgradeTag.
Because if we are using this function the companyname will be empty.
In this case we can do also the database upgrades in a seperate page without concerns the Automatic Data Upgrade.
So in MS SaaS the move to another version is always successfull.

Bertverbeek4PS:
> Code coverage analysis of build system returned 0%. Tests are missing.

@JesperSchulz  Should I create tests for those functions? Or in this case it isn't a problem?

JesperSchulz:
It would be great if you could add some tests to get coverage of the added functions. Once completed, we can proceed with this PR.

Modules/System/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
Bertverbeek4PS: Changed

Bertverbeek4PS:
@JesperSchulz added some test functions to get coverage and two small fixes.

JesperSchulz (approved):

JesperSchulz:
Changes will go out with 17.4. The changes may appear to have gotten reverted with the upcoming 17.3 update here on GitHub, but that's just because I just missed that release train :-) The changes will be back.

JesperSchulz:

PR #10187, opened 2020-12-09
Fix to show caption instead of name when choosing the Email Scenario

mynavblog:
When choosing the Email Scenario then instead of Caption the Name of Enum values is shown.

Before the change (names with prefix XXX):
image

and after then change (captions):
image

### DIFF
--- a/Modules/System/Email/src/Scenario/EmailScenarioImpl.Codeunit.al
+++ b/Modules/System/Email/src/Scenario/EmailScenarioImpl.Codeunit.al
@@ -209,6 +209,7 @@ codeunit 8892 "Email Scenario Impl."
     procedure GetAvailableScenariosForAccount(EmailAccount: Record "Email Account Scenario"; var EmailScenarios: Record "Email Account Scenario")
     var
         Scenario: Record "Email Scenario";
+        EmailScenario: Enum "Email Scenario";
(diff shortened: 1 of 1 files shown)


JesperSchulz:
Thank you for this suggestion. We'll be on it right away.

JesperSchulz (approved):

Modules/System/Email/src/Scenario/EmailScenarioImpl.Codeunit.al
```diff
@@ -209,6 +209,7 @@ codeunit 8892 "Email Scenario Impl."
     procedure GetAvailableScenariosForAccount(EmailAccount: Record "Email Account Scenario"; var EmailScenarios: Record "Email Account Scenario")
     var
         Scenario: Record "Email Scenario";
+        EmailScenario: Enum "Email Scenario";
```
JesperSchulz, line 212: We're probably going to do this change without the variable...

mynavblog: Great :) I just did not know how :)

JesperSchulz:

JesperSchulz:
Availability update: We will publish a fix for this issue in the next update for releasee 17.

PR #10476, opened 2021-01-07
Show Dialogs in some TestRunner long-running actions

JavierFuentes:
I have added a dialog box to the Get and Update Test actions of the Test Runner Module because I have noticed that when you select all the codeunits it takes a long time and the system is completely blocked without the user being able to know what is really happening.

I hope you find it useful to do the merge.

image

### DIFF
--- a/Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
+++ b/Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
@@ -194,7 +194,7 @@ page 130451 "AL Test Tool"
                     PromotedCategory = Process;
                     PromotedIsBig = true;
                     PromotedOnly = true;
-                    ToolTip = 'Runs selected tests.';
+                    ToolTip = 'Runs marked tests.';
 
                     trigger OnAction()
                     var
@@ -215,6 +215,7 @@ page 130451 "AL Test Tool"
                     PromotedCategory = Process;
                     PromotedIsBig = true;
                     PromotedOnly = true;
+                    ToolTip = 'Runs selected tests.';
 
                     trigger OnAction()
                     var
@@ -303,7 +304,13 @@ page 130451 "AL Test Tool"
                     var
                         TestMethodLine: Record "Test Method Line";
                         TestSuiteMgt: Codeunit "Test Suite Mgt.";
+                        DeleteQuestion: Label 'Are you sure you want to Delete the Selected Lines?';
                     begin
+                        if GuiAllowed() then begin
+                            if not Confirm(DeleteQuestion, false) then
+                                exit;
+                        end;
+
                         CurrPage.SetSelectionFilter(TestMethodLine);
                         TestMethodLine.DeleteAll(true);
                         TestSuiteMgt.CalcTestResults(Rec, Success, Failure, Skipped, NotExecuted);
--- a/Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
+++ b/Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
@@ -304,7 +304,6 @@ page 130451 "AL Test Tool"
                     var
                         TestMethodLine: Record "Test Method Line";
                         TestSuiteMgt: Codeunit "Test Suite Mgt.";
-                        DeleteQuestion: Label 'Are you sure you want to Delete the Selected Lines?';
                     begin
                         if GuiAllowed() then begin
                             if not Confirm(DeleteQuestion, false) then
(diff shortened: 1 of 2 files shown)


JesperSchulz:
Thanks a lot! That indeed sounds like a very nice addition :-) We'll take a look for sure!

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
```diff
@@ -303,12 +304,79 @@ page 130451 "AL Test Tool"
...
+                                exit;
+                        end;
+
                         CurrPage.SetSelectionFilter(TestMethodLine);
                         TestMethodLine.DeleteAll(true);
                         TestSuiteMgt.CalcTestResults(Rec, Success, Failure, Skipped, NotExecuted);
                     end;
                 }
+
+                action(SwitchRunCheckLines)
```
JesperSchulz, line 320: Invert is probably the better word.

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
```diff
@@ -303,12 +304,79 @@ page 130451 "AL Test Tool"
...
+                    Promoted = true;
+                    PromotedCategory = Process;
+                    PromotedIsBig = true;
+                    PromotedOnly = true;
+                    ToolTip = 'Switch Run Check on selected lines.';
+
+                    trigger OnAction()
+                    var
+                        TestMethodLine: Record "Test Method Line";
+                        SwitchQuestion: Label 'Are you sure you want to %1 the Selected Lines?';
```
JesperSchulz, line 334: Placeholders for individual words are very hard to translate. The test runner isn't translated, so here it would be ok, but this is generally considered an anti-pattern.

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
```diff
@@ -303,12 +304,79 @@ page 130451 "AL Test Tool"
...
+                        Check: Label 'Check';
+                        Uncheck: Label 'Uncheck';
+                        NewRunValue: Boolean;
+                        ActionToDo: Text[30];
+                        DialogContent: Label '-- Switch Run Check --\#1#\#2#';
+                        i: Integer;
```
JesperSchulz, line 340: No need to use such short variable names any more in times of intellisense etc. The more descriptive a variable name is, better.

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
```diff
@@ -194,7 +194,7 @@ page 130451 "AL Test Tool"
                     PromotedCategory = Process;
                     PromotedIsBig = true;
                     PromotedOnly = true;
-                    ToolTip = 'Runs selected tests.';
+                    ToolTip = 'Runs marked tests.';
```
nikolakukrika, line 197: Nice spot. I would drop marked since it is a synonym with selected. So just runs tests or Runs tests from the suite.

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
```diff
@@ -303,7 +304,13 @@ page 130451 "AL Test Tool"
                     var
                         TestMethodLine: Record "Test Method Line";
                         TestSuiteMgt: Codeunit "Test Suite Mgt.";
+                        DeleteQuestion: Label 'Are you sure you want to Delete the Selected Lines?';
```
nikolakukrika, line 307: Minor - move to global variables since it will be translated.

Modules/DevTools/TestFramework/TestRunner/src/TestSuiteMgt.Codeunit.al
```diff
@@ -250,34 +250,73 @@ codeunit 130456 "Test Suite Mgt."
     procedure GetTestMethods(var ALTestSuite: Record "AL Test Suite"; var AllObjWithCaption: Record AllObjWithCaption)
     var
         TestLineNo: Integer;
+        DialogContent: Label '-- Get Test Methods --\#1#\#2#';
```
nikolakukrika, line 253: Minor - Getting Test Methods

Modules/DevTools/TestFramework/TestRunner/src/TestSuiteMgt.Codeunit.al
```diff
@@ -250,34 +250,73 @@ codeunit 130456 "Test Suite Mgt."
...
+        i: Integer;
```
nikolakukrika, line 254: Also minor - we use capital naming convention (PascalCase), thus it should be I and full names for variables, n -> TotalNumberOfTests or AllObjWithCaptionCount and etc., d - ProgressWindow or similar naming we have in the product.

nikolakukrika (changes requested):
Thanks a lot for the contribution, it is a great addition. I have left minor comments, the ones regarding the naming should be fixed.

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
JavierFuentes: Do you mean __InvertRunCheckLines__?

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
JavierFuentes: Do you mean that if all these new Labels are defined as Local they won't be translated?

Do you want me to move all of them, including the DialogContent?

Modules/DevTools/TestFramework/TestRunner/src/TestSuiteMgt.Codeunit.al
JavierFuentes: OK, I will do this change in all new DialogContent Labels, including the change from Switching to Inverting as @JesperSchulz  has proposed above.

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
nikolakukrika: Yes all of them -  They may be translated, however I'm not 100% certain if the internal tooling will pick it up. We add these as globals in all code, thus it is good to align.

Modules/DevTools/TestFramework/TestRunner/src/TestSuiteMgt.Codeunit.al
JavierFuentes: OK, I will do this change in all new DialogContent Labels, including the change from Switching to Inverting as @JesperSchulz has proposed

Modules/DevTools/TestFramework/TestRunner/src/TestSuiteMgt.Codeunit.al
JavierFuentes: I understand the style rules, but I try to avoid unnecessary verbosity... 

If you all see it correctly, I will rename these variables as follows using more agnostic words:
i -> Counter
n -> TotalCount
d -> Window (already used in standard code too)

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
JavierFuentes: For an end user with prior Navision / BC experience, selecting and marking has always been considered different actions in the commonly used language ...

That is why I wanted to differentiate between the two actions by using each word in its context.

JavierFuentes: (sorry for my English, it is not so good as I would like)

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
JavierFuentes: OK, I will make this change... thanks...

Modules/DevTools/TestFramework/TestRunner/src/ALTestTool.Page.al
```diff
@@ -304,11 +305,72 @@ page 130451 "AL Test Tool"
...
+                    ApplicationArea = All;
+                    Caption = '&Invert Run Check';
+                    Image = Change;
+                    Promoted = true;
+                    PromotedCategory = Process;
+                    PromotedIsBig = true;
+                    PromotedOnly = true;
+                    ToolTip = 'Invert Run Check on selected lines.';
+
+                    trigger OnAction()
```
JesperSchulz, line 329: I think all of this can be done with and handful of lines of code ;-)
        
CurrPage.SetSelectionFilter(TestMethodLine);

if TestMethodLine.FindSet(true, false) then
    repeat
        TestMethodLine.Validate(Run, not TestMethodLine.Run);
        TestMethodLine.Modify(true);
    until TestMethodLine.Next() = 0;

There is no need to ask questions or show status - this is a low risk, fast operation. You don't need to fix this. I'll do that with my final check-in. Just wanted to update you :-)

JesperSchulz:
We're going to implement this code for the 18.0 release in November. The code will appear "reverted" when I update the the repo to 17.4, but your changes will return, when 18.0 is pushed out. I hope that is ok!
Thanks for the contribution!

JesperSchulz (approved):

JesperSchulz:
Short update: the progress dialog on _GetTestMethods()_ had to get reverted. It is causing our build system to become instable. We're working on figuring out the root cause, but for now, that progress dialog won't be in the product.

JesperSchulz:

PR #10577, opened 2021-01-13
Convert Text Encoding

angela2389:
Hi Microsoft team,

Can you please process this pull request?


JesperSchulz:
Excellent :) Great to see how quickly you are able to create a new module now. 5 hours ago I suggested that you create a module, and here it is. We'll be on it right away!

Modules/System/Convert/app.json
```diff
@@ -0,0 +1,33 @@
+{
+    "id":  "ad94bc69-99fa-4bf0-85b3-6dbc72bd1543",
+    "name":  "Convert",
```
JesperSchulz, line 3: That is a very generic name you chose :-) If someone wants to build a picture converter, should that get added to this module? Or currency conversion? I think this module would get too wide.
I think we should try to find a module name, which is specific, yet broad enough to fit more functions than the convert function of the encoding class. Finding the right balance... That is not easy, when all we've got is a single function you'd like to add :-D
So, Encoding is an option... Then the module could hold all functionality which concerns encoding. But again, maybe a little broad? "String Encoding" - now it's narrowed down. Maybe too narrow? "String Manipulation" is another option; it could turn into a module, which has all sorts of functions to manipulate strings.
I think I need to give this a little thought. If you have any good ideas, just shoot :-)

angela2389: String Converter? Or should it have DotNet in front of it, to make it more specific? 
DotNet String Manipulation? 
DotNet String Converter?

JesperSchulz: Definitely not anything with DotNet in front of it. Whether you internally in the module use .NET, AL or something else entirely, should not matter to the module itself at all. The module simply does, what it's API specifies; and the API won't give away, that .NET is used.
I'll chew on it a day :-)

angela2389: Hi Jesper,

Did you give it some thought? Is the rest of the Pull request approved? :)

JesperSchulz: The code as such is fine :-) I will work on the module name - no action item for you. We do have 2 more weeks, before we cut the next round of minor releases. I'll work it out in time to catch that release train!

m-zhelezova: What about calling the module _Encoding_?

Modules/System/Convert/src/Convert.Codeunit.al
```diff
@@ -0,0 +1,26 @@
...
+    var
+        ConvertImpl: Codeunit "Convert Impl.";
+
+    /// <summary>
+    /// Converts a text from one encoding to another.
+    /// </summary>
+    /// <param name="SrcCodepage">Code page identifier of the source.</param>
+    /// <param name="DstCodepage">Code page identifier of the output.</param>
+    /// <param name="Text">The text containing the characters to encode.</param>
+    procedure ConvertTextEncoding(SrcCodepage: Integer; DstCodepage: Integer; Text: Text): Text
```
m-zhelezova, line 22: Maybe an overload with `TextEncoding`?

I can now see that AL's `TextEncoding` isn't quite the same as .Net's `Encoding`. So maybe new enum?

Modules/System/Convert/src/Convert.Codeunit.al
```diff
@@ -0,0 +1,26 @@
...
+
+    var
+        ConvertImpl: Codeunit "Convert Impl.";
+
+    /// <summary>
+    /// Converts a text from one encoding to another.
+    /// </summary>
+    /// <param name="SrcCodepage">Code page identifier of the source.</param>
+    /// <param name="DstCodepage">Code page identifier of the output.</param>
+    /// <param name="Text">The text containing the characters to encode.</param>
```
m-zhelezova, line 21:
```suggestion
    /// <param name="Text">The text containing the characters to convert.</param>
```

Modules/System/Convert/src/Convert.Codeunit.al
m-zhelezova, line 22:
```suggestion
    procedure Convert(SrcCodepage: Integer; DstCodepage: Integer; Text: Text): Text
```

Modules/System/Convert/src/ConvertImpl.Codeunit.al
```diff
@@ -0,0 +1,25 @@
...
+
+    procedure ConvertTextEncoding(SrcCodepage: Integer; DstCodepage: Integer; Text: Text) ConvertedText: Text
+    var
+        Encoding: DotNet Encoding;
+        SrcEncoding: DotNet Encoding;
+        DstEncoding: DotNet Encoding;
+        SrcBytes: DotNet Array;
+        DstBytes: DotNet Array;
+    begin
+        SrcEncoding := Encoding.GetEncoding(SrcCodepage);
```
m-zhelezova, line 18: Parameter validation is needed here :)

What if `SrcCodepage` is -1?

Modules/System Tests/Convert/src/ConvertTest.Codeunit.al
```diff
@@ -0,0 +1,30 @@
...
+        Convert: Codeunit Convert;
+        TextToConvert: Text;
+        ConvertedText: Text;
+        ExpectedText: Text;
+    begin
+        // [GIVEN] Assign text to convert
+        TextToConvert := 'Unidoce èëüöï$æôà&. ÐÊÏp $ € A';
+        ExpectedText := 'Unidoce èëüöï$æôà&. ÐÊÏp $ ? A';
+
+        // [WHEN] Get XmlDocument to text
```
m-zhelezova, line 24: Comment is off

Modules/System Tests/Convert/src/ConvertTest.Codeunit.al
```diff
@@ -0,0 +1,30 @@
...
+        ConvertedText := Convert.ConvertTextEncoding(65001, 28591, TextToConvert);
```
m-zhelezova, line 25: UTF8 looks better than 65001, doesn't it? ;)

m-zhelezova:
Awesome effort, @angela2389 :)

Just a few more things and we will ready to merge it!

Modules/System/Convert/src/ConvertImpl.Codeunit.al
angela2389: If it is -1 than an error appears: 
image

Isn't that what you would expect? If you would exit the code you won't know that the text is not encoded...

Modules/System/Convert/app.json
JesperSchulz: Ok, let's go with Encoding as module name.

Modules/System/Convert/src/Convert.Codeunit.al
m-zhelezova: It is more of a suggestion than a change request. It's a minor thing so don't bother.

angela2389:
Hi @JesperSchulz and @m-zhelezova, is the module fine now? Can you go ahead with merging?

JesperSchulz:

JesperSchulz:

JesperSchulz:
Availability update: We will publish a fix for this issue in the next update for release 16.

JesperSchulz:
As you can see, the module went in into version 18, 17 and 16 within the last 24 hours. Case closed :-)

JesperSchulz:
Reopening, so that I can merge.

JesperSchulz (approved):

PR #10618, opened 2021-01-18
Added BCPTLine in function as var

Bertverbeek4PS:
Added the "BCPT Line" record into the functions GetDefaultParameters and ValidateParameters.
In this way you can define parameters per test codeunit.
Added the functions 2 times because of any breaking changes. But please advice if this is correct.
@BardurKnudsen

### DIFF
--- a/Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTTestParamProvider.Interface.al
+++ b/Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTTestParamProvider.Interface.al
@@ -1,6 +1,6 @@
 interface "BCPT Test Param. Provider"
 {
-    procedure GetDefaultParameters(): Text[1000];
+    procedure GetDefaultParameters(BCPTLine: Record "BCPT Line"): Text[1000];
 
-    procedure ValidateParameters(Params: Text[1000]);
+    procedure ValidateParameters(BCPTLine: Record "BCPT Line"; Params: Text[1000]);
 }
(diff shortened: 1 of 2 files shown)


Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTTestParamProvider.Interface.al
```diff
@@ -1,6 +1,6 @@
 interface "BCPT Test Param. Provider"
 {
-    procedure GetDefaultParameters(): Text[1000];
+    procedure GetDefaultParameters(BCPTLine: Record "BCPT Line"): Text[1000];
```
ChethanT, line 4: This interface will be implemented by the test codeunit and does not have any relation to the BCPTLine.

Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTTestParamProvider.Interface.al
```diff
@@ -1,6 +1,6 @@
...
 
-    procedure ValidateParameters(Params: Text[1000]);
+    procedure ValidateParameters(BCPTLine: Record "BCPT Line"; Params: Text[1000]);
```
ChethanT, line 7: Same as above

ChethanT:
Function GetDefaultParameters is used during design time of the suite to fill in the default parameters defined by the test codeunit. The test codeunit that supports parameters implement this function to return the defaults and does not have any relation to BCPTLine object. 
Similarly. function ValidateParameters is used during design time to validate the parameters that has been set on the BCPTLine. The test codeunit that supports parameters implement this function to take Params set on the BCPTLine and validate if all the required parameters are filled. There is no need to passing BCPTLine as long as we are passing Params from that line.
 
During runtime, it’s the GetParameters and GetParameter function that is used to read the parameters from the current BCPTLine.

@BardurKnudsen will work with @Bertverbeek4PS to understand the need for this change and provide an alternate solution.

JesperSchulz (changes requested):
Please do changes as suggested by Chethan.

JesperSchulz:
Closing PR, as this solution could not be accepted. Please open new PR, when a valid solution has been found.

PR #10681, opened 2021-01-25
Continuous integration for System Application

IhorHandziuk:
The number of PRs to System Application is growing substantially and it's becoming increasingly more difficult to check the code manually all the time.

Introducing CI for System Application will help to improve the time to take in new changes.


mazhelez (approved):
Love it!

PR #11158, opened 2021-02-19
Move dimensions before lines

ajkauffmann:
Dimensions are placed after the lines in sales and purchase document APIs. Because of that, a deep insert with dimensions and lines will create the lines first and then the dimensions on the document. As a result, the lines do not inherit the dimensions from the header, while that is default behavior.

When inserting lines individually this works as expected. But the order in which lines and dimensions are defined in the API prevents this for deep inserts. The only workaround is to include the dimensions on the individual lines as well.

Moving up the dimensions before the lines solves this issue.

See also this thread on Yammer: https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=1085532101779456


nikolakukrika (approved):

JesperSchulz (approved):

JesperSchulz:

JesperSchulz:

Issue #7551, opened 2020-06-25
New procedure in Codeunit 3024 DotNet_URI 


MattiLE:
Hi al-team,

I think it would be helpful if there was a "URL combine" function in the Base App. 
With the Uri class it is possible to realize something like this.
It would be great if you could add this or a similar function to the DotNet_URI codeunit.

```
procedure CombineUri(BaseUri: Text; ToCombineUri: Text): Text;
    var
        DotNetUri2: DotNet Uri;
    begin
        DotNetUri := DotNetUri.Uri(DotNetUri2.Uri(BaseUri),ToCombineUri);
        exit(DotNetUri.ToString());
    end;
```

mazhelez:
Hey @MattiLE 

It's indeed a great idea 👍  However, it looks like this functionality is more suitable for the _System Application_. 
How do you feel about contributing by making a module that handles the .Net URI class?

Sure, most of the functionality already exists in the _Base Application_, but there's more that can added (`CombineUri` as a lead example).

mazhelez:
@MattiLE there's a new URI module on the way in the new release (17.0)

Since the requested functionality isn't part of the .Net class itself (it's more like a helper function), it won't be included in the module. However, if you think there's value for the AL developers, you can still contribute by adding it!

MattiLE:
Hi @mazhelez, thanks for the feedback.
I will take a look at the new module when it is released and then contribute an addition if necessary. :)

mazhelez:
Sounds great!

I'll now close the issue. Feel free to reopen it if needed.

PR #11172, opened 2021-02-19
Add Escape Uri String

MattiLE:
Hi al-team, 

I have added one more function to the Uri codeunit.
Actually only the C# function EscapeUriString is called. https://docs.microsoft.com/en-us/dotnet/api/system.uri.escapeuristring?view=net-5.0

Can you include the customization in the base version?

### DIFF
--- a/Modules/System/URI/src/Uri.Codeunit.al
+++ b/Modules/System/URI/src/Uri.Codeunit.al
@@ -129,6 +129,17 @@ codeunit 3060 Uri
     end;
 
     /// <summary>
+    /// Converts a uri string to its escaped representation.
+    /// </summary>
+    /// <remarks>Visit https://docs.microsoft.com/en-us/dotnet/api/system.uri.escapeuristring for more information.</remarks>
+    /// <param name="TextToEscape"></param>
+    /// <returns>A string that contains the escaped representation of <paramref name="TextToEscape:"/>.</returns>
+    procedure EscapeUriString(TextToEscape: Text): Text
+    begin
+        exit(Uri.EscapeUriString(TextToEscape));
+    end;
+
+    /// <summary>
     /// Checks if the provded string is a valid URI.
     /// </summary>
     /// <param name="UriString">The string to check.</param>


bjarkihall:
Usually you'd want to use EscapeDataString, which Uri codeunit already provides:
https://github.com/microsoft/ALAppExtensions/blob/2109d868be60047a3db66b223e453239add5c269/Modules/System/URI/src/Uri.Codeunit.al#L115

Here's the discussion about it:
https://stackoverflow.com/questions/4396598/whats-the-difference-between-escapeuristring-and-escapedatastring

I'm a bit curious about the motivation for the PR as I think the difference between the two procedures is pretty confusing.

Modules/System/URI/src/Uri.Codeunit.al
```diff
@@ -128,6 +128,17 @@ codeunit 3060 Uri
         exit(Uri.UnescapeDataString(TextToUnescape));
     end;
 
+    /// <summary>
+    /// Converts a uri string to its escaped representation.
+    /// </summary>
+    /// <remarks>Visit https://docs.microsoft.com/en-us/dotnet/api/system.uri.escapeuristring for more information.</remarks>
+    /// <param name="TextToEscape"></param>
```
mazhelez, line 135: Doc is missing.

Modules/System/URI/src/Uri.Codeunit.al
```diff
@@ -128,6 +128,17 @@ codeunit 3060 Uri
...
+    /// <returns>A string that contains the escaped representation of <paramref name="TextToEscape:"/>.</returns>
```
mazhelez, line 136:
```suggestion
    /// <returns>A string that contains the escaped representation of <paramref name="TextToEscape"/>.</returns>
```

mazhelez (changes requested):
Looks okay.

A couple of minor things need fixing. Maybe a test or two would be great.

MattiLE:
Hi @bjarkihall, background of this PR is that i have created a URL Combine function. A URL is to be built from separate parts.
At the end of the function, all characters that do not belong to a URL should be escaped. Therefore, if the URL does not contain "%", I can simply use the EscpaeURIString method. https://stackoverflow.com/a/47636037/62600

However, you are right that it can be confusing if both functions are included.

mazhelez:
Hey @MattiLE 

Do you plan on making another iteration? We'd love to uptake your PR, however, there're still comments that need to be fixed.

MattiLE:
Hi @mazhelez, 

it sounded to me that PR was being rejected. I might have misunderstood @bjarkihall.
I will now implement the annotations :)

bjarkihall:
Sorry @MattiLE I don't have any authority here, I was just curious for the motivation of the PR as there are better alternatives already present in BC.

For your Url combine function, you can have a look at the latest changes: https://github.com/microsoft/ALAppExtensions/tree/master/Modules/System/URI/src - now all of the uri properties are exposed: scheme, host, port, path, query, fragment.

Maybe it would make sense to move the HttpUtility wrappers from Base app (Type Helper codeunit) to System app (Uri module) as well or overload the constructor, similar to what you suggested in #7551. Just a thought.

I just want to make sure the API doesn't get out of hand, since these functions are already confusing most .NET programmers and I haven't heard very good things about EscapeUriString, but the decision itself is up to MS of course.

mazhelez:
@MattiLE It's an open discussion :)

The idea is to contribute something that you and other AL developers might potentially us

mazhelez:
@MattiLE Do you plan on working on this PR? If not, can you please close it?

MattiLE:
Hi @mazhelez, I can't say when I will find time to work on this further. I'll close the PR for now and might do a new one later. :)

PR #11346, opened 2021-02-26
(fix): 0 value name changed from "" to " "

kine:
If the value is "" it cannot be used in the AL code. Standard for BC is that empty value is " " (one space) not "".

If used in al code, you get error:
```
'"Credit Memo Type CZL"' does not contain a definition for ' '
```


JesperSchulz (approved):

JesperSchulz:

JesperSchulz:

PR #11557, opened 2021-03-09
"Azure AD Device Plan

bgestel:
Added MembershipEntitlement.type::"Azure AD Device Plan" to the filter of know plans to include , since the device plan is also a known plan to business central.


GiorgosPa:
@bgestel this is not relevant after 18.0 as Membership entitlements usages has been replaced. Given that 17.5 has already been shipped I do not see this Pull Request offering any value. The device Plan should be recognized as a valid Business Central Plan on 18.0.

PR #12061, opened 2021-04-16
Added Tag and Parameters in BCPT Import/Export

Bertverbeek4PS:
In our devops pipelines we import a standard suite. And with this suite we are using parameters.
So we need the field parameters and tag in our import to get more detail for analyses and parameters to set the right parameters and not only the standard parameters.

@BardurKnudsen you now we work with parameters :)


JesperSchulz (approved):
Code reviewed by BCPT owners.

JesperSchulz:

JesperSchulz:

PR #12073, opened 2021-04-19
Improve CI for System Application

IhorHandziuk:
**Before:**
The build workflow compiles "System Application"

**Now:**
The build workflow compiles and publishes:
- System Application
- System Application Test Libraries
- Test framework (Test Runner, Assert, Variable Storage, Any)
- System Application Tests

The build workflow runs System Application Tests and presents the results in a comment on the PR:
image
and in detailed annotations:
image


Modules/System Tests/Assisted Setup/src/AssistedSetupVisibleTransName.PageExt.al
```diff
@@ -3,7 +3,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
 
-pageextension 132588 AssistedSetupVisibleTransName extends "Assisted Setup"
+pageextension 132586 AssistedSetupVisibleTransName extends "Assisted Setup"
```
darjoo, line 6: Did this require a id update?

darjoo (approved):
Looks good

Modules/System Tests/Assisted Setup/src/AssistedSetupVisibleTransName.PageExt.al
IhorHandziuk: With the old ID publishing of System Application Test Libraries fails with the error:
`Your program license does not allow you to publish AssistedSetupVisibleTransName`
You can see it in the logs [here](https://github.com/IhorHandziuk/ALAppExtensions/pull/12).

Since pageextensions piggyback on pages in the license, I changed this ID to be the same as in one of the pages.

JesperSchulz (approved):

PR #12093, opened 2021-04-21
Missing Captions added

DanielGoehler:
I have added the missing Captions in **System** so that they can be translated into other languages.

### DIFF
--- a/Modules/System/BLOB Storage/src/PersistentBlob.Table.al
+++ b/Modules/System/BLOB Storage/src/PersistentBlob.Table.al
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------------------------------------------
+// ------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
@@ -6,6 +6,7 @@
 table 4151 "Persistent Blob"
 {
     Access = Internal;
+    Caption = 'Persistent Blob';
 
     fields
     {
--- a/Modules/System/BLOB Storage/src/TempBlob.Table.al
+++ b/Modules/System/BLOB Storage/src/TempBlob.Table.al
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------------------------------------------
+// ------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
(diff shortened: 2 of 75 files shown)


JesperSchulz:
Thanks Daniel! Let me try to follow up with the compiler team. Those captions really shouldn't be necessary! Well, first they weren't (C/AL), then they were (AL), but then I was promised that now they aren't again (18.0) :-D I'm confused myself, so please allow me to look into this. I'd really prefer a platform fix, rather than having to "pollute" / "enrich" our code with captions everywhere.

DanielGoehler:
Last time you promised to add Captions and a CodeCop rule (https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=944456655069184). If captions are no longer necessary, that's fine with me. Did I somehow miss this announcement?

JesperSchulz:
You didn't miss an announcement. I indeed went into discussions about implementing a CodeCop rule. But then we concluded, that it would be more elegant, if a table/page/report's caption based on the object's name would get injected into the translation file, if a caption was missing. But that work stranded due to some outstanding conclusions and hence didn't make it all the way into 18.0! We're going to look at getting that fixed, as I think that makes most sense. Translations for objects would then behave the same, as they do for e.g. table fields. I'd prefer that over yet another CodeCop rule.
However, your PR is not just object captions! It's a bunch of action captions as well I see. Those we definitely need to get processed, as those indeed are missing captions, which won't get addressed by any platform solution. Please allow me to keep this PR open, while I figure things out. 18.1 has sailed anyway. I'm going to target a fix for this particular PR for 18.2.

Thanks a lot for bringing this up again. I hope we nail it down this time around.

DanielGoehler:
I think that this is not necessary.

In the meantime, I looked at the [documentation](https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-work-with-translation-files#generating-the-xliff-file). I was not aware that there is a new function **GenerateCaptions** that generates missing captions for objects, actions and fields. So my Actions Caption are not necessary either. You could think about adding `"Features": [ "TranslationFile", "GenerateCaptions" ]` into the **app.json** to make this stand out better.

I think you can could reject this PR.

JesperSchulz:
The GenerateCaptions function is exactly what we were going to look into. The compiler team kindly implemented this for us as a result of the previous discussion (CodeCop etc.), but I somehow didn't get the note that this had been completed and I hence never uptook it in our apps. That's going to change now! I didn't know that that function will add captions for actions as well though! I was told, that it only works on objects and fields - I'd have to double check. Let's keep this PR open, until we've got a solution for all issues in here. That way, I've got a hook to attach my work to and to make sure, that it gets properly done this time :-)

mazhelez:
It now confirmed: using `GenerateCaption` from as described in the [documentation](https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-work-with-translation-files) makes duplicating the object name as caption unnecessary.

@DanielGoehler thanks again for paying attention! In this case, I think, we reached the optimal solution :)

PR #12281, opened 2021-05-07
:sparkles: Add procedure ToRecord to TempBlob

pri-kise:
This PR adds a new procedure ToRecord for Blob Storage Module

### DIFF
--- a/Modules/System/BLOB Storage/README.md
+++ b/Modules/System/BLOB Storage/README.md
@@ -213,6 +213,29 @@ Any Record variable.
 
 The field number of the BLOB field to be read.
 
+### ToRecord (Method) <a name="ToRecord"></a> 
+
+ Copies the value of the TempBlob to the specified field on the Record.
+ 
+#### Syntax
+```
+procedure ToRecord(var RecordVariant: Variant; FieldNo: Integer)
+```
+#### Parameters
+*RecordVariant ([Variant](https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/variant/variant-data-type))* 
+
+Any Record variable.
+
+*FieldNo ([Integer](https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/integer/integer-data-type))* 
+
+The field number of the Blob field to be written.
+
+#### Return Value
+*[Variant](https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/variant/variant-data-type)* 
+
+Returns the Record with the updated Blob field as Variant.
+
+
 ### FromRecordRef (Method) <a name="FromRecordRef"></a> 
 
  Copies the value of the BLOB field on the RecordRef in the specified field to the TempBlob.
--- a/Modules/System/BLOB Storage/README.md
+++ b/Modules/System/BLOB Storage/README.md
@@ -219,7 +219,7 @@ The field number of the BLOB field to be read.
  
 #### Syntax
 ```
-procedure ToRecord(var RecordVariant: Variant; FieldNo: Integer)
+procedure ToRecord(RecordVariant: Variant; FieldNo: Integer): Variant
 ```
 #### Parameters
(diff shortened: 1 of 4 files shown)


JesperSchulz:
A similar PR like this was actually submitted a couple of months back, but back then, it provoked runtime errors.
With the introduction of complex return types, the code you submitted actually works, but when we tried to implement it a couple of months back, we found that this implementation is approx. 50% slower than the RecRef implementation. We need to get that fixed, before we can continue with this PR. This might take a little while, but we're on it :-)

pri-kise:
Okay thanks.
Then I would keep this PR open until someone responses that the performance has changed?

One small side note: @JesperSchulz could you check why the Build has crashed? Looks like the license isn't working for some test files...?

JesperSchulz:
Yes, please keep the PR open! We'll what we can do about this performance issue.
And thanks for pointing out the build error. We added a table in a test, which isn't in the demo license and hence publishing the tests now fails. We're working on a fix for 18.2 :-)

Modules/System/BLOB Storage/README.md
```diff
@@ -213,6 +213,29 @@ Any Record variable.
 
 The field number of the BLOB field to be read.
 
+### ToRecord (Method) <a name="ToRecord"></a> 
+
+ Copies the value of the TempBlob to the specified field on the Record.
+ 
+#### Syntax
+```
+procedure ToRecord(var RecordVariant: Variant; FieldNo: Integer)
```
jwikman, line 222:
```suggestion
procedure ToRecord(RecordVariant: Variant; FieldNo: Integer): Variant
```
The parameter is not "var" and the return type was missing :)

pri-kise:
@JesperSchulz do you know if there  has anything changed on the performance?
I added a test to enable the processing of the PR.

IhorHandziuk:
@pri-kise , the assignment to/from Variant causes a record copy. The variant doesn’t use references and hence the perf-cost.
It is not something we have any plans for changing in the near future, so unfortunately, we cannot move forward with this PR.

You can submit the idea of having `var Variant`s on aka.ms/bcideas, and if a lot of people will find it useful, we might prioritize this work.

JesperSchulz:
As the provided code will add a substantial performance cost, which our platform cannot address at the moment, we have two options:
1) We accept the code and the performance penalty, making it clear in the code documentation, that this code has inferior performance and shouldn't be used in bulk operations.
2) We reject this code, forcing developers to stik to the performant code construct.

Before making the decision, I would like to hear your opinion, @pri-kise.

pri-kise:
Thanks for your replies.
If i can wish me one solution, then I would suggest to add the current implementation with a warning that this has a high performance cost and shouldn't be used in bulk operations. At the moment I'm mostly thinking about use cases where it would become easier to enable the user to upload some files and save them to a blob field of a record. And those aren't bulk operations.

Additionally I created an idea that might reach enough upvotes so that the AL Language will be enriched with the ability to pass variants by reference.
https://experience.dynamics.com/ideas/idea/?ideaid=1117bb5c-1cd3-eb11-ba5e-0003ff45a074
Then in the future this implementation could be osoleted and getting replaced by  a more performant implementation.
https://github.com/microsoft/ALAppExtensions/pull/9874

JesperSchulz:
This case and your wish were discussed in an architectural forum and it was decided, that we won't move ahead with this PR. At this point in time, we cannot submit any code, which will jeopardize the investments we are doing, in improving the performance of the product. Adding a code construct, which will result in poor performance, is completely orthogonal to our ambitions. Even if we would add a warning, there is a good likelihood, that other developers would ignore that warning and use this code construct in bulk operations after all - we're seeing plenty of such examples in other BLOB related issues, where data is copied around because the code is prettier, rather than moving pointers, where the code is a little more complex. And the problem with BLOB fields is, that copying the data around doesn't cause a small performance penalty, it almost always results in large costs.
I'm hence very sorry to inform you, that we cannot move forward with this PR at this point in time. I am happy that you logged this problem as an idea, and when our platform supports this, we will gladly implement these changes. Until then, we will have to stick with uglier, but much better performing code constructs we've got today. I hope that makes sense, else please let me know and I'll try to explain differently. But thanks for taking the time to create this suggestion!

PR #12358, opened 2021-05-14
Update RecurrenceScheduleCard.Page.al

RadekMiernik:
Typo fixed.

### DIFF
--- a/Modules/System/Recurrence Schedule/src/RecurrenceScheduleCard.Page.al
+++ b/Modules/System/Recurrence Schedule/src/RecurrenceScheduleCard.Page.al
@@ -131,7 +131,7 @@ page 4690 "Recurrence Schedule Card"
             {
                 Caption = 'Monthly';
                 Visible = "Pattern" = RecurrencePatterns::Monthly;
-                field(MontlyFrequency; "Recurs Every")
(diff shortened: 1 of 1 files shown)


darjoo (approved):

darjoo:
Hi @radekmiernik, nice catch on that. 
However we're unable to make that is a breaking change unless we obsolete it first and create a replacement with preprocessor tags.

`#if CLEAN19
...
#elseif
...`

RadekMiernik:
I am totally aware of it. I just want to point out this typo, and I know that a pull request is the best way to 'not to forget about something.

pri-kise:
@darjoo is this really a breaking change, since this page is set to `extensible=false`?
https://github.com/microsoft/ALAppExtensions/blob/5c490ab80d8f5c4ac26641604eba79dbe1b1eb47/Modules/System/Recurrence%20Schedule/src/RecurrenceScheduleCard.Page.al#L11

JesperSchulz:
This one indeed shouldn't be breaking. Let's hope our build system agrees ;-) This should be able to go in "as is".

JesperSchulz (approved):

JesperSchulz:

JesperSchulz:

PR #12513, opened 2021-05-24
Fixed "with" issues, "overflow" warning and "Default Delay Times"

waldo1001:
Fixed "with" issues, "overflow" warning and "Default Delay Times"

### DIFF
--- a/Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTLines.Page.al
+++ b/Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTLines.Page.al
@@ -265,6 +265,6 @@ page 149004 "BCPT Lines"
     internal procedure Refresh()
     begin
         CurrPage.Update(false);
-        if Find() then;
+        if Rec.Find() then;
     end;
 }
--- a/Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTLines.Page.al
+++ b/Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTLines.Page.al
@@ -236,15 +236,18 @@ page 149004 "BCPT Lines"
         BCPTHeader: Record "BCPT Header";
         BCPTLineCU: Codeunit "BCPT Line";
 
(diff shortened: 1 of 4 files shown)


waldo1001:
I see some kind of test failed - but it does work on my end against latest sandbox:
image

darjoo:
> I see some kind of test failed - but it does work on my end against latest sandbox:
> image

Yeah, don't worry about the failure. We know about it internally and it is due to the license being used.

darjoo (approved):
Approved by owning team. Will merge once I've processed it internally.

Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTLines.Page.al
```diff
@@ -242,10 +242,9 @@ page 149004 "BCPT Lines"
             exit(true);
         if Rec."BCPT Code" <> BCPTHeader.Code then
             if BCPTHeader.Get(Rec."BCPT Code") then;
-        if Rec."Min. User Delay (ms)" = 0 then
```
BardurKnudsen, line 245: The idea here was that maybe the user had entered some value before the line gets inserted (delayed insert).
An unconditional assignment would probably fit better in the OnNewRecord trigger, I think.

BardurKnudsen (approved):

Modules/DevTools/BusinessCentralPerformanceToolkit/src/BCPTLines.Page.al
darjoo: I've moved the assignments to OnNewRecord trigger. There should be no issue with the default being 0 because of minValue property being set on the table. The validation will always run when it is changed on the page.

waldo1001: Makes sense!

darjoo:
Added changes made in internal PR. Merging this now.

JesperSchulz:

JesperSchulz:

PR #12559, opened 2021-05-26
expose variables in the EmailEditor page for extensibility

quadfin:
To enable custom validation on the recipient address fields (using a page extension with OnValidate and OnAssistEdit triggers for the To,Cc,Bcc fields), I moved the variables that retain the information into the protected scope.

### DIFF
--- a/Other/Tests/BCPT-SampleTests/src/BCPTMakeWebCall.Codeunit.al
+++ b/Other/Tests/BCPT-SampleTests/src/BCPTMakeWebCall.Codeunit.al
@@ -9,7 +9,7 @@ codeunit 149108 "BCPT Make Web Call"
         ResponseMessage: HttpResponseMessage;
         Client: HttpClient;
     begin
-        NewUrl := 'https://wwww.microsoft.com';
+        NewUrl := 'https://www.microsoft.com';
 
         RequestMessage.SetRequestUri(NewUrl);
         RequestMessage.Method('GET');


Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
```diff
@@ -385,4 +383,8 @@ page 13 "Email Editor"
         CloseThePageQst: Label 'The email has not been sent.';
         OptionsOnClosePageNewEmailLbl: Label 'Keep as draft in Email Outbox,Discard email';
         PageCaptionTxt: Label 'Compose an Email';
+
+    protected var
+        EmailMessage: Codeunit "Email Message Impl.";
```
mazhelez, line 388: Why do you want to expose this?

This is internal implementation, but it's exposed as a var, it means it'll be a breaking change to change this implementation.

quadfin: We wish to implement customized validation and lookup on the recipient addresses (To,Cc,Bcc).
Think of an address book feature - we wish to provide the user with assistance finding and inputting addressees.
The validation part can be supported by exposing the ToRecipient, CcRecipient, BccRecipient variables in the protected scope (as I have done in the PR).
The assisted lookup must be performed in a new trigger (OnAssistEdit, or OnLookup).
However, because the page uses the EmailMessage variable to persist the email addresses, and because it refreshes the values of the ToRecipient... variables in the OnAfterGetRecord trigger, the only way to persist the data from a page extension is to update the recipients in the EmailMessage variable - thus requiring that it be moved into the protected scope so the page extension has access to it.
There are other ways to implement this request however they would require a more impactful redesign of the page to move the persistence of the email addresses out of the EmailMessage variable and into local variables.

JesperSchulz: Alternatively, we add an extensible address book to the module itself. It's actually in our backlog. We cannot expose our internal implementations - it's against the purpose of what we're trying to achieve ;-) Having a simple address book shouldn't be partner added value; it should be part of the base product. Adding more complex capabilities and tailored experiences, would be something we would want partners to spend their time on. I guess what I'm saying is, that we will try to add something simple for version 19 for you to extend. At least, I'll see if I somehow can find budget to do so. I will have to get back to you on this one!

leonardo-matta:
@JesperSchulz  is there anything on the roadmap for setting CCs and BCCs automatically per Email Scenario given that exposing the Codeunit is not an option? The company I work for is implementing BC as our ERP, but we're a Google shop so we'll use SMTP for sending emails. In our case, we would like to copy the "from" field as BCC so we have a copy of the email in our mailbox.

JesperSchulz:
@leonardo-matta, the related PR #12867 should be able to open up for your scenario :-)

GiorgosPa:
@leonardo-matta you can do that by implementing your own connector that can edit the message and pass it to SMTP connector to send it. The related PR #12877 will not help your case.

leonardo-matta:
Thanks, guys.

PR #12560, opened 2021-05-27
Allow sending emails without the "To" recipients

IhorHandziuk:
**Problem:**
It is not possible to send emails from BC if there is no email address in the "To" field.

**Solution:**
Change the validation logic so that it's enough to have at least one recipient in either of the fields: "To", "CC", "BCC"

### DIFF
--- a/Modules/System Tests/Email/src/EmailMessageUnitTest.Codeunit.al
+++ b/Modules/System Tests/Email/src/EmailMessageUnitTest.Codeunit.al
@@ -290,7 +290,7 @@ codeunit 134689 "Email Message Unit Test"
 
         // [WHEN] An email is sent
         // [THEN] A validation error occurs
-        Email.Send(EmailMessage, TempEmailAccount."Account Id", TempEmailAccount.Connector);
+        asserterror Email.Send(EmailMessage, TempEmailAccount."Account Id", TempEmailAccount.Connector);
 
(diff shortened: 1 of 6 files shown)


Modules/System Tests/Email/src/EmailMessageUnitTest.Codeunit.al
```diff
@@ -247,6 +248,54 @@ codeunit 134689 "Email Message Unit Test"
...
+        // Initialize
+        ConnectorMock.Initialize();
+        ConnectorMock.AddAccount(TempEmailAccount);
+
+        // [GIVEN] The message doesn't have any recipients
+        EmailMessage.Create(Recipients, 'Test subject', 'Test body', true, RecipientsCC, RecipientsBCC);
+
+        // [WHEN] An email is sent
+        // [THEN] A validation error occurs
+        Email.Send(EmailMessage, TempEmailAccount."Account Id", TempEmailAccount.Connector);
```
NKarolak, line 293: Isn't an asserterror missing here?

IhorHandziuk: Yes, thanks for quickly catching the problem! I would expect the build workflow to show the failed test as well, will try to make some changes to make that happen

JesperSchulz (approved):
Great work, Ihor ;-)

PR #12867, opened 2021-06-15
expose recipient variables for page extensibility

quadfin:
Modify EmailDialog to move the recipient address variables (ToRecipient, CcRecipient, BccRecipient) into the protected scope and enable setting those variables into the (not publicly accessible) EmailMessage variable via the new (public) SetRecipients procedure.
With that, one can then implement a page extension that manipulates the (To,Cc,Bcc) input using the OnBeforeValidate and OnAssistEdit triggers for each field.

### DIFF
--- a/Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
+++ b/Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
@@ -62,7 +62,7 @@ page 13 "Email Editor"
 
                             trigger OnValidate()
                             begin
-                                EmailMessage.SetRecipients(Enum::"Email Recipient Type"::"To", ToRecipient);
+                                SetRecipients(Enum::"Email Recipient Type"::"To", ToRecipient);
                             end;
                         }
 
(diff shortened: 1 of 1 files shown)


quadfin:
This is a follow on PR to (now closed) PR #12559.
I have changed the implementation to access the (non public) EmailMessage variable via a setter procedure for the recipient data.
In retrospect, perhaps also adding a getter procedure would be useful.

GiorgosPa (changes requested):
I am afraid this approach also poses issues as it allows for modifications of the email just before it is sent.

What we can do though to support your scenario is to provide OnLookup event that you can use for your needs.

You can add the following trigger on each page field
```
    trigger OnLookup(var Text: Text): Boolean
    var
        LookupOk: Boolean;
    begin
        LookupOk := true;
        OnLookupToRecipient(Text, LookupOk);
        exit(LookupOk)
    end;
```

and this event on the page

```
    local procedure OnLookupToRecipient(var ToRecipient: Text; LookupOk: Boolean)
    begin
    end;
```

quadfin:
I will update per the OnLookup design pattern.
I am closing this PR, will open a new one.

PR #12873, opened 2021-06-16
Added Tenant in "Email Application AAD Registration"

Nyryus:
Modified "Email Application AAD Registration" adding the Tenant field, I've kept the old value "Common" as default, but we also needed to use a custom value.

There is still a problem I wasn't able to fix with the cache.
When closing the "Email Application AAD Registration" page, the cache should be cleared and the new setup should be cached.
Currently, if you set the TenantID with a valid TenantID and send some emails, then you change the value with a not working one (Common in my case), emails are still sent due to the cache and the setup is skipped.

### DIFF
--- a/Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
+++ b/Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
@@ -40,7 +40,7 @@ codeunit 4507 "Email - OAuth Client" implements "Email - OAuth Client"
                 if OAuth2.AcquireOnBehalfOfToken('', GraphResourceURLTxt, AccessToken) then;
         end else begin
             Initialize();
-            if (not OAuth2.AcquireAuthorizationCodeTokenFromCache(ClientId, ClientSecret, RedirectURL, StrSubstNo(OAuthAuthorityUrlTxt, TenantId), GraphResourceURLTxt, AccessToken)) or (AccessToken = '') or Force then
+            if (not OAuth2.AcquireAuthorizationCodeTokenFromCache(ClientId, ClientSecret, StrSubstNo(RedirectURL, TenantId), StrSubstNo(OAuthAuthorityUrlTxt, TenantId), GraphResourceURLTxt, AccessToken)) or (AccessToken = '') or Force then
                 OAuth2.AcquireTokenByAuthorizationCode(ClientId, ClientSecret, StrSubstNo(OAuthAuthorityUrlTxt, TenantId), StrSubstNo(RedirectURL, TenantId), GraphResourceURLTxt, Enum::"Prompt Interaction"::None, AccessToken, OAuthErr);
         end;
 
@@ -184,7 +184,6 @@ codeunit 4507 "Email - OAuth Client" implements "Email - OAuth Client"
 
     var
         OAuth2: Codeunit OAuth2;
-        
         [NonDebuggable]
         ClientId: Text;
         [NonDebuggable]
--- a/Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
+++ b/Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
@@ -37,11 +37,11 @@ codeunit 4507 "Email - OAuth Client" implements "Email - OAuth Client"
     begin
         if EnvironmentInformation.IsSaaSInfrastructure() then begin
             if (not AcquireOnBehalfOfTokenFromStoredTokenCache(AccessToken)) or (AccessToken = '') then
-                if OAuth2.AcquireOnBehalfOfToken('', GraphResourceURLTxt, AccessToken) then;
+                if OAuth2.AcquireOnBehalfOfToken('', Scopes, AccessToken) then;
         end else begin
             Initialize();
-            if (not OAuth2.AcquireAuthorizationCodeTokenFromCache(ClientId, ClientSecret, StrSubstNo(RedirectURL, TenantId), StrSubstNo(OAuthAuthorityUrlTxt, TenantId), GraphResourceURLTxt, AccessToken)) or (AccessToken = '') or Force then
-                OAuth2.AcquireTokenByAuthorizationCode(ClientId, ClientSecret, StrSubstNo(OAuthAuthorityUrlTxt, TenantId), StrSubstNo(RedirectURL, TenantId), GraphResourceURLTxt, Enum::"Prompt Interaction"::None, AccessToken, OAuthErr);
+            if (not OAuth2.AcquireAuthorizationCodeTokenFromCache(ClientId, ClientSecret, RedirectURL, StrSubstNo(OAuthAuthorityUrlTxt, TenantId), Scopes, AccessToken)) or (AccessToken = '') or Force then
+                OAuth2.AcquireTokenByAuthorizationCode(ClientId, ClientSecret, StrSubstNo(OAuthAuthorityUrlTxt, TenantId), RedirectURL, Scopes, Enum::"Prompt Interaction"::None, AccessToken, OAuthErr);
         end;
 
         if AccessToken = '' then
@@ -64,6 +64,11 @@ codeunit 4507 "Email - OAuth Client" implements "Email - OAuth Client"
             TenantId := EmailOutlookAPIHelper.GetTenantID();
         end;
 
+        Scopes.Add('https://graph.microsoft.com/User.Read');
+        Scopes.Add('https://graph.microsoft.com/mail.ReadWrite');
+        Scopes.Add('https://graph.microsoft.com/mail.send');
+        Scopes.Add('https://graph.microsoft.com/offline_access');
+
         IsInitialized := true;
     end;
 
@@ -81,7 +86,7 @@ codeunit 4507 "Email - OAuth Client" implements "Email - OAuth Client"
             exit(false);
         end;
 
-        if (not OAuth2.AcquireOnBehalfOfTokenByTokenCache(User."Authentication Email", '', GraphResourceURLTxt, TokenCache, AccessToken, NewTokenCache)) or (AccessToken = '') then begin
+        if (not OAuth2.AcquireOnBehalfOfTokenByTokenCache(User."Authentication Email", '', Scopes, TokenCache, AccessToken, NewTokenCache)) or (AccessToken = '') then begin
             Session.LogMessage('000040B', StrSubstNo(CouldNotAcquireAccessTokenFromCacheErr, UserSecurityId()), Verbosity::Error, DataClassification::EndUserPseudonymousIdentifiers, TelemetryScope::ExtensionPublisher, 'Category', EmailCategoryLbl);
             if not IsolatedStorage.Delete(TokenCacheTok + UserSecurityId()) then
                 Session.LogMessage('000040C', StrSubstNo(CouldNotDeleteTokenCacheTxt, UserSecurityId()), Verbosity::Warning, DataClassification::EndUserPseudonymousIdentifiers, TelemetryScope::ExtensionPublisher, 'Category', EmailCategoryLbl);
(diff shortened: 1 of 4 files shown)


IhorHandziuk:
Hi @Nyryus and thanks for making a contribution!

Can you describe the scenario that you are trying to address with this change? Is it for BC running in a multitenant deployment?

Nyryus:
Hi @IhorHandziuk, yes. We have multiple tenants and with this change, we are able to address the right app registration

IhorHandziuk:
@Nyryus , sorry for a long delay in replies, but this is indeed a great idea. We will definitely need to fix the problem with cache though: wouldn't it work to call `SignInUsingAuthorizationCode` function from `Email - OAuth Client` (that's how we currently make the cache to refresh from storing the token for the previous user to to storing the cache for the new user)?

Apps/W1/Email - Outlook REST API/app/src/EmailOutlookAPISetup.Table.al
```diff
@@ -29,6 +29,12 @@ table 4509 "Email - Outlook API Setup"
             Caption = 'Redirect URL';
             DataClassification = CustomerContent;
         }
+        field(5; TenantID; Text[36])
```
IhorHandziuk, line 32: Maybe type should be `Guid` (to be clear what data will be sored)?

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
```diff
@@ -34,8 +40,8 @@ codeunit 4507 "Email - OAuth Client" implements "Email - OAuth Client"
                 if OAuth2.AcquireOnBehalfOfToken('', GraphResourceURLTxt, AccessToken) then;
         end else begin
             Initialize();
-            if (not OAuth2.AcquireAuthorizationCodeTokenFromCache(ClientId, ClientSecret, RedirectURL, OAuthAuthorityUrlTxt, GraphResourceURLTxt, AccessToken)) or (AccessToken = '') then
-                OAuth2.AcquireTokenByAuthorizationCode(ClientId, ClientSecret, OAuthAuthorityUrlTxt, RedirectURL, GraphResourceURLTxt, Enum::"Prompt Interaction"::None, AccessToken, OAuthErr);
+            if (not OAuth2.AcquireAuthorizationCodeTokenFromCache(ClientId, ClientSecret, RedirectURL, StrSubstNo(OAuthAuthorityUrlTxt, TenantId), GraphResourceURLTxt, AccessToken)) or (AccessToken = '') or Force then
```
IhorHandziuk, line 43: Shouldn't this be `StrSubstNo(RedirectURL, TenantId)` as well? Same applies to lines 172 and 182.

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
```diff
@@ -133,7 +140,7 @@ codeunit 4507 "Email - OAuth Client" implements "Email - OAuth Client"
         AccessToken: Text;
         TokenCache: Text;
     begin
-        if not OAuth2.AcquireOnBehalfAccessTokenAndTokenCache(OAuthAuthorityUrlTxt, '', GraphResourceURLTxt, AccessToken, TokenCache) then begin
+        if not OAuth2.AcquireOnBehalfAccessTokenAndTokenCache(StrSubstNo(OAuthAuthorityUrlTxt, TenantId), '', GraphResourceURLTxt, AccessToken, TokenCache) then begin
```
IhorHandziuk, line 143: Is there a reason you want to modify the SaaS logic path as well (is it to make sure guest / external users are getting the correct endpoint)?

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
```diff
@@ -172,19 +179,20 @@ codeunit 4507 "Email - OAuth Client" implements "Email - OAuth Client"
...
     begin
         Initialize();
-        exit(OAuth2.AcquireTokenByAuthorizationCode(ClientID, ClientSecret, OAuthAuthorityUrlTxt, RedirectURL, GraphResourceURLTxt, Enum::"Prompt Interaction"::"Select Account", AccessToken, OAuthErr) and (AccessToken <> ''));
+        exit(OAuth2.AcquireTokenByAuthorizationCode(ClientID, ClientSecret, StrSubstNo(OAuthAuthorityUrlTxt, TenantId), RedirectURL, GraphResourceURLTxt, Enum::"Prompt Interaction"::"Select Account", AccessToken, OAuthErr) and (AccessToken <> ''));
     end;
 
     var
         OAuth2: Codeunit OAuth2;
-
+
```
IhorHandziuk, line 187: Empty line

Apps/W1/Email - Outlook REST API/app/src/EmailOutlookAPISetup.Page.al
```diff
@@ -268,6 +281,7 @@ page 4509 "Email - Outlook API Setup"
         UnsuccessfulTestMsg: Label 'We could not get access token with the current setup.\\Please verify the values on the page and try again.';
         SuccessfulTestMsg: Label 'Success! Your authentication was verified.';
         HiddenValueTxt: Label '******', Locked = true;
+        DefaultTenantId: Label 'Common', Locked = true;
```
IhorHandziuk, line 284: You were right to default to the value we are currently using, but now I'm thinking: maybe we should default to the actual tenant ID if we know the tenant (i. e. to `TenantInformation.GetTenantID()`). What do you think?

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
```diff
@@ -12,19 +12,25 @@ codeunit 4507 "Email - OAuth Client" implements "Email - OAuth Client"
...
+    procedure TryGetAccessToken(var AccessToken: Text; Force: Boolean): Boolean
+    begin
+        exit(TryGetAccessTokenInternal(AccessToken, Force));
     end;
 
     // Interfaces do not support properties for the procedures, so using an internal function
     [TryFunction]
     [NonDebuggable]
-    local procedure TryGetAccessTokenInternal(var AccessToken: Text)
+    local procedure TryGetAccessTokenInternal(var AccessToken: Text; Force: Boolean)
```
IhorHandziuk, line 33: What is the purpose of the `Force` parameter? From what I can see it just allows to skip attempting to get the token from cache.
Is it a workaround to the problem with cache you mentioned in the PR's summary? If yes, I think we should find a better solution (e. g. using `SignInUsingAuthorizationCode`, see comment above).

IhorHandziuk (changes requested):

IhorHandziuk:
Hi @Nyryus , it seems like this PR hasn't had any activity recently. Did you find another way to address the problem? 😊

Apps/W1/Email - Outlook REST API/app/src/EmailOutlookAPISetup.Table.al
Nyryus: I've set it to Text to allow the standard value "Common". Another solution could be to get the tenant ID, but OnPremise this would not work properly

Apps/W1/Email - Outlook REST API/app/src/EmailOutlookAPISetup.Page.al
Nyryus: I think that is not a problem, when opening the page "Email - Outlook API Setup" if the value is equal to "Common" try to get the TenantID with that command. By the way, for OnPrem customers, we should keep "common" as the default value and allow them to change it.

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
Nyryus: This change is needed because I changed the default value of OAuthAuthorityUrlTxt, inserting the "%1" instead of the "common"

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
Nyryus: Fixed with my latest commit

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
Nyryus: Yes, you are right, fixed

Nyryus:
Hi @IhorHandziuk sorry, you are right, no, I didn't found any alternative solution, but was busy with work :)
I'll try to fix it today. From tomorrow I will be on holiday for 2 weeks. When I get back I will try to answer your further questions as soon as possible.

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
Nyryus: My bad, RedirectURL don't need the StrSubNo with the tenantID

Apps/W1/Email - Outlook REST API/app/src/EmailOutlookAPISetup.Page.al
Nyryus: I've update the functionality, making it use "Common" for OnPrem environments and the tenantID as suggeted for SaaS with TenantInformation.GetTenantID()

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
Nyryus: Yes, this is a workaround for the problem that I've mentioned.
I've tried to fix it as you suggested using the SignInUsingAuthorizationCode, but it opens a Window asking the Account, probably I'm missing something, could you please help me figure it out?
I think the problem is the fact that the cache is never deleted, then if you change the setup, the previous cache still exists allowing you to send an email with that one.

Apps/W1/Email - Outlook REST API/app/src/EmailOutlookAPISetup.Table.al
IhorHandziuk: Maybe "Common" can be represented as a null guid in the table, and then we could have a small helper function that substitutes the tenant ID into the string with the logic to convert null guid into `Common`?

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
IhorHandziuk: 👍

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
IhorHandziuk: Thanks!

JesperSchulz:
Hi @Nyryus,
I hope you had a great vacation and are fully reenergized to get this PR completed :-) Could you please address the outstanding issues to help us complete this PR? Thanks a lot!

Apps/W1/Email - Outlook REST API/app/src/EmailOutlookAPISetup.Table.al
Nyryus: I've done this, could you please give me feedback. Is it in the way you thought?

Apps/W1/Email - Outlook REST API/app/src/EmailOAuthClient.Codeunit.al
Nyryus: I think that it's ok, but from the UX perspective, I don't find it clear to the end-user.
Probably nobody will change the Tenant of the secret key. But I think think that the path to making it work is not clear.

Anyway, looking to the PR purpose, I've added the tenant field and made it work, I'll add Ideas for this purpose

IhorHandziuk:
> Hi @IhorHandziuk, yes. We have multiple tenants and with this change, we are able to address the right app registration

@Nyryus , can you please tell us more about your scenario? We had some internal discussions and would like to hear more details about what exactly turns out to be the problem if you use a multitenant AAD application registration together with a `common` endpoint.

IhorHandziuk:
Hi @Nyryus , we have noticed that this PR hasn't been active in a long period of time.
If there is no activity, we, unfortunately, will have to close this PR.

PR #12878, opened 2021-06-17
update per GiorgosPa OnLookup pattern

quadfin:
Previously - GiorgosPa wrote:

I am afraid this approach also poses issues as it allows for modifications of the email just before it is sent.
What we can do though to support your scenario is to provide OnLookup event that you can use for your needs.
You can add triggers like bellow on each field:

    trigger OnLookup(var Text: Text): Boolean
    var
        LookupOk: Boolean;
    begin
        OnLookupToRecipient(Text, LookupOk);
        exit(LookupOk)
    end;

and the necessary events on the page like:

    local procedure OnLookupToRecipient(var ToRecipient: Text; LookupOk: Boolean)
    begin
    end;
------------------------

I prototyped the pattern - it works if there is a way for the subscriber to know which instance of the Email Editor raised the event.
This is necessary because the EmailEditor page extension loads content that is specific to the message -- e.g., Customer/Contact addresses for a specific sales order, etc.
I updated my PR to include the suggested pattern - please let me know if it is acceptable.

Here is a quick demo of the (rough) prototype:  [https://youtu.be/Ah26At-VLuA]
FYI - my intent is to release this as a free app (it will be a teaser for my paid app).

### DIFF
--- a/Modules/System Tests/Translation/src/TranslationTestTableTwo.Table.al
+++ b/Modules/System Tests/Translation/src/TranslationTestTableTwo.Table.al
@@ -1,40 +1,41 @@
-﻿// ------------------------------------------------------------------------------------------------
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-// ------------------------------------------------------------------------------------------------
+﻿// // ------------------------------------------------------------------------------------------------
+// // Copyright (c) Microsoft Corporation. All rights reserved.
+// // Licensed under the MIT License. See License.txt in the project root for license information.
+// // ------------------------------------------------------------------------------------------------
 
-table 137122 "Translation Test Table Two"
-{
+//TODO: Enable in BC19 and add to License
+// table 137122 "Translation Test Table Two"
+// {
 
-    fields
-    {
-        field(1; PK; Integer)
-        {
-        }
-        field(2; TextField; Text[2048])
-        {
-            Editable = false;
-        }
-        field(3; SecondTextField; Text[2048])
-        {
-            Editable = false;
-        }
-    }
+//     fields
+//     {
+//         field(1; PK; Integer)
+//         {
+//         }
+//         field(2; TextField; Text[2048])
+//         {
+//             Editable = false;
+//         }
+//         field(3; SecondTextField; Text[2048])
+//         {
+//             Editable = false;
+//         }
+//     }
 
-    keys
-    {
-        key(Key1; PK)
-        {
-            Clustered = true;
-        }
-    }
+//     keys
+//     {
+//         key(Key1; PK)
+//         {
+//             Clustered = true;
+//         }
+//     }
 
-    trigger OnDelete()
-    begin
-        Translation.Delete(Rec);
-    end;
+//     trigger OnDelete()
+//     begin
+//         Translation.Delete(Rec);
+//     end;
 
-    var
-        Translation: Codeunit Translation;
-}
+//     var
+//         Translation: Codeunit Translation;
+// }
 
--- a/Modules/System Tests/Translation/src/TranslationTestTableTwo.Table.al
+++ b/Modules/System Tests/Translation/src/TranslationTestTableTwo.Table.al
@@ -1,0 +1,40 @@
+﻿// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+table 137122 "Translation Test Table Two"
+{
+
+    fields
+    {
+        field(1; PK; Integer)
+        {
+        }
+        field(2; TextField; Text[2048])
+        {
+            Editable = false;
+        }
+        field(3; SecondTextField; Text[2048])
+        {
+            Editable = false;
+        }
+    }
+
+    keys
+    {
+        key(Key1; PK)
+        {
+            Clustered = true;
+        }
+    }
+
+    trigger OnDelete()
+    begin
+        Translation.Delete(Rec);
+    end;
+
+    var
+        Translation: Codeunit Translation;
+}
+
--- a/Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
+++ b/Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
@@ -59,10 +59,20 @@ page 13 "Email Editor"
                             ToolTip = 'Specifies the email addresses to send the email to.';
                             Editable = not EmailScheduled;
                             Importance = Promoted;
+                            Lookup = true;
 
                             trigger OnValidate()
                             begin
                                 EmailMessage.SetRecipients(Enum::"Email Recipient Type"::"To", ToRecipient);
+                            end;
+
+                            trigger OnLookup(var Text: Text): Boolean
+                            var
+                                LookupOk: Boolean;
+                            begin
+                                LookupOk := true;
+                                OnLookupToRecipient(Text, LookupOk);
+                                exit(LookupOk)
                             end;
                         }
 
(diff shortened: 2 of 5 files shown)


Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
```diff
@@ -59,11 +59,21 @@ page 13 "Email Editor"
...
                             trigger OnValidate()
                             begin
                                 EmailMessage.SetRecipients(Enum::"Email Recipient Type"::"To", ToRecipient);
                             end;
+
+                            trigger OnLookup(var Text: Text): Boolean
+                            var
+                                LookupOk: Boolean;
+                            begin
+                                LookupOk := true;
```
GiorgosPa, line 73: Is this needed?

Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
```diff
@@ -369,13 +400,18 @@ page 13 "Email Editor"
         IsNewOutbox := true;
     end;
 
+    procedure GetEmailEditorId(): Guid
```
GiorgosPa, line 403: This is achieved with Manual event Subscription. Instead of exposing an instance Id variable you should manually bind the event subscription on a specific codeunit instance on the OpenPage of your page extension, now you can set on this instance any other value you need, like contact, Invoice etc.

Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
```diff
@@ -385,4 +421,13 @@ page 13 "Email Editor"
...
         OptionsOnClosePageNewEmailLbl: Label 'Keep as draft in Email Outbox,Discard email';
         PageCaptionTxt: Label 'Compose an Email';
+
+    protected var
+        ToRecipient, CcRecipient, BccRecipient : Text;
+
+
+    [IntegrationEvent(false, false)]
+    local procedure OnLookupRecipient(EmailEditorInstanceId: Guid; RecipientType: Enum "Email Recipient Type"; var Recipient: Text; var LookupOk: Boolean)
+    begin
```
GiorgosPa, line 431: I would prefer three separate events, one for each field.

GiorgosPa (changes requested):

Modules/System Tests/Translation/src/TranslationTestTableTwo.Table.al
```diff
@@ -0,0 +1,41 @@
+﻿// // ------------------------------------------------------------------------------------------------
+// // Copyright (c) Microsoft Corporation. All rights reserved.
+// // Licensed under the MIT License. See License.txt in the project root for license information.
+// // ------------------------------------------------------------------------------------------------
+
+//TODO: Enable in BC19 and add to License
+// table 137122 "Translation Test Table Two"
```
grobyns, line 7: looks like something went wrong in the merge and you pulled in some unrelated things

Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
```diff
@@ -73,11 +83,21 @@ page 13 "Email Editor"
...
                             trigger OnValidate()
                             begin
                                 EmailMessage.SetRecipients(Enum::"Email Recipient Type"::Cc, CcRecipient);
                             end;
+
+                            trigger OnLookup(var Text: Text): Boolean
+                            var
+                                LookupOk: Boolean;
+                            begin
+                                LookupOk := true;
```
grobyns, line 97: you shouldn't initialize LookupOk to true. If no one subscribes and nothing happens, the lookup should return false. (same for the other 2 events)

Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
```diff
@@ -385,4 +414,23 @@ page 13 "Email Editor"
         CloseThePageQst: Label 'The email has not been sent.';
         OptionsOnClosePageNewEmailLbl: Label 'Keep as draft in Email Outbox,Discard email';
         PageCaptionTxt: Label 'Compose an Email';
+
+    protected var
```
grobyns, line 418: there's no need for these variables to be exposed. you can undo this change

Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al
```diff
@@ -59,11 +59,21 @@ page 13 "Email Editor"
...
                             begin
                                 EmailMessage.SetRecipients(Enum::"Email Recipient Type"::"To", ToRecipient);
                             end;
+
+                            trigger OnLookup(var Text: Text): Boolean
+                            var
+                                LookupOk: Boolean;
+                            begin
+                                LookupOk := true;
+                                OnLookupToRecipient(Text, LookupOk);
```
grobyns, line 74: it should be possible to write a test or two, one that does a lookup to enter a value and one that doesn't... in both cases the values should be validated... Maybe @GiorgosPa add some more guidance...

grobyns (changes requested):

quadfin:
Please note:
where @grobyns comments: _there's no need for these variables to be exposed. you can undo this change_
In order for a page extension to perform pre-validation (in the OnBeforeValidate trigger) the variables must be in the protected scope.  that is one of the goals for this PR.

GiorgosPa:
@quadfin please update your branch it seams your merge with master went wrong and there are unrelated changes included

quadfin:
@GiorgosPa, I fetched/merged upstream - branch is now in sync with microsoft:master + this change https://github.com/quadfin/ALAppExtensions/commit/34bb842e45a9868d9e60576829f62f9481b6d040
PR shows 1 file changed:  Modules/System/Email/src/Email/Outbox/EmailEditor.Page.al

JesperSchulz:
I just watched your teaser - it looks really neat! However, as I mentioned in one of your previous PRs (https://github.com/microsoft/ALAppExtensions/pull/12559#discussion_r641424316), this simple version of the address book is going to overlap to a high degree, with what we're working on / have planned. From your message, it sounds as if you plan to make an even more capable, payed version, where we won't be going. But your freemium and our "base" address book will overlap to some extend. I understand the concept of a freemium version, at attract buying customers, but somehow that model also goes against my hopes with this open source project. Having an address book in the email editor, was one of those functionalities I had hoped would get contributed - or at least we would work together on. It should be considered "must" functionality, that has to be existing by default when the product is "installed" and not be an add-on, that every customer must install.
Be that as it may, we will of course make sure that our address book is fully replaceable with more capable ones - such as yours. We're most likely going to add a handled pattern to support this.
But before we go ahead and provide those, I just have to ask directly: you are not interested in providing a PR a little richer, than integration events, are you? ;-) Right now, we don't really have a good reward model yet for partners who actively contribute, other than a mention at launch events or the like, but I'm thinking about ways to make it worthwhile for the community, to contribute with a little more, than events.
On that note, I wish you a great weekend. If nothing else, we'll process this PR to enable you to add your very cool address book :-)

quadfin:
@JesperSchulz-
Thanks for the reply and insight.
I am a bit uncertain what this means: “…you are not interested in providing a PR a little richer, than integration events, are you?”
Are you are asking if I am interested to open source my implementation?
If so, the answer is yes, as long as I have the ability to position my (paid) product as a more capable successor.
If you would like to discuss in more detail can we take this conversation offline?
I have your direct email from a previous conversation – should I contact you?
Regards,
Phil

JesperSchulz:
Please do so, yes :-) I'd be happy to have a call with you! My week actually looks pretty open, for once (summer vacation time kicking in, I guess). Feel free to book a 30 min call at any time!

JesperSchulz:
Update: we're going to push a PR of an alpha version of our address book on GitHub later today. It's far from complete, but should give you an idea of in which direction we're heading. Of course we would be excited if you chip in and help us complete creating an address book, that  our customers will love, and that you can extend to become even greater :-) I will go on an extended summer break soon, but others on my team should be able to assist, if you have any questions/feedback. Looking forward to collaborate with you on this!

JesperSchulz:
Let's track further progress/requirements in the new PR #13165.

JesperSchulz:

Issue #12833, opened 2021-06-12
[Feature Request] Possibility to refresh a OBO token from cache for another Tenant


jwikman:
I need to communicate to other BC Tenants through Web Services from within a BC tenant. The user is licensed and has permissions in all Tenants, so that should be possible.

Scenario:
Tenant0: An administrative tenant where the user belong. Let's call the user "Admin"
Tenant1, Tenant2, etc.: Other tenants that should be handled from Tenant0. 
Admin will call Tenant1 through WS to start different tasks.
Some tasks that Admin starts in Tenant1 takes a long time, so the WS call can not wait for it to finish since the WS would time out.
In Tenant0 a background session will need to check now and then if the task in Tenant1 is completed. 

We use the On-Behalf-Of flow to solve above.

I use `OAuth2.AcquireOnBehalfAccessTokenAndTokenCache()` in Tenant0 to get the Access Token and Token Cache for Tenant1. This works great, the WS call gets authenticated and a task gets started in Tenant1.
As long as the Access Token is valid the background session can use the same Access Token to call Tenant1 to check if task is completed. 

BUT if the Access Token is expired, the background session would need to get a new Access Token, using the Token Cache returned by `OAuth2.AcquireOnBehalfAccessTokenAndTokenCache()`. To get a new token from that cache we would use `OAuth2.AcquireOnBehalfOfTokenByTokenCache(LoginHint,..)`. `OAuth2.AcquireOnBehalfOfTokenByTokenCache(LoginHint,..)` works fine as long you are requesting a token for the tenant you are running in, but it cannot handle requests for Access Tokens for another tenant....

I would like to have an overloaded function for `OAuth2.AcquireOnBehalfOfTokenByTokenCache(LoginHint,..)` that also have the `OAuthAuthorityUrl` parameter, like this:
```
    [NonDebuggable]
    procedure AcquireOnBehalfOfTokenByTokenCache(OAuthAuthorityUrl: Text; LoginHint: Text; RedirectURL: Text; Scopes: List of [Text]; RefreshToken: Text; var AccessToken: Text; var NewRefreshToken: Text)
    var
        ScopesArray: DotNet StringArray;
    begin
        FillScopesArray(Scopes, ScopesArray);
        Initialize(OAuthAuthorityUrl, RedirectURL);
        AccessToken := AuthFlow.ALAcquireTokenFromTokenCacheState(ScopesArray, LoginHint, RefreshToken, NewRefreshToken);
    end;
```
[...]

PR #12960, opened 2021-06-23
 Possibility to refresh a OBO token from cache for another Tenant

jwikman:
Implements #12833.

There are no tests of the OAuth2 module since before, so there are no tests for this either.
I'm having a hard time to get the OBO functions in the OAuth2 module to work in my docker environment (they all work fine in SaaS), so I don't have the possibility to test this code at all. But it _should_ work, according to how the other code is working.

If someone could push me in the right direction to get the OBO flow to work in a docker environment I could confirm that this works as intended.

### DIFF
--- a/Modules/System/OAuth2/OAuth2.Codeunit.al
+++ b/Modules/System/OAuth2/OAuth2.Codeunit.al
@@ -551,6 +551,23 @@ codeunit 501 OAuth2
     /// <summary>
     /// Gets the token and token cache via the On-Behalf-Of OAuth2 v1.0 protocol flow. 
     /// </summary>
+    /// <param name="OAuthAuthorityUrl">The identity authorization provider URL.</param>
+    /// <param name="LoginHint">The user login hint, i.e. authentication email.</param>
+    /// <param name="RedirectURL">The redirectURL of your app, where authentication responses can be sent and received by your app. It must exactly match one of the redirectURLs you registered in the portal. If this parameter is empty, the default Business Central URL will be used.</param>
+    /// <param name="Scopes">A list of scopes that you want the user to consent to.</param>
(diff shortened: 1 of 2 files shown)


jwikman:
FYI, I've really tried to make this work on Docker but constantly hit the wall. And since this is related to security, it is **very** hard to debug, since I do not get any good error messages.
Last month I created an issue in the BCTech repo, to get any help with this: https://github.com/microsoft/BCTech/issues/66  
But no response there yet...

JesperSchulz:
Thanks Johannes! We'll process your PR :-)

IhorHandziuk:
Hi Johannes!

Sorry for a little back-and-forth with the decision on this functionality.
We no longer recommend to use functions in the OAuth2 module that _do not_ accept CliendID + ClientSecret or ClientID + certificate, so unfortunately we cannot move forward with this PR.

We believe the best way to cover your scenario would be to create an Azure app registration and use the functions AcquireTokenByAuthorizationCode / AcquireAuthorizationCodeTokenFromCache from the OAuth2 module (similarly to the OnPrem logic of [EmailOAuthClient.Codeunit.al](https://github.com/microsoft/ALAppExtensions/blob/master/Apps/W1/Email%20-%20Outlook%20REST%20API/app/src/EmailOAuthClient.Codeunit.al)). 
Please note, that these functions also have the advantage of not having to deal with refreshing the token "manually" (e. g. there is no need to write any special logic for refreshing the token in the background session, the logic will work fine both in foreground and in background sessions 😊)

jwikman:
Hi Ihor!

I added our solution to this problem in #12833.

We will just abandon this PR not then, now that we've got a proper solution in place.

Sometimes the road ahead is not that straight as we wish, but at least now we get safely to the destination. 😀

Thanks,
Johannes

PR #13725, opened 2021-08-02
GetOutboxEmailsForRec., GetOutboxEmailRecordStatus

tyurm:
issue 13699

two new procedures and related tests were added:
GetOutboxEmailsForRecord - gets Outbox Emails for Source Record
GetOutboxEmailRecordStatus - gets Email Status of record in Outbox Emails table

### DIFF
--- a/Modules/System Tests/Email/src/EmailTest.Codeunit.al
+++ b/Modules/System Tests/Email/src/EmailTest.Codeunit.al
@@ -1207,7 +1207,6 @@ codeunit 134685 "Email Test"
     end;
 
     [Test]
-    [Scope('OnPrem')]
     procedure GetSourceRecordInOutbox()
     var
         SourceEmailOutbox, EmailOutbox : Record "Email Outbox";
@@ -1219,42 +1218,33 @@ codeunit 134685 "Email Test"
         SystemId: Guid;
         TableId: Integer;
         NumberOfEmails, i : Integer;
-        RecRef: RecordRef;
     begin
         BindSubscription(EmailTest);
-
         PermissionsMock.Set('Email Edit');
-
         // [Scenario] Emails with source document, GetEmailOutboxForRecord procedure will return Outbox Emails
         // [Given] Source Record - Email Outbox used as a source record for test email
         CreateEmail(EmailMessage);
         Email.SaveAsDraft(EmailMessage, SourceEmailOutbox);
         TableId := Database::"Email Outbox";
         SystemId := SourceEmailOutbox.SystemId;
-
         // [When] Several emails are created and saved as draft
         NumberOfEmails := Any.IntegerInRange(2, 5);
-
         for i := 1 to NumberOfEmails do begin
             Clear(EmailOutbox);
             CreateEmailWithSource(EmailMessage, TableId, SystemId);
             Email.SaveAsDraft(EmailMessage, EmailOutbox);
             MessageIds.Add(EmailMessage.GetId());
         end;
-
         // [Then] GetEmailOutboxForRecord procedure return related Email Outbox
         Email.GetEmailOutboxForRecord(SourceEmailOutbox, TempEmailOutbox);
         Assert.AreEqual(NumberOfEmails, TempEmailOutbox.Count(), 'Email Outbox count is not equal to Number of Emails created.');
-
         for i := 1 to NumberOfEmails do begin
             TempEmailOutbox.SetCurrentKey("Message Id");
             TempEmailOutbox.SetRange("Message Id", MessageIds.Get(i));
             Assert.AreEqual(1, TempEmailOutbox.Count(), 'Did not find the email in Email Outbox');
         end;
     end;
-
     [Test]
-    [Scope('OnPrem')]
     procedure GetEmailOutboxRecordStatus()
     var
         EmailOutbox: Record "Email Outbox";
@@ -1266,29 +1256,22 @@ codeunit 134685 "Email Test"
         SystemId: Guid;
     begin
         BindSubscription(EmailTest);
-
         PermissionsMock.Set('Email Edit');
-
         // [Scenario] Emails with source document, GetOutboxEmailRecordStatus will return Outbox Email Status
         // [Given] An Email with table id and source system id
         TableId := Any.IntegerInRange(1, 10000);
         SystemId := Any.GuidValue();
-
         // [When] The email is created and saved as draft
         CreateEmailWithSource(EmailMessage, TableId, SystemId);
-
         // [When] The email is created and saved as draft
         Email.SaveAsDraft(EmailMessage, EmailOutbox);
-
         // [Then] Email Status of created Email Outbox record is equal to GetOutboxEmailRecordStatus result
         EmailStatus := Email.GetOutboxEmailRecordStatus(EmailOutbox."Message Id");
         Assert.AreEqual(EmailStatus, EmailOutbox.Status, 'Email Status should be the same as on Email Outbox record');
     end;
-
     [Test]
     procedure GetSentEmailsForRecordByVariant()
     var
-        SourceEmailOutbox: Record "Email Outbox";
         TempSentEmail: Record "Sent Email" temporary;
         Account: Record "Email Account";
         EmailMessage: Codeunit "Email Message";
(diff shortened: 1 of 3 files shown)


Modules/System/Email/src/Email/Email.Codeunit.al
```diff
@@ -239,6 +239,17 @@ codeunit 8901 "Email"
         exit(EmailImpl.GetSentEmailsForRecord(TableId, SystemId));
     end;
 
+    ///<summary>
+    /// Gets the outbox emails related to a record.
+    ///</summary>
+    ///<param name="TableId">The table ID of the record.</param>
+    ///<param name="SystemId">The system ID of the record.</param>
+    ///<returns>The outbox emails related to a record.</returns>
+    procedure GetEmailOutboxForRecord(TableId: Integer; SystemId: Guid) ResultEmailOutbox: Record "Email Outbox" temporary;
```
GiorgosPa, line 248: This results in copying the result record to a new instance and unfortunately this only copies the fields and filters not the recordset. There is a nice [blogpost ](https://www.waldo.be/2021/07/06/the-complexity-of-complex-return-types/) that describes the issue in depth

GiorgosPa, line 248: It would be better to get a Record Variant as a parameter, these values can be retrieved from the record variant and it would be a cleaner and simpler  interface.

Modules/System Tests/Email/src/EmailTest.Codeunit.al
```diff
@@ -1206,6 +1206,70 @@ codeunit 134685 "Email Test"
         Assert.ExpectedError(EmailMessageOpenPermissionErr);
     end;
 
+    [Test]
```
GiorgosPa, line 1209: A test is needed that verifies that at least two email outbox can be retrieved and looped over

Modules/System/Email/src/Email/EmailImpl.Codeunit.al
```diff
@@ -320,6 +320,50 @@ codeunit 8900 "Email Impl"
...
+        EmailRelatedRecord.SetRange("System Id", SystemId);
+
+        if not EmailRelatedRecord.FindSet() then
+            exit;
+
+        if not EmailAccountImpl.IsUserEmailAdmin() then
+            EmailOutbox.SetRange("User Security Id", UserSecurityId());
+
+        repeat
+            EmailOutbox.SetCurrentKey("Message Id");
```
GiorgosPa, line 344: this is not needed

Modules/System/Email/src/Email/EmailImpl.Codeunit.al
```diff
@@ -320,6 +320,50 @@ codeunit 8900 "Email Impl"
...
+
+    procedure GetOutboxEmailRecordStatus(MessageId: Guid) ResultStatus: Enum "Email Status"
+    var
+        EmailOutbox: Record "Email Outbox";
+        EmailAccountImpl: Codeunit "Email Account Impl.";
+    begin
+        if not EmailAccountImpl.IsUserEmailAdmin() then
+            EmailOutbox.SetRange("User Security Id", UserSecurityId());
+
+        EmailOutbox.SetCurrentKey("Message Id");
```
GiorgosPa, line 361: not needed

Modules/System/Email/src/Email/EmailImpl.Codeunit.al
```diff
@@ -320,6 +320,50 @@ codeunit 8900 "Email Impl"
...
+        EmailOutbox.SetRange("Message Id", MessageId);
+        if EmailOutbox.FindFirst() then
```
GiorgosPa, line 363: an error is more appropriate in case there is no Outbox Record

GiorgosPa (changes requested):

Modules/System/Email/src/Email/Email.Codeunit.al
tyurm: initially, I have created a copy of **GetSentEmailsForRecord** function, with only difference in the resulting record.
I totally agree with your idea that it would be better to return results by reference, so I have created additional overload for **GetEmailOutboxForRecord** and **GetSentEmailsForRecord** procedures in Email Codeunit to keep the same logic for both procedures

tyurm: Same as in previous comment - two additional overloads for **GetEmailOutboxForRecord** and **GetSentEmailsForRecord** procedures were added to keep the same logic for both procedures. But I got an issue in test function. Will describe below.

Modules/System/Email/src/Email/EmailImpl.Codeunit.al
tyurm: removed, but same logic still exist in **GetSentEmailsForRecord** procedure

Modules/System/Email/src/Email/EmailImpl.Codeunit.al
tyurm: Removed

Modules/System Tests/Email/src/EmailTest.Codeunit.al
tyurm: I have changed test according to your request, but I got another issue with "RecRef" parameter. Standard tests use random values in TableId and SystemId of source records:

```
TableId := Any.IntegerInRange(1, 1000);
SystemId := Any.GuidValue();
```
Using this we will always get errors on RecRef.Open and RecRef.GetBySystemId
In last commit, I have used the first record from the Company table as a source for TableId and SystemId, but I think it is inappropriate.

I would assume to use the previous overload of GetEmailOutboxForRecord(TableId,SystemID) procedure for testing

Modules/System/Email/src/Email/EmailImpl.Codeunit.al
tyurm: I think it is better to keep same behaviour for both **GetSentEmailsForRecord** and **GetEmailOutboxForRecord** procedures.
Should I add errors for both procedures, keep them without error, or change only new one?

GiorgosPa (changes requested):

Modules/System/Email/src/Email/EmailImpl.Codeunit.al
GiorgosPa: It is not the same. In the cases above an empty list of records is returned, while here the first Email Status is returned which is not the same as the Outbox record does not exist/ or got deleted.

Modules/System/Email/src/Email/Email.Codeunit.al
```diff
@@ -239,6 +239,38 @@ codeunit 8901 "Email"
         exit(EmailImpl.GetSentEmailsForRecord(TableId, SystemId));
     end;
 
+    ///<summary>
+    /// Gets the sent emails related to a record.
+    ///</summary>
+    ///<param name="RecRef">RecordRef of Source Record.</param>
+    ///<param name="ResultSentEmails">The sent email related to a record.</param>
+    procedure GetSentEmailsForRecord(RecRef: RecordRef; var ResultSentEmails: Record "Sent Email" temporary);
```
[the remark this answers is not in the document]
tyurm: Test GetSentEmailsForRecordByVariant was added

Modules/System/Email/src/Email/Email.Codeunit.al
tyurm: I have removed 

` GetEmailOutboxForRecord(TableId: Integer; SystemId: Guid) ResultEmailOutbox: Record "Email Outbox" temporary;`

and changed procedure GetEmailOutboxForRecord to use Variant instead of RecordRef: 
`GetEmailOutboxForRecord(RecordVariant: Variant; var ResultEmailOutbox: Record "Email Outbox" temporary);`

Modules/System/Email/src/Email/EmailImpl.Codeunit.al
tyurm: sorry, understood it worng. I've removed "If Then" condition

tyurm:
@GiorgosPa Hello, All requested changes were added, could you please check if everything is ok now?

GiorgosPa (approved):
That looks great, thank you.

JesperSchulz:

PR #13731, opened 2021-08-03
Added Xml Signature validation

kennetlindberg:
This adds the ability to validate/check a signature in a Xml Document to the SignedXml codeunits. There are three overloads of the new procedure CheckSignature and a new procedure called LoadXml that is needed to load the signature element before checking. It follows the structures and procedures of the DotNet implementation closely.

I had to change the visibility of InitializeX509Certificate() in codeunit 1285 "X509Certificate2 Impl." from local to internal to enable loading a X509Certificate to use when validating the signature.

I also had to make some modifications to codeunit 1462 "Xml DotNet Convert" where I set PreserveWhitespace to true in a few places. Without this it is not possible to validate signatures since the Xml Document needs to be exactly identical in order for the signature to be valid.


kennetlindberg:
Is there a plan for when this might be looked at? It has been almost two months since the PR. Can I do something?

darjoo:
Hi @kennetlindberg, I'll reach out to the owning team and ask them to give an update.

PaulIvan:
Hi @kennetlindberg we'll start on this next week. sorry for the delay :)

DavidHolusa:
Hi @kennetlindberg. Why is not possible validate signature without preserverwhitespace property? It depends whether the xml has been signed with or without preservewhitespace. Or not?

kennetlindberg:
That is not relevant here. This is used in the conversion from BC Native XmlDocument datatype to DotNet XmlDocument datatype and vice versa. If the conversion would change the xml content it would not be possible to validate the signature. Using PreserveWhitespace ensures that this does not happen.

The conversion writes the XmlDocument (dotnet or BC) to a text (string) variable and then reads it back to an XmlDocument (BC or dotnet). In order to preserve the exact same Xml the PreserveWhitespace is used both when writing to the string and reading from it.

PR #14026, opened 2021-08-17
Message added as dimension to AddLogAppInsights

Bertverbeek4PS:
If the test is getting in an error you want also investigate the message.
This because the docker system will be removed because the performance toolkit you run periodic and you want also save the message.
@BardurKnudsen


JesperSchulz:
We are currently discussing, if your PR could have privacy issues. Error messages can very easily contain PII (personal identifiable information), which we may not store in our telemetry - e.g. "The contact John Doe could not get updated". We will get back to you with a conclusion very soon.

JesperSchulz:
As we don't have and control over what information is contained in the error message, I'm afraid we cannot emit this information into our telemetry store. I regret that I have to reject this PR :-(

PR #14268, opened 2021-09-08
Enabling Date only Dialog

miljance:


mazhelez:
Hey, @miljance 

What is the use case for a date-only dialog? There's already a date-picker by default.

Modules/System/Date-Time Dialog/src/DateTimeDialog.Page.al
```diff
@@ -63,5 +68,24 @@ page 684 "Date-Time Dialog"
...
     end;
+
+    /// <summary>
+    /// Setter method to initialize the Date on the page.
+    /// </summary>
+    /// <param name="Date">The value to set.</param>
+    procedure SetDate(Date: Date)
+    begin
+        DateValue := Date;
+        TimeVisible := false;
```
mazhelez, line 79: I'd rather have a separate explicit function to use the page as a date-picker.

Something like
```
procedure UseDateOnly()
```

mazhelez (commented):
The code itself looks okay.

However, I really like to question the need for this solution.
As I mentioned on my previous comment, the client supports a date-picker on a field if the underlying variable is of type date.

Modules/System/Date-Time Dialog/src/DateTimeDialog.Page.al
mazhelez: @miljance Generally, whenever you take something simple as the date-time dialog and try to make a date-only dialog, there'll be some confusion.

My idea of the API was:
(_existing_)
SetDateTime
GetDateTime

(_new_)
UseDateOnly()
GetDate
SetDate

This still leaves the question of what should happen if someone does `SetDate` and then 'GetDatetime` or vice versa.
One way to handle such cases is to "hide" the page behind managing codeunits (e.g. "Date Picker" and "Date-Time Picker") and each codeunit does whatever it is supposed to.

This might be considered over-engineering but handles the usages quite nicely and covers for future requests like "Time Picker" ;)

miljance: The suggested method names make sense. I will change them accordingly.

Do you really think additional codeunits make sense? It does sound as over-engineering to be honest.
It does abstract the implementation and if you confirm that this approach is aligned with general guidelines for System App I can change it accordingly. The other Idea would be to separate the Date-Picker from DateTime-Picker completely. I think in this case the implementation would be simpler without needing to abstract that much. Perhaps with two separate implementations the code would be much cleaner.

darjoo:
@miljance I've reopened it for you :)

Modules/System/Date-Time Dialog/src/DateTimeDialog.Page.al
mazhelez: > Do you really think additional codeunits make sense?

It does make _some_ sense. This feature is way to simple for that, but I'll accept if you do it. 

> The other Idea would be to separate the Date-Picker from DateTime-Picker completely.

Like... separate pages? That's definitely a no-go. Pages are too precious to be spent like that, especially when they'll have so much in common 😃 

Generally, it is up to you. I just suggested being explicit on when to use the date picker and when to use the date-time picker. `SetDate` implies that the date will be set before the page is opened, it doesn't imply that the datetime picker becomes a date picker.

JesperSchulz:

JesperSchulz:
Reopening to merge.

JesperSchulz:

PR #14428, opened 2021-09-23
Implement RSA encryption and decryption

maciejmlocek:
This adds the ability to Encrypt and Decrypt Data using RSA algorithm by wrapping two methods from RSACryptoServiceProvider DotNet Class.

### DIFF
--- a/Modules/System/Cryptography Management/src/RSACryptoServiceProvider.Codeunit.al
+++ b/Modules/System/Cryptography Management/src/RSACryptoServiceProvider.Codeunit.al
@@ -47,4 +47,28 @@ codeunit 1445 "RSACryptoServiceProvider"
     begin
         exit(RSACryptoServiceProviderImpl.VerifyData(XmlString, DataInStream, HashAlgorithm, SignatureInStream));
     end;
+
+    /// <summary>
+    /// Encrypts the specified text.
+    /// </summary>
+    /// <param name="XmlString">The XML string containing RSA key information.</param>
+    /// <param name="PlainTextInStream">The input stream to encrypt.</param>
+    /// <param name="RSAEncryptionPadding">The RSA encryption padding to use to encrypt the specified text.</param>
+    /// <param name="EncryptedTextOutStream">The RSA encryption stream for the specified text.</param>
+    procedure Encrypt(XmlString: Text; PlainTextInStream: InStream; RSAEncryptionPadding: Enum RSAEncryptionPadding; EncryptedTextOutStream: OutStream)
+    begin
+        RSACryptoServiceProviderImpl.Encrypt(XmlString, PlainTextInStream, RSAEncryptionPadding, EncryptedTextOutStream);
+    end;
+
+    /// <summary>
+    /// Decrypts the specified text.
+    /// </summary>
+    /// <param name="XmlString">The XML string containing RSA key information.</param>
+    /// <param name="EncryptedTextInStream">The input stream to decrypt.</param>
+    /// <param name="RSAEncryptionPadding">The RSA encryption padding to use to decrypt the specified text.</param>
+    /// <param name="DecryptedTextOutStream">The RSA decryption stream for the specified text.</param>
+    procedure Decrypt(XmlString: Text; EncryptedTextInStream: InStream; RSAEncryptionPadding: Enum RSAEncryptionPadding; DecryptedTextOutStream: OutStream)
+    begin
+        RSACryptoServiceProviderImpl.Decrypt(XmlString, EncryptedTextInStream, RSAEncryptionPadding, DecryptedTextOutStream);
+    end;
 }
--- a/Modules/System/Cryptography Management/src/RSACryptoServiceProviderImpl.Codeunit.al
+++ b/Modules/System/Cryptography Management/src/RSACryptoServiceProviderImpl.Codeunit.al
@@ -87,6 +87,51 @@ codeunit 1446 "RSACryptoServiceProvider Impl." implements SignatureAlgorithm
     end;
     #endregion
 
+    #region Encryption
+    procedure Encrypt(XmlString: Text; PlainTextInStream: InStream; RSAEncryptionPadding: Enum RSAEncryptionPadding; EncryptedTextOutStream: OutStream)
+    var
+        PlainTextBytes: DotNet Array;
+        EncryptedTextBytes: DotNet Array;
+        DotNetRSAEncryptionPadding: DotNet RSAEncryptionPadding;
+    begin
+        FromXmlString(XmlString);
+        InStreamToArray(PlainTextInStream, PlainTextBytes);
+        SetRSAEncryptionPadding(DotNetRSAEncryptionPadding, RSAEncryptionPadding);
+        EncryptedTextBytes := DotNetRSACryptoServiceProvider.Encrypt(PlainTextBytes, DotNetRSAEncryptionPadding);
+        ArrayToOutStream(EncryptedTextBytes, EncryptedTextOutStream);
+    end;
+
+
+    procedure Decrypt(XmlString: Text; EncryptedTextInStream: InStream; RSAEncryptionPadding: Enum RSAEncryptionPadding; DecryptedTextOutStream: OutStream)
+    var
+        EncryptedTextBytes: DotNet Array;
+        DecryptedTextBytes: DotNet Array;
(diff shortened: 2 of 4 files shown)


darjoo:
@PredragMaricic

darjoo:
@maciejmlocek Predrag or someone on his team will do the main review, but please add some tests :)

maciejmlocek:
> @maciejmlocek Predrag or someone on his team will do the main review, but please add some tests :)

@darjoo okay, but it is hard to write automated tests for padding encryption as you don't actually know what the expected value is going to be and RSACryptoServiceProvider doesn't have methods for encryption without padding.

The only scenario I can think of is that the Decrypt method returns the encrypted plain text :)

darjoo:
Is there no way to have specific values just for the test? 
Below is a hastily written test, but wouldn't something like that be possible? To test both encrypt and decrypt.

````
procedure Test()
var
    TempBlobWrite: Codeunit "Temp Blob";
    TempBlobEncrypt: Codeunit "Temp Blob";
    TempBlobDecrypt: Codeunit "Temp Blob";
    RSA: Codeunit "RSACryptoServiceProvider";
    XmlString: Text;
    EncryptInStream: InStream;
    EncryptOutStream: OutStream;
    DecryptInStream: InStream;
    DecryptOutStream: OutStream;
    WriteOutStream: OutStream;
    Padding: Boolean;
    StringToEncrypt: Label 'Please encrypt me.';
    EncryptedString: Label 'Someencryptedstringvalue';
    CompareString: Text;
begin
    XmlString := 'RSA Something';
    Padding := true; // (Or false, I'm not sure how this affects the output)
    TempBlobWrite.CreateOutStream(WriteOutStream);
    WriteOutStream.Write(StringToEncrypt);
    TempBlobWrite.CreateInStream(EncryptInStream);
    TempBlobEncrypt.CreateOutStream(EncryptOutStream);
    RSA.Encrypt(XmlString, EncryptInStream, Padding, EncryptOutStream);
    TempBlobEncrypt.CreateInStream(DecryptInStream);
    DecryptInStream.ReadText(CompareString);
    Assert.AreEqual(EncryptedString, CompareString, 'The encrypted strings are not equal');
    TempBlobEncrypt.CreateInStream(DecryptInStream);    
    TempBlobDecrypt.CreateOutStream(DecryptOutStream);
    RSA.Decrypt(XmlString, DecryptInStream, Padding, DecryptOutStream);
    TempBlobDecrypt.CreateInStream(DecryptInStream);
    DecryptInstream.ReadText(CompareString);
    Assert.AreEqual(StringToEncrypt, CompareString, 'The strings are not equal');
end;
````

**Comment by maciejmlocek** [comment]
@darjoo 

Unfortunately we cannot predict the encrypted value of some text.
For example:
Let's say we have EncryptMe text with the value "encrypt me" and we are encrypting this text with the padding mode on. 
You get some encryptedTextValue.
But now, when you encrypt the same EncryptMe text again, then you (most) probably will get different encrypted text.

We cannot turn the padding off with the RSACryptoServiceProvider DotNet Class. You either choose OAEP Padding or PKCS#1 Padding. 

I will provide few tests for decrypting, but I have no idea how to check if encryption works fine. But if this is just calling DotNet Class Method I would assume that this is correct - I know, that it is not good approach to assume that something works fine, though. 😄

**Comment by darjoo** [comment]
Ah okay, I see what you mean. Thanks for the clarification! The security stuff isn't my area :D 
I guess yeah, if we can't guarantee a certain output from encryption, I guess using any of them and then providing only the tests for the decryption should be valid and enough :)

**Comment by maciejmlocek** [comment]
@darjoo
Tests added. Do you have any idea on when can we expect this to be merged? We really need that implementation for our (@wokw) Polish Functionality - Starter Pack Application. It is one of requirements to send something encrypted with RSA through polish government API.

**Comment by darjoo** [comment]
I can't give a timeline unfortunately, but as soon as the owning team gets to it. 
@PredragMaricic

**Comment by PredragMaricic** [comment]
Hi
We will start the process tomorrow, I hope to have it reviewed before Friday.

**Comment by maciejmlocek** [comment]
@PredragMaricic that's great news, thanks for letting me know!

File: Modules/System Tests/Cryptography Management/src/RSACryptoServiceProviderTests.Codeunit.al [hunk 1]
```diff
@@ -0,0 +1,183 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+codeunit 132602 RSACryptoServiceProviderTests
```
PredragMaricic, line 5: 132602 already occupied, I will change it.

Modules/System Tests/Cryptography Management/app.json
```diff
@@ -29,6 +29,18 @@
                              "publisher":  "Microsoft",
                              "version":  "18.3.0.0"
                          },
+                         {
+                            "id": "5d86850b-0d76-4eca-bd7b-951ad998e997",
+                            "name": "Tests-TestLibraries",
```
PredragMaricic, line 34: Unfortunately, it is not possible to reference "Tests-TestLibraries", so LibraryRandom should be replaced with something else here.

maciejmlocek: Can you suggest what other library I can use for randomization? Should implement my own procedure for it or just hardcode some text without randomization?

Modules/System Tests/Cryptography Management/src/RSACryptoServiceProviderTests.Codeunit.al
maciejmlocek: ok, sure

Modules/System Tests/Cryptography Management/app.json
maciejmlocek: I've replaced that with "Any" codeunit. 
By the way, why it is not possible to reference "Tests-TestLibraries" is this application marked for removal, so we should not use it in our projects?

maciejmlocek (commented):
Changed reference from Library Random to "Any" Codeunit. 
it's up to @PredragMaricic to renumber the test codeunit to not occupied ID.

Modules/System Tests/Cryptography Management/app.json
darjoo: "Test-TestLibraries" is part of BaseApp's tests. The System Application modules & its tests should not have any dependency to BaseApp or BaseApp's test libraries.

Basically BaseApp & its Tests/TestLibraries can depend on System Application etc but not the otherway around. :)

maciejmlocek: Okay, I see. Thanks for explanation!

maciejmlocek:
@PredragMaricic there are conflicting files, but it will be resolved as soon as you change or tell me what number for Codeunit RSACryptoServiceProviderTests I can use, so I am not resolving it so far.

PredragMaricic:
> @PredragMaricic there are conflicting files, but it will be resolved as soon as you change or tell me what number for Codeunit RSACryptoServiceProviderTests I can use, so I am not resolving it so far.

Which files are conflicting? I fixed some smaller details and I am actually pushing your changes into master right now.

maciejmlocek:
app.json for testing app is conflicting

maciejmlocek:
@PredragMaricic how are we doing on that?

PredragMaricic:
> @PredragMaricic how are we doing on that?

Submitted and tested in Master. Submission to 19 in progress. Will be delivered with 19.1.

maciejmlocek:
> > @PredragMaricic how are we doing on that?
> 
> Submitted and tested in Master. Submission to 19 in progress. Will be delivered with 19.1.

Great, thanks!
Are we going to fix those details here and close the PR?

PredragMaricic:
> > > @PredragMaricic how are we doing on that?
> > 
> > 
> > Submitted and tested in Master. Submission to 19 in progress. Will be delivered with 19.1.
> 
> Great, thanks! Are we going to fix those details here and close the PR?

Once submitted, the PR will be closed and the changes will be shared. Expected time - today EOD.

PredragMaricic (approved):

JesperSchulz:

JesperSchulz:

PR #14713, opened 2021-10-08
Changed the PutBlobBlockBlobUI implementation to work in the UI

ghost:
Slightly changed the PutBlobBlockBlobUI function. The current version does not work as `'*.*'` is an invalid filter string, and the Filename variable is never set. Furthermore, according to https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/file/file-uploadintostream-string-instream-method#remarks, it is recommended to use the File Management Codeunit to upload/download files.

### DIFF
--- a/Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
+++ b/Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
@@ -116,11 +116,16 @@ codeunit 9051 "ABS Client Impl."
     procedure PutBlobBlockBlobUI(OptionalParameters: Codeunit "ABS Optional Parameters"): Codeunit "ABS Operation Response"
     var
         OperationResponse: Codeunit "ABS Operation Response";
+        TempBlob: Codeunit "Temp Blob";
         Filename: Text;
         SourceStream: InStream;
+        ChooseFileTitleMsg: Label 'Choose the file to upload.';
     begin
-        if UploadIntoStream('*.*', SourceStream) then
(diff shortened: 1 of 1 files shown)


Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
```diff
@@ -116,11 +116,16 @@ codeunit 9051 "ABS Client Impl."
     procedure PutBlobBlockBlobUI(OptionalParameters: Codeunit "ABS Optional Parameters"): Codeunit "ABS Operation Response"
     var
         OperationResponse: Codeunit "ABS Operation Response";
+        TempBlob: Codeunit "Temp Blob";
         Filename: Text;
         SourceStream: InStream;
+        ChooseFileTitleMsg: Label 'Choose the file to upload.';
     begin
-        if UploadIntoStream('*.*', SourceStream) then
+        Filename := FileManagement.BLOBImportWithFilter(TempBlob, ChooseFileTitleMsg, '', 'All Files (*.*)|*.*', '*.*');
```
mazhelez, line 124: Where does `FileManagement` come from?

kevinsky25: Ah, you're 100% right about the File Management codeunit being in the base application. I did my testing in a separate extension. Makes sense it doesn't work. Excuse me!

I assume the fix for the UploadIntoStream function is in a different branch? Because it appears to be `if UploadIntoStream('*.*', SourceStream) then` in the master branch:

https://github.com/microsoft/ALAppExtensions/blob/main/Modules/System/Azure%20Blob%20Services%20API/src/ABSClientImpl.Codeunit.al#L122

PR #14845, opened 2021-10-17
Regex allowing precompiled patterns for subsequent calls

MODUSCarstenScholling:
Often Regex needs subsequent calls with the same pattern and options. This PR allows to pass pattern and options to the constructor and offers additional procedures working on that prepared instance.
I have added 2 Test procedures at the end to allow a performance comparision. Using the same Regex DotNet instance wth the same pattern reduced the runtime by factor 60+ (from 1000 down to 15 milliseconds for 100 matches).

Added Regex() constructors
Added instance specific procedures
Added instance tests


darjoo (approved):
Very nice, I don't see any changed required and I'll process this internally now and try to get it in asap for 19.3

PR #15221, opened 2021-11-17
Add support for Lease operations on Blobs and Containers

SimonOfHH:
This PR adds support for [Lease Blob](https://docs.microsoft.com/en-us/rest/api/storageservices/lease-blob)- and [Lease Container](https://docs.microsoft.com/en-us/rest/api/storageservices/lease-container)-operations for the "Azure Blob Services API".

The tests contain one deactivated part, because the operation ("Lease Change") works against a "real" Azure Storage Account, but not against the Azurite emulator (I'll try to create a PR for the bug in Azurite as well).

### DIFF
--- a/Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
+++ b/Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
@@ -119,13 +119,13 @@ codeunit 132919 "ABS Container Client Test"
         Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseRenew failed');
 
         // This part works when testing against a "real" Azure Storage and not Azurite
-        // leaving this here for completeness
-        // // [3rd] Change Lease on Container
-        // ProposedLeaseId := CreateGuid();
-        // Response := ABSContainerClient.LeaseChange(ContainerName, LeaseId, ProposedLeaseId);
-        // Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseChange failed');
-        // LeaseId := Response.GetLeaseId();
-        // Assert.IsFalse(IsNullGuid(LeaseId), 'Operation LeaseChange failed (no LeaseId returned)');
+        if (AzuriteTestLibrary.GetStorageAccountName() <> 'devstoreaccount1') then begin // "devstoreaccount1" is the hardcoded name for Azurite test-account
+            // [3rd] Change Lease on Container
+            ProposedLeaseId := CreateGuid();
+            Response := ABSContainerClient.LeaseChange(ContainerName, LeaseId, ProposedLeaseId);
+            Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseChange failed');
+            Assert.IsFalse(IsNullGuid(LeaseId), 'Operation LeaseChange failed (no LeaseId returned)');
+        end;
 
         // [4th] Break Lease on Container
         Response := ABSContainerClient.LeaseBreak(ContainerName, LeaseId);
--- a/Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
+++ b/Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
@@ -110,9 +110,8 @@ codeunit 132919 "ABS Container Client Test"
         Assert.IsTrue(Response.IsSuccessful(), 'Operation CreateContainer failed');
 
         // [1st] Acquire Lease on Container
-        Response := ABSContainerClient.LeaseAcquire(ContainerName, 60);
+        Response := ABSContainerClient.LeaseAcquire(ContainerName, 60, LeaseId);
         Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseAcquire failed');
-        LeaseId := Response.GetLeaseId();
         Assert.IsFalse(IsNullGuid(LeaseId), 'Operation LeaseAcquire failed (no LeaseId returned)');
 
         // [2nd] Renew Lease on Container
--- a/Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
+++ b/Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
@@ -7,7 +7,7 @@ codeunit 132919 "ABS Container Client Test"
 {
     Subtype = Test;
 
-    //[Test]
+    [Test]
     procedure CreateContainerSharedKeyTest()
     var
         Response: Codeunit "ABS Operation Response";
@@ -29,7 +29,7 @@ codeunit 132919 "ABS Container Client Test"
         Assert.IsTrue(Response.IsSuccessful(), 'Operation DeleteContainer failed');
     end;
 
-    //[Test]
+    [Test]
     procedure CreateContainerFailedTest()
     var
         Response: Codeunit "ABS Operation Response";
@@ -54,7 +54,7 @@ codeunit 132919 "ABS Container Client Test"
         Assert.IsTrue(Response.IsSuccessful(), 'Operation DeleteContainer failed');
     end;
 
-    //[Test]
+    [Test]
     procedure ListContainersTest()
     var
         Containers: Record "ABS Container";
@@ -87,6 +87,57 @@ codeunit 132919 "ABS Container Client Test"
             Response := ABSContainerClient.DeleteContainer(ContainerName);
             Assert.IsTrue(Response.IsSuccessful(), 'Operation DeleteContainer failed');
         end;
+    end;
+
+    [Test]
+    procedure LeaseContainerTest()
+    var
+        Response: Codeunit "ABS Operation Response";
+        ContainerName: Text;
+        LeaseId: Guid;
+        ProposedLeaseId: Guid;
+    begin
+        // [Scenarion] Given a storage account and a container name, CreateContainer creates a container and subsequent lease-operations
+        // (1) create a lease, (2) renew a lease, [(3) change a lease], (4) break a lease and (5) release the lease        
+        SharedKeyAuthorization := StorageServiceAuthorization.CreateSharedKey(AzuriteTestLibrary.GetAccessKey());
+
+        ABSContainerClient.Initialize(AzuriteTestLibrary.GetStorageAccountName(), SharedKeyAuthorization);
+        ABSContainerClient.SetBaseUrl(AzuriteTestLibrary.GetBlobStorageBaseUrl());
+
+        ContainerName := ABSTestLibrary.GetContainerName();
+        Response := ABSContainerClient.CreateContainer(ContainerName);
+
+        Assert.IsTrue(Response.IsSuccessful(), 'Operation CreateContainer failed');
+
+        // [1st] Acquire Lease on Container
+        Response := ABSContainerClient.LeaseAcquire(ContainerName, 60);
+        Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseAcquire failed');
+        LeaseId := Response.GetLeaseId();
+        Assert.IsFalse(IsNullGuid(LeaseId), 'Operation LeaseAcquire failed (no LeaseId returned)');
+
+        // [2nd] Renew Lease on Container
+        Response := ABSContainerClient.LeaseRenew(ContainerName, LeaseId);
+        Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseRenew failed');
+
+        // This part works when testing against a "real" Azure Storage and not Azurite
+        // leaving this here for completeness
+        // // [3rd] Change Lease on Container
+        // ProposedLeaseId := CreateGuid();
+        // Response := ABSContainerClient.LeaseChange(ContainerName, LeaseId, ProposedLeaseId);
+        // Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseChange failed');
+        // LeaseId := Response.GetLeaseId();
+        // Assert.IsFalse(IsNullGuid(LeaseId), 'Operation LeaseChange failed (no LeaseId returned)');
+
+        // [4th] Break Lease on Container
+        Response := ABSContainerClient.LeaseBreak(ContainerName, LeaseId);
+        Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseBreak failed');
+
+        // [5th] Release Lease on Container
+        Response := ABSContainerClient.LeaseRelease(ContainerName, LeaseId);
+        Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseRelease failed');
+
+        // Clean-up
+        ABSContainerClient.DeleteContainer(ContainerName);
     end;
 
     var
--- a/Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
+++ b/Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
@@ -110,29 +110,29 @@ codeunit 132919 "ABS Container Client Test"
         Assert.IsTrue(Response.IsSuccessful(), 'Operation CreateContainer failed');
 
         // [1st] Acquire Lease on Container
-        Response := ABSContainerClient.LeaseAcquire(ContainerName, 60, LeaseId);
+        Response := ABSContainerClient.AcquireLease(ContainerName, 60, LeaseId);
         Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseAcquire failed');
         Assert.IsFalse(IsNullGuid(LeaseId), 'Operation LeaseAcquire failed (no LeaseId returned)');
 
         // [2nd] Renew Lease on Container
-        Response := ABSContainerClient.LeaseRenew(ContainerName, LeaseId);
+        Response := ABSContainerClient.RenewLease(ContainerName, LeaseId);
         Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseRenew failed');
 
         // This part works when testing against a "real" Azure Storage and not Azurite
         if (AzuriteTestLibrary.GetStorageAccountName() <> 'devstoreaccount1') then begin // "devstoreaccount1" is the hardcoded name for Azurite test-account
             // [3rd] Change Lease on Container
             ProposedLeaseId := CreateGuid();
-            Response := ABSContainerClient.LeaseChange(ContainerName, LeaseId, ProposedLeaseId);
+            Response := ABSContainerClient.ChangeLease(ContainerName, LeaseId, ProposedLeaseId);
             Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseChange failed');
             Assert.IsFalse(IsNullGuid(LeaseId), 'Operation LeaseChange failed (no LeaseId returned)');
         end;
 
         // [4th] Break Lease on Container
-        Response := ABSContainerClient.LeaseBreak(ContainerName, LeaseId);
+        Response := ABSContainerClient.BreakLease(ContainerName, LeaseId);
         Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseBreak failed');
 
         // [5th] Release Lease on Container
-        Response := ABSContainerClient.LeaseRelease(ContainerName, LeaseId);
+        Response := ABSContainerClient.ReleaseLease(ContainerName, LeaseId);
         Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseRelease failed');
 
         // Clean-up
--- a/Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
+++ b/Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
@@ -607,12 +607,12 @@ codeunit 9053 "ABS Blob Client"
     /// <param name="BlobName">The name of the blob.</param>
     /// <param name="LeaseId">Guid containing the response value from x-ms-lease-id HttpHeader</param>
     /// <returns>An operation reponse object</returns>
-    procedure LeaseAcquire(BlobName: Text; var LeaseId: Guid): Codeunit "ABS Operation Response"
+    procedure AcquireLease(BlobName: Text; var LeaseId: Guid): Codeunit "ABS Operation Response"
     var
         OptionalParameters: Codeunit "ABS Optional Parameters";
         ProposedLeaseId: Guid;
     begin
-        exit(BlobServicesApiImpl.BlobLeaseAcquire(BlobName, OptionalParameters, -1, ProposedLeaseId, LeaseId)); // Infinite duration, null Guid
+        exit(BlobServicesApiImpl.BlobAcquireLease(BlobName, OptionalParameters, -1, ProposedLeaseId, LeaseId)); // Infinite duration, null Guid
     end;
 
     /// <summary>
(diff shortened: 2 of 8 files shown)


Modules/System/Azure Blob Services API/src/ABSContainerClient.Codeunit.al
```diff
@@ -120,6 +120,201 @@ codeunit 9052 "ABS Container Client"
...
+    /// see: https://docs.microsoft.com/en-us/rest/api/storageservices/lease-container
+    /// </summary>    
+    /// <param name="DurationSeconds">Specifies the duration of the lease, in seconds, or negative one (-1) for a lease that never expires</param>
+    /// <returns>An operation reponse object</returns>
+    procedure LeaseAcquire(ContainerName: Text; DurationSeconds: Integer): Codeunit "ABS Operation Response"
+    var
+        OptionalParameters: Codeunit "ABS Optional Parameters";
+        ProposedLeaseId: Guid;
+    begin
+        exit(LeaseAcquire(ContainerName, DurationSeconds, ProposedLeaseId, OptionalParameters)); // Custom duration, null Guid
```
mazhelez, line 160: Why note call the implementation here directly? To save a couple of function calls.

Modules/System/Azure Blob Services API/src/ABSOperationResponse.Codeunit.al
```diff
@@ -61,6 +61,29 @@ codeunit 9050 "ABS Operation Response"
         Response := NewResponse;
     end;
 
+    procedure GetLeaseId(): Guid
```
mazhelez, line 64: I'm not particularly fond of this approach.

So every set of functions would add their return value here? Can we do better?
Maybe a wrapping codeunit.. or an interface?

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
```diff
@@ -226,6 +226,66 @@ codeunit 9047 "ABS Optional Parameters"
         SetRequestHeader('x-ms-requires-sync', ValueText);
     end;
 
+    /// <summary>
+    /// Sets the value for 'x-ms-lease-action' HttpHeader for a request.
+    /// </summary>
+    /// <param name="Value">Enum "ABS Lease Action" value specifying the HttpHeader value</param>
+    procedure LeaseAction("Value": Enum "ABS Lease Action")
```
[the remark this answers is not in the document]
mazhelez: Also, why is this one needed? I thought everything is handled in the new API methods: `LeaseRenew`, `LeaseBelease`, etc.

mazhelez (changes requested):
Great addition!

It just need a bit more polishing.
And I'm a bit concerned about `GetLeaseId` in the response object. I believe it can be achieved better. I'll think about it more as well.

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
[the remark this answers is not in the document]
SimonOfHH: Yeah, I guess you're right. I'll remove it.

Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
```diff
@@ -110,6 +113,51 @@ codeunit 9051 "ABS Client Impl."
...
         exit(OperationResponse);
     end;
+
+    procedure ContainerLeaseAcquire(ContainerName: Text; OptionalParameters: Codeunit "ABS Optional Parameters"; DurationSeconds: Integer; ProposedLeaseId: Guid): Codeunit "ABS Operation Response"
+    var
+        Operation: Enum "ABS Operation";
+    begin
+        OperationPayload.SetOperation(Operation::LeaseContainer);
+        OperationPayload.SetContainerName(ContainerName);
+        exit(LeaseAcquire(OptionalParameters, DurationSeconds, ProposedLeaseId, StrSubstNo(LeaseOperationNotSuccessfulErr, 'acquire', 'Container', OperationPayload.GetContainerName())));
```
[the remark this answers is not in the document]
SimonOfHH: Noted (and changed)

Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
```diff
@@ -638,5 +686,153 @@ codeunit 9051 "ABS Client Impl."
...
+    #region Private Lease-functions
+    local procedure LeaseAcquire(OptionalParameters: Codeunit "ABS Optional Parameters"; DurationSeconds: Integer; ProposedLeaseId: Guid; OperationNotSuccessfulErr: Text): Codeunit "ABS Operation Response"
+    var
+        OperationResponse: Codeunit "ABS Operation Response";
+        LeaseAction: Enum "ABS Lease Action";
+    begin
+        // Duration can be:
+        //   between 15 and 60 seconds
+        //   -1 (= infinite)
+        if ((DurationSeconds > 0) and ((DurationSeconds < 15) or (DurationSeconds > 60))) xor (not (DurationSeconds <> -1)) then
```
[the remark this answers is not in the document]
SimonOfHH: Well, it's either allowed to have a value between 15 and 60 seconds or -1 (for an infinite lease) (compare description for "x-ms-lease-duration" [here](https://docs.microsoft.com/en-us/rest/api/storageservices/lease-container#request)).
I remember that I had some troubles with that condition in my previous repo, so I just took it over. I'll re-test this and get back on this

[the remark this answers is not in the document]
SimonOfHH: You were right (of course) - I just checked it again and your proposal makes more sense

Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
```diff
@@ -89,6 +89,57 @@ codeunit 132919 "ABS Container Client Test"
...
+        Response := ABSContainerClient.LeaseAcquire(ContainerName, 60);
+        Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseAcquire failed');
+        LeaseId := Response.GetLeaseId();
+        Assert.IsFalse(IsNullGuid(LeaseId), 'Operation LeaseAcquire failed (no LeaseId returned)');
+
+        // [2nd] Renew Lease on Container
+        Response := ABSContainerClient.LeaseRenew(ContainerName, LeaseId);
+        Assert.IsTrue(Response.IsSuccessful(), 'Operation LeaseRenew failed');
+
+        // This part works when testing against a "real" Azure Storage and not Azurite
```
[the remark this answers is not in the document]
SimonOfHH: Can you help me/give me a hint how to do this the right way?

Modules/System/Azure Blob Services API/src/ABSContainerClient.Codeunit.al
SimonOfHH: Changed it

Modules/System/Azure Blob Services API/src/ABSOperationResponse.Codeunit.al
SimonOfHH: Well, in my opinion the other way around this would be to only make the procedure `GetHeaderValueFromResponseHeaders` public, because there needs to be a way to retrieve values from the response headers.
A wrapping-codeunit would be a bit overkill (at least from what I think), because `"ABS Operation Response"` already is kind of a wrapping-codeunit for the HttpResponse.

But of course I'm open to suggestions 🤗

SimonOfHH:
Hey @mazhelez I updated according to your comments, but also have some open questions. Looking forward to your feedback 👍

Modules/System/Azure Blob Services API/src/ABSOperationResponse.Codeunit.al
mazhelez: I have a couple of different ideas:

_**Idea 1**_:
Have the lease ID as a return parameter in the function where it makes sense to be returned:
E.g
```
procedure LeaseAcquire(<some params>, var LeaseId: Guid): Codeunit "ABS Operation Response"
```

_**Idea 2**_:
It's a bit more work but hear me out...
Not necessarily wrap, but more like expose the "ABS Operation Response" via different interfaces.
E.g.
```
// Generic response
interface "ABS Operation Response"
{
    procedure IsSuccessful(): Boolean;
    procedure GetError(): Text;
}
```
```
// Response for lease-related operations
interface "ABS Lease Operation Response"
{
    // "inherited" from the generic one (here AL is still limiting us)
    procedure IsSuccessful(): Boolean;
    procedure GetError(): Text;
    
    // Lease-specific results
    procedure GetLeaseId(): Guid;
}
```

Then `LeaseAcquire` will look like
```
procedure LeaseAcquire(<some params>): Interface "ABS Lease Operation Response"
```

But `ListBlobs` will look like
```
procedure ListBlobs(<some params>): Interface "ABS Operation Response"
```

Now this approach requires codeunit "ABS Operation Response" to become _internal_ and to implement the interfaces so it can be accessed through them.


Ok, enough input from me. Let me hear your opinion!

Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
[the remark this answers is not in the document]
mazhelez: I'd do literally
```
...
// The current test is supposed to run only against real Azure Storage Services
If (not real Azure Storage Services)
    exit;
...
```
I'll let you get creative on how to determine if it's "real Azure" or not.

mazhelez:
> Hey @mazhelez I updated according to your comments, but also have some open questions. Looking forward to your feedback 👍

Getting better and better. Let's finalize the ongoing discussion and be over with it.

Modules/System/Azure Blob Services API/src/ABSOperationResponse.Codeunit.al
SimonOfHH: Wouldn't be **_Idea 2_** kind of a breaking change (making the existing public `"ABS Operation Response"` internal)? Of course, I wouldn't mind changing it like this. Would be similar like we have it in C#, e.g. with HttpResponseHeaders or HttpRequestHeaders, which inherits from HttpHeaders.

So, depending on if this is considered a breaking change I'd either proceed with Idea 1 or Idea 2

Modules/System Tests/Azure Blob Services API/src/ABSContainerClientTest.Codeunit.al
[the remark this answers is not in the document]
SimonOfHH: Ah, ok. I thought you'd think of something like using preprocessor directives or similar. But ok, I'll think about it and add something 👍

Modules/System/Azure Blob Services API/src/ABSOperationResponse.Codeunit.al
mazhelez: Technically, yes. There's no way to directly mark it as internal.
The process is as usual — mark it as obsolete and, after X amount of time, mark it as internal.

I agree that the code would look ugly for a while.

SimonOfHH: But then I'd propose to first add Idea 1 and then create another PR later introducing the refactoring (to not stretch this one out too much).
Would you agree?

SimonOfHH: Changed it to Idea 1 first, to get this PR (hopefully 😉) done, as a first step.

SimonOfHH:
> > Hey @mazhelez I updated according to your comments, but also have some open questions. Looking forward to your feedback 👍
> 
> Getting better and better. Let's finalize the ongoing discussion and be over with it.

If you agree with the latest changes I'd think we are ready to go :)

Modules/System/Azure Blob Services API/src/API Enums/ABSLeaseAction.Enum.al
```diff
@@ -0,0 +1,54 @@
...
+/// </summary>
+enum 9049 "ABS Lease Action"
+{
+    Access = Internal;
+    Extensible = false;
+
+    /// <summary>
+    /// Requests a new lease.
+    /// </summary>
+    value(0; acquire)
```
stkillen, line 18: Minor thing, but naming convention for enum values would be `Acquire` here (same applies to below values).

Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
```diff
@@ -600,6 +600,223 @@ codeunit 9053 "ABS Blob Client"
...
     end;
 
+    /// <summary>
+    /// Requests a new lease. If the blob does not have an active lease, the Blob service creates a lease on the blob. The lease duration can be 15 to 60 seconds or can be infinite
+    /// see: https://docs.microsoft.com/en-us/rest/api/storageservices/lease-blob
+    /// </summary>
+    /// <param name="BlobName">The name of the blob.</param>
+    /// <param name="LeaseId">Guid containing the response value from x-ms-lease-id HttpHeader</param>
+    /// <returns>An operation reponse object</returns>
+    procedure LeaseAcquire(BlobName: Text; var LeaseId: Guid): Codeunit "ABS Operation Response"
```
IhorHandziuk, line 610: Why "<entity><verb>" and not the more usual "<verb><entity>" for function names (e. g. LeaseAcquire vs AcquireLease)?

Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
```diff
@@ -638,5 +694,157 @@ codeunit 9051 "ABS Client Impl."
...
+        OperationPayload.SetOptionalParameters(OptionalParameters);
+
+        OperationResponse := BlobAPIWebRequestHelper.PutOperation(OperationPayload, OperationNotSuccessfulErr);
+        LeaseId := FormatHelper.RemoveCurlyBracketsFromString(OperationResponse.GetHeaderValueFromResponseHeaders('x-ms-lease-id'));
+        exit(OperationResponse);
+    end;
+
+    local procedure TestParameterSpecified("Value": Variant; ParameterName: Text; HeaderIdentifer: Text)
+    begin
+        case true of
```
IhorHandziuk, line 843: Suggestion: `if "Value".IsGuid() then`

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
```diff
@@ -226,6 +226,66 @@ codeunit 9047 "ABS Optional Parameters"
...
+        if not (LeaseAction in [LeaseAction::acquire]) then
+            Error(HeaderCanOnlyBeSetOnConditionErr, 'x-ms-lease-duration', 'x-ms-lease-action', 'acquire');
+        SetRequestHeader('x-ms-lease-duration', Format("Value"));
+    end;
+
+    /// <summary>
+    /// Sets the value for 'x-ms-proposed-lease-id' HttpHeader for a request.
+    /// </summary>
+    /// <param name="Value">Guid value specifying the HttpHeader value.</param>
+    procedure ProposedLeaseId("Value": Guid)
```
IhorHandziuk, line 270: Suggestion: `SetProposedLeaseId`

mazhelez: This isn't how this codeunit rolls, the setters are anonymous 😄

mazhelez (approved):
Looking good!

See if you want to address the few comments and let's uptake this!

Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
SimonOfHH: The only real reason was that all Lease-operations are shown alphabetically in Intellisense (like when you type "Lease" you see all operations for this. But now that I think about it: it wouldn't matter, since Intellisense is not only looking at the start of the name.
I'll adjust this to have it more coherent 👍

Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
SimonOfHH: Changed

Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
SimonOfHH: Changed

Modules/System/Azure Blob Services API/src/API Enums/ABSLeaseAction.Enum.al
SimonOfHH: I changed that, but when I tested it I noticed a small issue. In `GetLeaseActionFromRequestHeader` in `"ABS Optional Parameters"` I now needed to change the "lookup" of the Enum (because `Enum::"ABS Lease Action".Names.IndexOf` returned nothing with the lowercase value. 
I added a small change to make the first char uppercase, but I'm not sure if that's best practice. Could you maybe validate this?

SimonOfHH:
Hi all,
I added changes for the latest suggestion (please excuse the typo and name-confusion in the last commit-message 😅). Due to the change of the enum-values to pascal-case I need to change another function a little bit (see my comment to @stkillen suggestion). If you're fine with this I'm also fine with it and the PR can be merged 😊 If not, please give me a short feedback.

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
```diff
@@ -226,6 +226,66 @@ codeunit 9047 "ABS Optional Parameters"
...
+    internal procedure LeaseAction("Value": Enum "ABS Lease Action")
+    begin
+        SetRequestHeader('x-ms-lease-action', Format("Value"));
+    end;
+
+    /// <summary>
+    /// Sets the value for 'x-ms-lease-break-period' HttpHeader for a request.
+    /// </summary>
+    /// <param name="Value">Integer value specifying the HttpHeader value.</param>
+    procedure LeaseBreakPeriod("Value": Integer)
```
stkillen, line 242: Since this optional parameter is conditional on the lease action being `LeaseAction::Break`, I believe it's cleaner  to remove this procedure from optional parameters codeunit and take it as an argument in the API instead. This will also make it more transparent to the consumer of the API.

I know this will add a couple of new API overloads (with the different permutations of arguments) but I still think it's better than an optional parameter that is tied up to a specific lease action, and errors out in all other cases.

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
```diff
@@ -226,6 +226,66 @@ codeunit 9047 "ABS Optional Parameters"
...
+        if not (LeaseAction in [LeaseAction::Break]) then
+            Error(HeaderCanOnlyBeSetOnConditionErr, 'x-ms-lease-break-period', 'x-ms-lease-action', 'break');
+        SetRequestHeader('x-ms-lease-break-period', Format("Value"));
+    end;
+
+    /// <summary>
+    /// Sets the value for 'x-ms-lease-duration' HttpHeader for a request.
+    /// </summary>
+    /// <param name="Value">Integer value specifying the HttpHeader value.</param>
+    procedure LeaseDuration("Value": Integer)
```
stkillen, line 256: Same as above. Could be moved to API instead of as a conditional, optional parameter

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
```diff
@@ -226,6 +226,66 @@ codeunit 9047 "ABS Optional Parameters"
...
+
+    /// <summary>
+    /// Sets the value for 'x-ms-proposed-lease-id' HttpHeader for a request.
+    /// </summary>
+    /// <param name="Value">Guid value specifying the HttpHeader value.</param>
+    procedure ProposedLeaseId("Value": Guid)
+    var
+        LeaseAction: Enum "ABS Lease Action";
+    begin
+        LeaseAction := GetLeaseActionFromRequestHeader();
```
stkillen, line 274: Same as above. Could be moved to API instead of as a conditional, optional parameter

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
```diff
@@ -226,6 +226,66 @@ codeunit 9047 "ABS Optional Parameters"
...
+        if not (LeaseAction in [LeaseAction::Acquire, LeaseAction::Change]) then
+            Error(HeaderCanOnlyBeSetOnTwoConditionsErr, 'x-ms-proposed-lease-id', 'x-ms-lease-action', 'acquire', 'change');
+        SetRequestHeader('x-ms-proposed-lease-id', "Value");
+    end;
+
+    internal procedure GetLeaseActionFromRequestHeader(): Enum "ABS Lease Action"
```
stkillen, line 280: This function could be removed if the above functions were removed and added to the API instead.

stkillen (commented):
Hi @SimonOfHH, looking into the `GetLeaseActionFromRequestHeader` function resulted in a few more suggestions. Let me know what you think.

Modules/System/Azure Blob Services API/src/API Enums/ABSLeaseAction.Enum.al
stkillen: Hi! Please take a look at my new comments, which if implemented would remove GetLeaseActionFromRequestHeader function.

If we were to keep the function, I think it would be better to introduce a helper function that maps string -> lease action enum, instead of depending on the enum names directly.

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
SimonOfHH: Done

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
SimonOfHH: Done

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
SimonOfHH: Done

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
SimonOfHH: Done

SimonOfHH:
Hi @stkillen I modified the code as to your suggestions. I agree that it makes it a bit more reliable without this helper for the enum. 
The setter-function still exist, but reduced to one line of code each and marked as `internal` - hope that's the way to go.

mazhelez:
> Hi @stkillen I modified the code as to your suggestions. I agree that it makes it a bit more reliable without this helper for the enum. The setter-function still exist, but reduced to one line of code each and marked as `internal` - hope that's the way to go.

Great to hear, Simon! Way to start the new year... for you — to end the old one :D 
I'll now start processing the PR internally, but do not expect any blockers!

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
```diff
@@ -331,4 +367,7 @@ codeunit 9047 "ABS Optional Parameters"
         ABSFormatHelper: Codeunit "ABS Format Helper";
         RequestHeaders: Dictionary of [Text, Text];
         Parameters: Dictionary of [Text, Text];
+        NeedToSpecifyHeaderErr: Label 'You need to specify the "%1"-header to set this header.', Comment = '%1 = HttpHeader identifier';
```
mazhelez, line 370: These strings appear to not be used anymore.

SimonOfHH: Yes, I guess you're right. This was used in one the removed functions. Could you remove it? Or should I push another commit?

mazhelez: I'll have them removed :)

mazhelez:
🥳 we managed to process the PR and it will be available in the next minor release (19.4, if I'm not mistaken).

Issue #15162, opened 2021-11-10
There is no overloaded function that allows to save an image in a specified format


nicolassaleron:
Hello,

The module Image does not allow to save an image in a new format .

https://github.com/microsoft/ALAppExtensions/blob/60b3c46df078f7f2414023c6ce75c900fc9a48c7/Modules/System/Image/src/Image.Codeunit.al#L104

What is the replacement for DotNet_Image.Save(OutStream, Codeunit "DotNet_ImageFormat") ?

AlexanderYakunin:
@JesperSchulz

JesperSchulz:
It looks like none has been implemented. Probably no other developer had the need for this capability! Luckily, we've open sourced the module, which allows you to add exactly the overloads that you need. Could you create a PR? We'll process it swiftly!

nicolassaleron:
Contributing to the System App: that's new for me  :)

I see many ways to achieve the same result. Have you any preference or is it up-to-me?

* I could propose an overload with ```Save(OutStream: OutStream, Format: Enum "Image Format")``` where the current image is just re-encoded to the specified format into the specified stream
* It could be a ```SetFormat(NewFormat: Enum "Image Format")``` where the format of the internal stream of Image is changed. Then the existing ```Save(OutStream: OutStream)``` would remain the same.

JesperSchulz:
Happy to hear that you're considering to create a PR :-D

Generally, that would be up to you! As you are the developer, you are free to choose the option which best suits your needs. Both options seem viable to me!

Looking forward to receiving your pull request!

If you get stuck or need help, please don't hesitate to reach out! We're here to help :-)

JesperSchulz:
Progress in PR instead 🥳

PR #15225, opened 2021-11-17
New SetFormat in module Image

nicolassaleron:
This PR is related to issue #15162. It adds a new procedure called SetFormat in the module Image.
This procedure allows the user to change the format of the image typically to re-encode a BMP to PNG or PNG to JPEG.

### DIFF
--- a/Modules/System/Image/README.md
+++ b/Modules/System/Image/README.md
@@ -56,6 +56,31 @@ begin
 
 end;
 ```
+
+##### Save an image in a new format
+```
+procedure Example()
+var
+    TempBlob:Codeunit "Temp Blob";
+    Image: Codeunit Image;
+    InStream: InStream;
+    OutStream: OutStream;
+    FileName: Text;
+begin
+    UploadIntoStream('', '', '', FileName, InStream);
+
+    Image.FromStream(InStream);
+    Image.SetFormat(Enum::"Image Format"::Png);
+
+    TempBlob.CreateOutStream(OutStream);
+    Image.Save(OutStream);	
+
+    TempBlob.CreateInStream(InStream);
+    FileName := FileName + '.png';
+    DownloadFromStream(InStream, '', '', '', FileName);
+end;
+```
+
 # Public Objects
 ## Image (Codeunit 3971)
 
@@ -121,6 +146,15 @@ procedure GetFormat(): Enum "Image Format"
 *[Enum "Image Format"]()*
 
 The enum value.
+### SetFormat (Method) <a name="SetFormat"></a> 
+
+ Sets the image format from an Enum "Image Format".
+ 
+
+#### Syntax
+```
+procedure SetFormat(ImageFormat: Enum "Image Format")
+```
 ### FromBase64 (Method) <a name="FromBase64"></a> 
 
  Creates an Image from base64 encoding.
--- a/Modules/System/Image/README.md
+++ b/Modules/System/Image/README.md
@@ -61,7 +61,7 @@ end;
 ```
 procedure Example()
 var
-    TempBlob:Codeunit "Temp Blob";
+    TempBlob: Codeunit "Temp Blob";
     Image: Codeunit Image;
     InStream: InStream;
     OutStream: OutStream;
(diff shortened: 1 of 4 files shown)


Modules/System/Image/README.md
```diff
@@ -56,6 +56,31 @@ begin
 
 end;
 ```
+
+##### Save an image in a new format
+```
+procedure Example()
+var
+    TempBlob:Codeunit "Temp Blob";
```
Groenbech96, line 64: Space

Groenbech96 (changes requested):
I would like you to add some tests for the setting the format. 

See the tests already written and add them there.
https://github.com/microsoft/ALAppExtensions/blob/main/Modules/System%20Tests/Image/src/ImageTests.Codeunit.al

Modules/System/Image/README.md
nicolassaleron: The space has been added.

nicolassaleron:
@Groenbech96 I have added the SetFormatTest procedure.
I assume ```ImageAsBase64Txt``` is a Png image, it is seems this is also the assumption done in the other tests.

Groenbech96 (commented):
This looks super! 
I just got thinking, what happens if you call SetFormat multiple times? 

Say you convert from jpeg to png and back, as you modify the image itself each time, and jpeg is a lossy format, meaning some data is lost when converting to it, would you end up with a distorted image, even though you only save it at the end?

nicolassaleron:
> This looks super! I just got thinking, what happens if you call SetFormat multiple times?
> 
> Say you convert from jpeg to png and back, as you modify the image itself each time, and jpeg is a lossy format, meaning some data is lost when converting to it, would you end up with a distorted image, even though you only save it at the end?

You got the way it works. If you start with a BMP file and you do SetFormat(Jpeg) -> SetFormat(Gif) -> SetFormat(Bmp), the final Bmp is not the same as the original one.

That's why I asked on the issue if a function Save(OutStream, Enum::ImageFormat) would be preferable.

On the other hand if you start with an image with a size 32x32 and you call Resize(8, 8) -> Resize(32, 32) you do not finish with the same image either.
That's why I considered SetFormat better than an overload of the Save procedure.

Let me know if the implementation should be changed.

Groenbech96:
@nicolassaleron Would you ever change the format without saving the image? I don't think I would, at least I cant think of any case. In most image applications, you save and select what format you want to save as, so maybe it is a better solution. 
Let me know what you think?

Whereas resize would be directly on the image in an application, like edits or whatever.

nicolassaleron:
> @nicolassaleron Would you ever change the format without saving the image? I don't think I would, at least I cant think of any case. In most image applications, you save and select what format you want to save as, so maybe it is a better solution. Let me know what you think?
> 
> Whereas resize would be directly on the image in an application, like edits or whatever.

So your preference goes to a Save method instead of a SetFormat method? If so I can make the change next week.

Groenbech96:
Yes, lets go with that.
I will take a look next week!

JesperSchulz:
It's still a matter of taste. Neither of the two solutions is wrong. As long as we document that calling SetFormat instantly reformats, which can result in quality loss, I don't see any problem here. If someone decides to convert, only to convert back, it's their own fault. @nicolassaleron, it's your call if you pick one or the other.

bjarkihall:
You could overload the Save procedure with all of the operations (parameters for resizing, cropping, rotating, etc.) with a similar reason, but doesn't an explicit SetFormat method allow us to have better control over the order of operations, like changing the format before or in the middle of editing? It's also better in cases where you just have to set the format temporarily without saving/downloading the changed image yet, right?
I agree with Jesper that it's probably enough to just mention that this method can be lossy.

nicolassaleron:
I have a little preference over the SetFormat method instead of Save.
I have added a remark on the method explaining it can be lossy.

Regarding this discussion, may I ask the reason why this object stores the image as a TempBlob instead of retaining a DotNet Image? (because of the memory consumption?)
If there is no particular reason, would not it be better to encode the image only when the user call the Save method?

Groenbech96:
@nicolassaleron To answer your question: When the module was originally designed, we came across some issues when cropping or cutting the image using the Dotnet Image.  (It is a while a go, so my memory is a little blurry here). In our internal commit history, we went from a Dotnet Image to a temp blob. 

As @JesperSchulz said, it is up to you with how you think the best implementation. 👍

nicolassaleron:
@Groenbech96 Thanks for the answer :)

I would like keeping this implementation if you do not mind.

JesperSchulz (approved):

JesperSchulz:

PR #15284, opened 2021-11-22
Fixed code sample: RoleCenterId is by reference

nicolassaleron:
There is a mismatch between the code example and the API definition few lines below.


JesperSchulz:
I think this PR takes the price for the smallest one yet :-D But sometimes a single word can make all the difference ;-)

stkillen (approved):

JesperSchulz:

JesperSchulz:

PR #15365, opened 2021-11-29
Remove "Internal" from Codeunit 9060

hemisphera:
I propose to remove the "Internal" access modifier from Codeunit 9060. This Codeunit does not seem to handle any sensitive data and/or security related stuff, so I do not see a real need for it to be internal.

On the other hand, it provides two important format functions that allows us to format dates according to ISO8601 and RFC1123. This are just a few examples of formats required when interfacing with external web services. Without the .NET DateTime we are just left to reinvent the wheel on this, messing around with strings - which is cumbersome and error-prone (especially with non-english languages). The wrapper Codeunit DotNet_DateTime does not really help here.

Exposing this codeunit would give us easy access to these formats.

### DIFF
--- a/Modules/System/Azure Storage Services Authorization/src/AuthFormatHelper.Codeunit.al
+++ b/Modules/System/Azure Storage Services Authorization/src/AuthFormatHelper.Codeunit.al
@@ -5,6 +5,7 @@
 
 codeunit 9060 "Auth. Format Helper"
 {
+    Access = Internal;
 
     procedure NewLine(): Text
     var
@@ -41,5 +42,4 @@ codeunit 9060 "Auth. Format Helper"
     begin
         exit(CryptographyManagement.GenerateBase64KeyedHashAsBase64String(StringToSign, AccessKey, HashAlgorithmType::HMACSHA256));
     end;
-
 }
--- a/Modules/System/Azure Storage Services Authorization/src/AuthFormatHelper.Codeunit.al
+++ b/Modules/System/Azure Storage Services Authorization/src/AuthFormatHelper.Codeunit.al
@@ -5,7 +5,6 @@
 
 codeunit 9060 "Auth. Format Helper"
 {
-    Access = Internal;
 
     procedure NewLine(): Text
     var
@@ -42,4 +41,5 @@ codeunit 9060 "Auth. Format Helper"
     begin
         exit(CryptographyManagement.GenerateBase64KeyedHashAsBase64String(StringToSign, AccessKey, HashAlgorithmType::HMACSHA256));
     end;
+
 }
--- a/Modules/System/Azure Storage Services Authorization/src/StorageServiceAuthorization.Codeunit.al
+++ b/Modules/System/Azure Storage Services Authorization/src/StorageServiceAuthorization.Codeunit.al
@@ -100,7 +100,7 @@ codeunit 9062 "Storage Service Authorization"
     var
         AuthHelper: Codeunit "Auth. Format Helper";
     begin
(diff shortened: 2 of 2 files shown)


JesperSchulz (changes requested):
Making codeunits internal is not about "handling sensitive data or security related stuff". It's about decoupling and encapsulation. You can read much more about that here: [https://alguidelines.dev/bcpatterns/facade-pattern/](https://alguidelines.dev/bcpatterns/facade-pattern/). If you would like to expose internal functions from that module, please do so through the facade of the module - in this case the codeunit StorageServiceAuthorization.Codeunit.al.

JesperSchulz:
One could also consider to move these useful functions out into a module of their own - if we expect more to be added over time. We could also give them their own facade, if the existing one would get polluted by adding these functions.. But we should not just make the internal codeunit public. It contains implementation details, we would like to hide.

hemisphera:
I had specifically chosen the "Auth. Format Helper" Codeunit because it did not seem like an implementation Codeunit, more like a helper / utility Codeunit that might be useful being public.

With that said, the issue lies somewhat deeper however. We are missing a lot of "basic" system functionality that .NET provided and we're now left without. I therefore I agree that this should go into a dedicated module, just as you suggested. But forging a new module from scratch is not something that's done quickly and should not be the subject of a PR, done by someone not on the team, I suppose?

For the time being I will revert the change and exposed them in the implementation Codeunit. Although I am not too happy with it. These kind of formatting and "low-level" things should really go into a common "formatting" Module and not into specific modules. They should be easily discoverable, it's nothing we as AL developers should need to care about.

JesperSchulz:
_"...forging a new module from scratch is not something that's done quickly and should not be the subject of a PR, done by someone not on the team, I suppose?"_ is not correct. Everyone is welcome to create new modules! We'd be happy to receive one or more modules with all the formatting .NET wrappers you'd wish for. You could also start slow and just add the ones you need, and then other members of the community can add the ones they need. That's the intention of the System Application and our open source initiative - collaboration on horizontal functionality. And once you've got the hang of it, it doesn't require that large of an investment to create a new module. It is after all just an app.
We've written a little how-to, in case you'd like to give it a try: [aka.ms/NewALModule](aka.ms/NewALModule). If you need more help, we're ready to assist!
Word of advise: Start small! Create a PR and get a code review on your new module and merge it in. Then expand it! Better to have many smaller iterations, than working on a new module for weeks and then learn, that the entire design is wrong. Use the Math module or RegEx as inspiration. Those are pretty clean wrappers.

JesperSchulz:
> > We've written a little how-to, in case you'd like to give it a try: [aka.ms/NewALModule](aka.ms/NewALModule).
> 
> the link seems wrong:
> 
> image

That's odd! The link is correct! Try to copy/paste the link into the address line, rather than clicking on it - then it works :-)

dnpb:
Quote-replying shows that your URL is relative, so it seems GitHub is treating it as such - I guess you need to put the full URL in.

image

JesperSchulz:
> Quote-replying shows that your URL is relative, so it seems GitHub is treating it as such - I guess you need to put the full URL within `(these)` so that it treats it absolutely.
> 
> image

Adding https:// to the link did the trick, yes :-) Good to know. Thanks!

JesperSchulz:
I assume there won't be a new module coming out of this discussion? 😉 I am hence closing this PR - but if you at any point want to add a module which adds the capabilities you are looking for, please don't hesitate to either reopen and update this PR or to create a new one!

Issue #15409, opened 2021-11-30
9064 "Stor. Serv. Auth. Shared Key" generates invalid keys for IT databases


hemisphera:
The codeunit 9064 "Stor. Serv. Auth. Shared Key" is responsible for generating the shared key used for storage service authentication. Specifically: the function `GetSharedKeySignature`. This function uses the label `SignaturePlaceHolderLbl` as a placeholder string where the two parts (account name & key) are then filled in via `StrSubstNo`. This label however does not have the `Locked` property set and is therefore translated in the various localizations.

In the IT version platform 19.0.29884.30666 + application 19.0.29894.30693 this label is translated wrong (a space is added before the colon). 

Therefore using an IT database in IT language causes the access key header to be wrong and essentially makes all storage operation calls fail.

![image](https://user-images.githubusercontent.com/8897936/144090335-ee693d14-0d18-4074-b378-b9cf2716abce.png)

This label should not be translated and therefore have the `Locked` property set.

darjoo:
I'll close this bug as we will process the PR :)

PR #15410, opened 2021-11-30
#15409: Lock 'SignaturePlaceHolderLbl' to avoid translation

hemisphera:
#15409


mazhelez (approved):
Nice catch!

Thank you for contributing the fix :)

darjoo (approved):

JesperSchulz:

JesperSchulz:

hemisphera:
Hi Jesper,

your last response was 14 days ago - at that time the "next update for
release 19" would have been BC19.02 which came out and does not yet contain
the fix. We have a customer that is waiting on this. Will this come with
BC19.03? Do you have any ETA?

Thanks

JesperSchulz:
The change went in on the 2nd of December and will hence ship with the January update, which is 2021 Wave II Update 3. As a rule of thumb, we branch for the upcoming monthly minor in the last week before the upcoming month and the update releases early in the month. Occasionally, internal or external issues prevent us from releasing at the planned time, but the release still remains locked for additions other than those which will resolve the release issues.

PR #15762, opened 2022-01-05
Add the ability to set the From Name and From Address on Email via the SMTP Connector

complord:
These updates allow the use of the "From Name" and "From Address" fields on the "Email Item" record when sending email using the "Email - SMTP Connector" extension.

The updated name and address, if used, are reflected in the "Sent Email" records.

### DIFF
--- a/Apps/W1/Email - SMTP Connector/app/src/SMTPConnectorImpl.Codeunit.al
+++ b/Apps/W1/Email - SMTP Connector/app/src/SMTPConnectorImpl.Codeunit.al
@@ -183,7 +183,17 @@ codeunit 4513 "SMTP Connector Impl." implements "Email Connector"
         AttachmentInStream: InStream;
     begin
         // From name/email address
-        SMTPMessage.AddFrom(SMTPAccount."Sender Name", SMTPAccount."Email Address");
+        if EmailMessage.GetFromAddress() <> '' then
+            FromAddress := EmailMessage.GetFromAddress()
+        else
+            FromAddress := SMTPAccount."Email Address";
+
+        if EmailMessage.GetFromName() <> '' then
+            FromName := EmailMessage.GetFromName()
+        else
+            FromName := SMTPAccount."Sender Name";
+
+        SMTPMessage.AddFrom(FromName, FromAddress);
 
         // To, Cc and Bcc Recipients
         EmailMessage.GetRecipients("Email Recipient Type"::"To", Recipients);
--- a/Modules/System/Email/src/Email/EmailImpl.Codeunit.al
+++ b/Modules/System/Email/src/Email/EmailImpl.Codeunit.al
@@ -228,7 +228,11 @@ codeunit 8900 "Email Impl"
         EmailOutbox."Account Id" := AccountId;
         EmailOutbox.Description := CopyStr(EmailMessageImpl.GetSubject(), 1, MaxStrLen(EmailOutbox.Description));
         EmailOutbox."User Security Id" := UserSecurityId();
-        EmailOutbox."Send From" := CopyStr(SentFrom, 1, MaxStrLen(EmailOutbox."Send From"));
+        if EmailMessage.GetFromName() <> '' then
+            EmailOutbox."Send From" := CopyStr(EmailMessage.GetFromName(), 1, MaxStrLen(EmailOutbox."Send From"))
+        else
+            EmailOutbox."Send From" := CopyStr(SentFrom, 1, MaxStrLen(EmailOutbox."Send From"));
+        EmailOutbox."Send From Address" := EmailMessage.GetFromAddress();
         EmailOutbox.Status := Status;
         if Status = Enum::"Email Status"::Queued then begin
             EmailOutbox."Date Queued" := CurrentDateTime();
--- a/Modules/System/Email/src/Message/EmailMessage.Codeunit.al
+++ b/Modules/System/Email/src/Message/EmailMessage.Codeunit.al
@@ -60,6 +60,16 @@ codeunit 8904 "Email Message"
     end;
 
     /// <summary>
+    /// Adds the mailbox that this email is being sent from.
+    /// </summary>
+    /// <param name="FromName">The name of the email sender</param>
+    /// <param name="FromAddress">The address of the email sender</param>
+    procedure SetFrom(FromName: Text[250]; FromAddress: Text[250])
+    begin
+        EmailMessageImpl.SetFrom(FromName, FromAddress);
+    end;
+
+    /// <summary>
     /// Gets the email message with the given ID.
     /// </summary>
     /// <param name="MessageId">The ID of the email message to get.</param>
@@ -76,6 +86,24 @@ codeunit 8904 "Email Message"
     procedure GetBody(): Text
     begin
         exit(EmailMessageImpl.GetBody());
+    end;
+
+    /// <summary>
+    /// Gets the sender address of the email message.
+    /// </summary>
+    /// <returns>The sender address of the email.</returns>
+    procedure GetFromAddress(): Text[250]
+    begin
+        exit(EmailMessageImpl.GetFromAddress());
+    end;
+
+    /// <summary>
+    /// Gets the sender name of the email message.
+    /// </summary>
+    /// <returns>The sender name of the email.</returns>
+    procedure GetFromName(): Text[250]
+    begin
+        exit(EmailMessageImpl.GetFromName());
     end;
 
     /// <summary>
--- a/Modules/System/Email/src/Message/EmailMessage.Codeunit.al
+++ b/Modules/System/Email/src/Message/EmailMessage.Codeunit.al
@@ -63,10 +63,18 @@ codeunit 8904 "Email Message"
     /// Adds the mailbox that this email is being sent from.
     /// </summary>
     /// <param name="FromName">The name of the email sender</param>
-    /// <param name="FromAddress">The address of the email sender</param>
-    procedure SetFrom(FromName: Text[250]; FromAddress: Text[250])
+    procedure SetFromName(FromName: Text[250])
     begin
-        EmailMessageImpl.SetFrom(FromName, FromAddress);
+        EmailMessageImpl.SetFromName(FromName);
+    end;
+
+    /// <summary>
+    /// Adds the mailbox that this email is being sent from.
+    /// </summary>
+    /// <param name="FromAddress">The address of the email sender</param>
+    procedure SetFromAddress(FromAddress: Text[250])
+    begin
+        EmailMessageImpl.SetFromAddress(FromAddress);
     end;
 
     /// <summary>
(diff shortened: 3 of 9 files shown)


complord:
I should add that this also uses the OnSendViaEmailModuleOnAfterCreateMessage event in codeunit 23049467 "CNP DP Mail Management";

    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Mail Management", 'OnSendViaEmailModuleOnAfterCreateMessage', '', false, false)]
    local procedure OnSendViaEmailModuleOnAfterCreateMessage(var Message: Codeunit "Email Message"; var TempEmailItem: Record "Email Item");
    begin
        Message.SetFrom(TempEmailItem."From Name", TempEmailItem."From Address");
    end;

JesperSchulz:
Apologies for the slight delay in processing this PR. We hope to get to do a review soon!

darjoo:
@complord Apologies for further delays, we're doing some refactoring to separate the dotnet stuff into an API app of its own from the connector. We're almost done and I might be able to help directly put in your changes as well. I  will get onto it asap!

mjmatthiesen:
Great PR. Covers exactly the use case we had with the previous smtp codeunit.

Would it be possible for you to add an override for Create that includes the from details as well? Not vital for me, but would be great imo.

complord:
I was trying to minimize the changes, hence the use of the event to set the From information.  Also, MS is making some changes to the base system, so some or all of this might become moot.

darjoo:
> I was trying to minimize the changes, hence the use of the event to set the From information. Also, MS is making some changes to the base system, so some or all of this might become moot.

Hi @complord, things have been moved around, but I will attempt to sort them out. My team and I are busy with Launch Event and Directions preparation this week, however this and the other Email PR are at the top of my agenda for next week.

complord:
Hi @darjoo 

Should I refactor this for version 20?

Apps/W1/Email - SMTP Connector/app/src/SMTPMessage.Codeunit.al
```diff
@@ -101,10 +101,22 @@ codeunit 4514 "SMTP Message"
...
         AttachmentInStream: InStream;
         Body: Text;
+        FromName: Text[250];
+        FromAddress: Text[250];
         Recipients: List of [Text];
     begin
         // Add From
-        AddFrom(SMTPAccount."Sender Name", SMTPAccount."Email Address");
+        //AddFrom(SMTPAccount."Sender Name", SMTPAccount."Email Address");
+        if Message.GetFromAddress() <> '' then
```
darjoo, line 110: Looks good, but this section has now moved into the `SMTPConnectorImpl.Codeunit.al` in the `BuildSMTPMessage` function. 

Should be able to copy this part, paste it in there and update the variable names.

darjoo (changes requested):
@complord This looks good, one change and to just add some tests, at least one in the EmailE2ETests to ensure that it saves and takes the expected name/email from the message into the outbox and sent emails.

Apps/W1/Email - SMTP Connector/app/src/SMTPMessage.Codeunit.al
complord: Done.

complord:
Moved the code as requested, and added a test function.

Modules/System/Email/src/Message/EmailMessage.Codeunit.al
```diff
@@ -59,6 +59,16 @@ codeunit 8904 "Email Message"
         EmailMessageImpl.Create(ToRecipients, Subject, Body, HtmlFormatted, CCRecipients, BCCRecipients);
     end;
 
+    /// <summary>
+    /// Adds the mailbox that this email is being sent from.
+    /// </summary>
+    /// <param name="FromName">The name of the email sender</param>
+    /// <param name="FromAddress">The address of the email sender</param>
+    procedure SetFrom(FromName: Text[250]; FromAddress: Text[250])
```
IhorHandziuk, line 67: Might make sense to have individual setters for each of the 2 fields

Modules/System/Email/src/Email/EmailImpl.Codeunit.al
```diff
@@ -228,7 +228,11 @@ codeunit 8900 "Email Impl"
         EmailOutbox."Account Id" := AccountId;
         EmailOutbox.Description := CopyStr(EmailMessageImpl.GetSubject(), 1, MaxStrLen(EmailOutbox.Description));
         EmailOutbox."User Security Id" := UserSecurityId();
-        EmailOutbox."Send From" := CopyStr(SentFrom, 1, MaxStrLen(EmailOutbox."Send From"));
+        if EmailMessage.GetFromName() <> '' then
+            EmailOutbox."Send From" := CopyStr(EmailMessage.GetFromName(), 1, MaxStrLen(EmailOutbox."Send From"))
```
IhorHandziuk, line 232: Seems weird to have 2 potential places with "Send From"

Apps/W1/Email - SMTP Connector/app/src/SMTPConnectorImpl.Codeunit.al
```diff
@@ -183,7 +183,17 @@ codeunit 4513 "SMTP Connector Impl." implements "Email Connector"
...
+            FromAddress := EmailMessage.GetFromAddress()
+        else
+            FromAddress := SMTPAccount."Email Address";
+
+        if EmailMessage.GetFromName() <> '' then
+            FromName := EmailMessage.GetFromName()
+        else
+            FromName := SMTPAccount."Sender Name";
+
+        SMTPMessage.AddFrom(FromName, FromAddress);
```
IhorHandziuk, line 196: Do we also want to update other 1st part connectors to respect the new fields?

complord: Add a modification to codeunit 4509 "Email - Outlook API Helper"

complord:
Just following up to confirm if I have addressed the issues and marked them properly.

IhorHandziuk:
Hi @complord , sorry for the very long delay. Thanks for addressing all the comments from before, but now I actually have some doubts about the design of this functionality (i. e. the message should not have the "responsibility" of knowing who to send the message to).

It seems that the entire purpose of this PR is to support the legacy Email Item fields with the new email system. I think it's a good idea, but I think it should be implemented through the accounts system. One way of doing it could be creating a new email connector (e. g. "Dynamic Sender" connector) which would provide the ability to add one account (similarly to the "Current User" connector). This connector then can subscribe to the events in email management / email item for populating the fields.
This way the responsibilities of components in the email system are clearly defined and are non overlapping (of course, there are also other ways of implementing this functionality 😊)

IhorHandziuk:
To answer your question:
Not necessarily - there are events that allow you to substitute what account is used for sending the current email before it reaches the email module (e. g. OnBeforeQualifyFromAddress in codeunit "Mail Management").

In general:
Can you please tell us more about your scenario (i. e. why would you like to use these fields from the Email Item)? They are not populated anywhere in the production code in the latest application and (after a discussing it for a while) we actually now think of deprecating them.

The new email system in BC is designed to have a clear separation between the email content, email accounts and email scenarios. So, the suggested way of dynamically changing the from address / name depending on some conditions (the functionality that in the past was handled by Email Item) is extending _email scenarios_.

complord:
@mjmatthiesen is correct.  We have an app, in which an important part is being able to set the from name/from address.

Changes to the From Name/From Address in the OnBeforeQualifyFromAddress event have no effect.

IhorHandziuk:
> We have an app, in which an important part is being able to set the from name/from address.
> Changes to the From Name/From Address in the OnBeforeQualifyFromAddress event have no effect.

I see, thanks for the explanation. I don't think we want to revive the mentioned functionality in exactly the same way it used to work before (for the reasons mentioned above).

For your case I have the following suggestion of achieving the same results:
1) create a separate SMTP Account for every unique pair of "From Name" / "From Address" you need to use;
2) either 
- subscribe to the `OnQualifyFromAddressOnAfterGetEmailAccount` event and update the `TempEmailModuleAccount` variable to correspond to the desirable account **_or_**
- define email scenarios for which you need to use different email accounts and assign the desirable email accounts to those scenarios

This way there will be more transparency for the users as to what account has been used for sending what email. All the while keeping the responsibilities of the different components of the email system separated 🙂.

JesperSchulz:
@complord, what do you think of @IhorHandziuk's suggestion? Will that resolve your issue? Do you still need this PR? We’re doing a bit of housekeeping and are looking to close PRs that haven’t had any attention for a while.

complord:
@JesperSchulz, technically it will work, but it puts the onus on the user(s) to maintain, whereas previously it only required a single SMTP mail account to be implemented.  Perhaps there are other options, e.g. a code change that is more acceptable?

IhorHandziuk:
@complord, what exactly do you mean by the additional onus on the users? Email accounts can be registered / assigned to scenarios / maintained programmatically. With this approach the users will have just the same functionality as before (i. e. the email accounts can be selected dynamically according to scenarios), but with better clarity as to what gets sent from what account. 🙂

complord:
When I last checked, the e-mail accounts have to be created manually for each required user.  The SMTP objects were all internal.

mjmatthiesen:
> @complord, what exactly do you mean by the additional onus on the users? Email accounts can be registered / assigned to scenarios / maintained programmatically. With this approach the users will have just the same functionality as before (i. e. the email accounts can be selected dynamically according to scenarios), but with better clarity as to what gets sent from what account. 🙂

For some context, what our solution used to be able to do was set the from email and the from name. This allowed independence of the two. Which meant you could send an email from one generic account, but sign it with the user name (we did this programmatically). Now we have an email scenario and people can use which email account to use, but unless they configure multiple with the same email but different names, they cannot reproduce the above scenario. This isn't a massive deal and our users have come to terms with it.

IhorHandziuk:
> When I last checked (although it's been a while), the e-mail accounts have to be created manually for each required user. The SMTP objects were all internal.

Sorry for a very delayed reply - this request has unfortunately fallen off our radar. Indeed, currently most of the functionality in the SMTP connector app is internal, but we would be happy to accept a contribution that unlocks the possibility to create SMTP account records programmatically. 🙂 Having a new public codeunit that exposes this functionality would probably be the the best.

JesperSchulz:
We're doing a bit of delayed "spring cleaning" on GitHub...

Unfortunately we cannot accept this PR in its current shape. As the suggested alternative implementation is very different, I am closing this PR and am asking you to create a new one (linked to this one), if you are interested in providing an implementation for the suggestion. If you choose to do so, we promise to turn that one around real quick. Given the long discussions in this PR, we're now rather sure about that that would be the right design.

Sorry about the back and forth on this one, but we're trying to keep our "new" System Application as clean as possible - which sometimes results in several iterations of trying to reach a certain goal.

PR #15877, opened 2022-01-17
Add 'SetDefaultSeed'

hemisphera:
Added 'SetDefaultSeed' that allows to initialize the instance with the same seed that a call to the system function `Randomize` would. This allows to generate non-predictable random numbers - without explicitly specifying a seed.

### DIFF
--- a/Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
+++ b/Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
@@ -12,6 +12,7 @@ codeunit 130500 "Any"
 {
     var
         Seed: Integer;
+        SeedSet: Boolean;
 
     /// <summary>
     /// Provides an integer between 1 and the given value
@@ -220,7 +221,18 @@ codeunit 130500 "Any"
     procedure SetSeed(NewSeed: Integer)
     begin
         Seed := NewSeed;
+        SeedSet := true;
         Randomize(Seed);
+    end;
+
+    /// <summary>.
+    /// Sets the default Seed for Pseudo-random number generation (no. of milliseconds since midnight of today).
+    /// Setting this value will change the numbers returned.
+    /// </summary>
+    procedure SetDefaultSeed()
+    begin
+        SeedSet := true;
+        Randomize();
     end;
 
     /// <summary>.
@@ -231,7 +243,7 @@ codeunit 130500 "Any"
     /// <returns>Pseudo-random integer value</returns>
     local procedure GetNextValue(MaxValue: Integer): Integer
     begin
-        if Seed = 0 then
+        if (not SeedSet) then
             SetSeed(1);
 
         exit(Random(MaxValue));


Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
```diff
@@ -220,9 +221,20 @@ codeunit 130500 "Any"
...
         Seed := NewSeed;
+        SeedSet := true;
         Randomize(Seed);
     end;
 
+    /// <summary>.
+    /// Sets the default Seed for Pseudo-random number generation (no. of milliseconds since midnight of today).
+    /// Setting this value will change the numbers returned.
+    /// </summary>
+    procedure SetDefaultSeed()
```
nikolakukrika, line 232: This method should not be added. Main reason is - it uses a random seed, which means that it will not be possible to reproduce the failure that occurred in the gate. This is rather important, for this reason we always set a seed to a specific number so the numbers are pseudo-random.

We should not encourage the usage of the randomly generated numbers, since it will cause issues with debugging down the road.

If anyone wants to use a random generator they can generate the seed randomly and assign it via SetSeed. You can also run a test in the  for loop and increment the seed if you want to test the different seeds and flush out any rounding issues and so on...

Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
```diff
@@ -231,7 +243,7 @@ codeunit 130500 "Any"
     /// <returns>Pseudo-random integer value</returns>
     local procedure GetNextValue(MaxValue: Integer): Integer
     begin
-        if Seed = 0 then
+        if (not SeedSet) then
```
nikolakukrika, line 246: We need and Seed = 0 if we want to keep a global Boolean to track if the Seed has not been set, otherwise we risk resetting it too many times. Then this change would allow setting seed to 0, however I'm not sure if that is needed...

nikolakukrika (changes requested):

Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
hemisphera: > Main reason is - it uses a random seed, which means that it will not be possible to reproduce the failure that occurred in the gate.

I agree with you about the debuggability, you are absolutely right. But there are scenarios where a fixed seed is cumbersome. I'm thinking about an extensive test library that has various functions that generates random data. Those functions have a local copy of `Any`. And therefore every time I call it, I get the same number because I always start at the same seed. So I really only have three options:
- I pass around a "root" instance of `Any` as a parameter to the function, which is obviously not very handy
- I have a global `Any` in my library codeunit, but that means that I always must make sure that it is correctly initialized before I call any function
- I use `SetSeed` to initialize my local `Any` instance. The way that `Randomize` does initialize the seed is exactly what I would do in that case. It would therefore be handy, if we had access to that same behavior in `Any`. Otherwise I need to reimplement the same logic in all my test libraries.

> We should not encourage the usage of the randomly generated number

I agree on this as well. But adding a method does not encourage anything IMHO. In most cases the developer will not even call `SetSeed` or any other method of initializing it, because seeding it with 1 is fine. If, on the other hand, I don't want that, I should be able to reuse existing logic.

> If anyone wants to use a random generator they can generate the seed randomly and assign it via SetSeed.

Same as above. That is exactly my point. `Randomize` already does this for us. But the codeunit `Any` abstracts this away. I agree that we can do this on or own, but it's already there, it already does exactly what we need.

Modules/DevTools/TestFramework/TestLibraries/Any/src/Any.Codeunit.al
hemisphera: Not sure I understand. Resetting what? This does not change anything. The behavior is the same as before. The only difference is that you do not always know the actual seed itself, you only know if it has been set or not. Which is all we really need to know.

Setting a Seed to 0 is not really needed, but I also don't think it should be prohibited? Not sure why it would make a difference.

JesperSchulz (approved):
Let's go with this :-)

JesperSchulz:

PR #16096, opened 2022-01-28
Allow Wildcards and Filter on Test Codeunit Name

ChrisBlankDe:
Now Filtering on CodeunitName is possible like its possible on TestMethodName.


Modules/DevTools/TestFramework/TestRunner/src/CommandLineTestTool.Page.al
```diff
@@ -355,24 +355,25 @@ page 130455 "Command Line Test Tool"
 
         CodeunitTestMethodLine.SetRange("Test Suite", GlobalALTestSuite.Name);
         CodeunitTestMethodLine.SetRange("Line Type", CodeunitTestMethodLine."Line Type"::Codeunit);
-        CodeunitTestMethodLine.SetRange(Name, CodeunitName);
-        if not CodeunitTestMethodLine.FindFirst() then
+        CodeunitTestMethodLine.SetFilter(Name, CodeunitName);
```
nikolakukrika, line 358: We need to escape it if we are placing a filter SetFilter(Name,'%1',CodeunitName), otherwise special characters may get in the way. Setfilter is tricky, it could lead to some issues with special characters.

ChrisBlankDe: I have also thought about this. the problem is that imho there is no optimal solution.
escaping problems have to be solved by the dev who calls the function. unfortunately we can't know if the given special character is a control character for filtering or a character in the object name.

for example at `codeunit 137062 "SCM Sales & Receivables"`. 
How is the application supposed to know whether codeunit "SCM Sales & Receivables" is meant or the two code units "SCM Sales" and "Receivables"? that's why i think the developer who uses the procedure/field has to decide and escape if necessary.

So i solved it like the existing solution (7 lines below) where same problem exist.

nikolakukrika: How do we disable "SCM Sales & Recievables" then or a single method that has & characters in it?
I'm afraid that we may disable too much with this function.

ChrisBlankDe: Dev should use filters as [documented](https://docs.microsoft.com/en-us/dynamics365/business-central/ui-enter-criteria-filters#symbols).

> If you want to filter on any of these symbols, place the filter expression in single quotes ('<expression with symbol>'). For example, if you wanted to filter on records that start with the text J & V, the filter expression would be 'J & V*'.

nikolakukrika: Can we get a new field / method to disable the tests with filters? We could alternatively implement the method to Get the tests with filter and then specify the same payload to disable the tests? 

I'm worried that the special characters will break the regular usage. The developers often put the special characters in method or codeunit names, and this will be a hard break to all users.

Modules/DevTools/TestFramework/TestRunner/src/CommandLineTestTool.Page.al
```diff
@@ -355,24 +355,25 @@ page 130455 "Command Line Test Tool"
...
-
+        repeat
+            TestMethodLine.SetRange("Test Suite", GlobalALTestSuite.Name);
+            TestMethodLine.SetRange("Line Type", "Line Type"::"Function");
+            TestMethodLine.SetRange("Test Codeunit", CodeunitTestMethodLine."Test Codeunit");
+            TestMethodLine.SetFilter(Name, TestMethodName);
+            TestMethodLine.ModifyAll(Run, false);
+
+
+            TestMethodLine.SETRANGE(Name);
```
PeterConijn, line 369: SetRange is in PascalCase in AL

Modules/DevTools/TestFramework/TestRunner/src/CommandLineTestTool.Page.al
```diff
@@ -355,24 +355,25 @@ page 130455 "Command Line Test Tool"
...
+            TestMethodLine.SETRANGE(Run, TRUE);
+            if TestMethodLine.IsEmpty() then begin
+                CodeunitTestMethodLine.VALIDATE(Run, FALSE);
+                CodeunitTestMethodLine.Modify(TRUE);
+            end;
+        until CodeunitTestMethodLine.Next = 0;
```
PeterConijn, line 375: Missing parentheses after Next

Modules/DevTools/TestFramework/TestRunner/src/CommandLineTestTool.Page.al
```diff
@@ -355,24 +355,25 @@ page 130455 "Command Line Test Tool"
...
+            TestMethodLine.SetRange("Line Type", "Line Type"::"Function");
+            TestMethodLine.SetRange("Test Codeunit", CodeunitTestMethodLine."Test Codeunit");
+            TestMethodLine.SetFilter(Name, TestMethodName);
+            TestMethodLine.ModifyAll(Run, false);
+
+
+            TestMethodLine.SETRANGE(Name);
+            TestMethodLine.SETRANGE(Run, TRUE);
+            if TestMethodLine.IsEmpty() then begin
+                CodeunitTestMethodLine.VALIDATE(Run, FALSE);
```
PeterConijn, line 372: Validate should be in PascalCase and TRUE/FALSE should be lowercase

Modules/DevTools/TestFramework/TestRunner/src/CommandLineTestTool.Page.al
```diff
@@ -355,24 +355,25 @@ page 130455 "Command Line Test Tool"
...
+                CodeunitTestMethodLine.Modify(TRUE);
```
PeterConijn, line 373: TRUE in lowercase

JesperSchulz:
@ChrisBlankDe / @nikolakukrika, this PR has been sitting idle for a year now. How far are we from getting the code approved in a review? I think we should either get this processed asap, or closed. What do you think? Can we land this one?

ChrisBlankDe:
> @ChrisBlankDe / @nikolakukrika, this PR has been sitting idle for a year now. How far are we from getting the code approved in a review? I think we should either get this processed asap, or closed. What do you think? Can we land this one?

i had hoped to improve the test runner for us and other partners.
I can understand the design issues of @nikolakukrika. however, I strictly follow the example of Microsoft a few lines below.
the problems listed appear in the same way in msft's solution.
fixing microsoft's code at this point would be a breaking change and would have to be synchronized with the bccontainerhelper.

by the way, the casing corresponds to the microsoft standard at the time of the PR. funny that after one year suddenly someone cares. and obviously more people are bothered than the weaknesses in the process :)

I can correct the casing if there is a chance that the pull request will be adopted.

JesperSchulz:
@nikolakukrika, could you help driving this PR to a conclusion?

Modules/DevTools/TestFramework/TestRunner/src/CommandLineTestTool.Page.al
nikolakukrika: Can we do a fix like this:
CodeunitTestMethodLine.SetFilter(Name, CodeunitName);
if CodeunitTestMethodLine.IsEmpty() then
     CodeunitTestMethodLine.SetRange(Name, CodeunitName);
     
 This way if we have a broken filter, it will use the setrange that will escape the values

nikolakukrika (changes requested):

Modules/DevTools/TestFramework/TestRunner/src/CommandLineTestTool.Page.al
ChrisBlankDe: this is a good compromise.
I think the change comes closest to what the developer expects.
I will update the pr.

nikolakukrika (approved):

JesperSchulz (approved):
Let's do this!

JesperSchulz:
Shipping with 23.0 🥳

JesperSchulz:

Build ID: 11839.

PR #16104, opened 2022-01-28
TempBlob Create*Stream with a stream as return

BenjaminHochstetter:
It would create if the tempblob would return the created streams, to pass them direct into other functions


BenjaminHochstetter:
Sorry I have "messed up" this branch.
I'll redo it with testing and encoding.

PR #16310, opened 2022-02-16
TempBlob Create*Stream with a stream as return

BenjaminHochstetter:
It would be useful if the Create*Steam functions returned the stream directly so that it can be passed on directly into a convert function, for example.


achim-t:
@BenjaminHochstetter 
You forgot to add TempBlobImpl.

BenjaminHochstetter:
@achim-t
I used the already existing functions in TempBlobImpl. There was no need to implement it different.

achim-t:
Oh, right. Sorry about that.

mazhelez (approved):

JesperSchulz:
Changes will release with 20.0.

JesperSchulz:

PR #16355, opened 2022-02-18
BLOB Storage Module Extended: ToRecord()

achim-t:


grobyns:
How is this PR different from #9874 and/or #12281 ?

PR #17410, opened 2022-04-27
GP to BC migration enhancements

ToddE123:
Added ability to exclude Inactive Checkbooks from the GP to BC cloud migration. 
  - A new option was added to the Cloud Migration Setup form allowing users to indicate if they want to migration inactive checkbooks or not.  
  - The option is labeled: Migrate Inactive Checkbooks.  
  - By default, the option will be selected.
  - If that option is marked, any checkbook marked as inactive in GP WILL BE migrated to BC as a bank account for the company or companies selected.
  - If the option is unmarked, any checkbook marked as inactive WILL NOT be migrated from GP to BC for the company or companies selected. 
  - The help text at the bottom of the form was also updated to the following:  Choose whether to migrate inactive customers, vendors and checkbooks, or uncheck the boxes to only migrate those that are active. 

Remit To address changes
  - Migrate the Vendor Remit to Address to be the Main Address in BC.
  - The Primary address as well as any other addresses in GP will be migrated to Order Addresses in BC.   

No need to migrate default 0000...0 Phone No. and Fax No. values so the migration was changed to leave those blank instead.


Apps/W1/HybridGP/app/src/Migration/Support/CheckBooks/GPCheckbookMSTR.table.al
```diff
@@ -182,20 +182,27 @@ table 40099 "GP Checkbook MSTR"
...
-                    BankAccount.Name := DelChr(DSCRIPTN, '>', ' ');
-                    BankAccount."Bank Account No." := DelChr(BNKACTNM, '>', ' ');
-                    BankAccount."Last Check No." := GetLastCheckNumber(NXTCHNUM);
-                    BankAccount."Balance Last Statement" := Last_Reconciled_Balance;
-                    BankAccount."Bank Acc. Posting Group" := GetBankAccPostingGroup(ACTINDX);
-                    UpdateBankInfo(DelChr(BANKID, '>', ' '), BankAccount);
-                    BankAccount.Insert(true);
-                end;
+                if not BankAccount.Get(CHEKBKID) then
+                    if MigrateInactiveCheckbooks or not INACTIVE then begin
```
mazhelez, line 195: Why don't you just filter out inactive entries if `MigrateInactiveCheckbooks` is set to false? That way you simply won't loop through them.

mazhelez:
That's one great contribution! 😄

Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
```diff
@@ -173,10 +173,18 @@ codeunit 4018 "GP Customer Migrator"
             CopyStr(MigrationGPCustomer.ADDRESS2, 1, 50), Country, CopyStr(MigrationGPCustomer.ZIPCODE, 1, 20),
             CopyStr(MigrationGPCustomer.CITY, 1, 30));
         CustomerDataMigrationFacade.SetContact(CopyStr(MigrationGPCustomer.CNTCPRSN, 1, 50));
+
+        if (CopyStr(MigrationGPCustomer.PHONE1, 1, 14) = '00000000000000') then
```
nikolakukrika, line 177: Minor - Consider introducing a function to check if it is blank and protentional generating the 00000000000000 or hardcoding it as a string. You could also try Replace(0,'') and that one would make it blank.

Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
```diff
@@ -173,10 +173,18 @@ codeunit 4018 "GP Customer Migrator"
...
+            MigrationGPCustomer.PHONE1 := '';
+
+        if (CopyStr(MigrationGPCustomer.FAX, 1, 14) = '00000000000000') then
+            MigrationGPCustomer.FAX := '';
```
nikolakukrika, line 181: It seems that MigrationGPCustomer is not modified. If we are not updating the value, maybe it is better to assign it to a variable and then use the variable instead. It makes the intent more clear.

Apps/W1/HybridGP/app/src/Migration/Customers/tab4048.GPCustomerAddress.al
```diff
@@ -91,6 +91,12 @@ table 4048 "GP Customer Address"
             ShipToAddress.County := STATE;
             ShipToAddress."Tax Area Code" := TAXSCHID;
 
+            if (CopyStr(ShipToAddress."Phone No.", 1, 14) = '00000000000000') then
```
nikolakukrika, line 94: Same comments on these fields as for Customer, see if they apply

Apps/W1/HybridGP/app/src/Migration/Support/CheckBooks/GPCheckbookMSTR.table.al
```diff
@@ -182,20 +182,27 @@ table 40099 "GP Checkbook MSTR"
     procedure MoveStagingData()
     var
         BankAccount: Record "Bank Account";
+        GPCompanyAdditionalSettings: Record "GP Company Additional Settings";
+        MigrateInactiveCheckbooks: Boolean;
     begin
+        MigrateInactiveCheckbooks := false;
+        if GPCompanyAdditionalSettings.Get(CompanyName()) then
```
nikolakukrika, line 189: Haven't we extracted these to a method on the GPCompanyAdditionalSettings?
There was a method that was called GetMigrateInactiveCheckbooks that would return if they are included or not. This way we can track where the feature is used and we do not duplicate the logic.

Apps/W1/HybridGP/app/src/Migration/Support/CheckBooks/GPCheckbookMSTR.table.al
```diff
@@ -182,20 +182,27 @@ table 40099 "GP Checkbook MSTR"
...
-                    BankAccount."Bank Account No." := DelChr(BNKACTNM, '>', ' ');
-                    BankAccount."Last Check No." := GetLastCheckNumber(NXTCHNUM);
-                    BankAccount."Balance Last Statement" := Last_Reconciled_Balance;
-                    BankAccount."Bank Acc. Posting Group" := GetBankAccPostingGroup(ACTINDX);
-                    UpdateBankInfo(DelChr(BANKID, '>', ' '), BankAccount);
-                    BankAccount.Insert(true);
-                end;
+                if not BankAccount.Get(CHEKBKID) then
+                    if MigrateInactiveCheckbooks or not INACTIVE then begin
+                        BankAccount.Init();
```
nikolakukrika, line 196: Init only clears non-key fields. It is useless. You should use Clear(BankAccount) here

Apps/W1/HybridGP/app/src/Migration/Support/CheckBooks/GPCheckbookMSTR.table.al
```diff
@@ -182,20 +182,27 @@ table 40099 "GP Checkbook MSTR"
...
+                        BankAccount."No." := DelChr(CHEKBKID, '>', ' ');
```
nikolakukrika, line 197: Minor - we have new constructs for deleting, replacing and other string operations. Feel free to wontfix this one, it is more FYI

nikolakukrika (approved):

Apps/W1/HybridGP/app/src/Migration/Support/CheckBooks/GPCheckbookMSTR.table.al
nikolakukrika: Be careful here of two things:
1. You are on the Rec. You could introduce a local variable and set the filters from the rec by using copy. This way you will not change the filters that are on the Rec. 
2. Check if there is an index on this field. If the table is large it is good to add the index so the search is faster.

Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
jaymckinney-enavate: I have created shared function to handle this.

Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
jaymckinney-enavate: I have updated this as well to use the new shared function.

Apps/W1/HybridGP/app/src/Migration/Support/CheckBooks/GPCheckbookMSTR.table.al
jaymckinney-enavate: I will make this change.

Apps/W1/HybridGP/app/src/Migration/Support/CheckBooks/GPCheckbookMSTR.table.al
jaymckinney-enavate: Thanks for the information. Is there any documentation on this that you can link to?

Apps/W1/HybridGP/app/src/Migration/Support/CheckBooks/GPCheckbookMSTR.table.al
jaymckinney-enavate: I'm going to hold off on this change for now.

Apps/W1/HybridGP/app/src/Migration/Support/CheckBooks/GPCheckbookMSTR.table.al
pri-kise: @jaymckinney-enavate 
I think @nikolakukrika is referin to the new string operations that were added to AL, e.g. TrimEnd (https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/text/text-trimend-method)

ToddE123: we will look to update the entire extension with the new constructs in a later PR.

PR #17512, opened 2022-05-03
Respect string length for Encrypt procedure in CryptographyManagement.Codeunit.al

sirhc101:
Input string for encryption cannot exceed a length of 215 plain characters. (see https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/system/system-encrypt-method)

This should be reflected in the Encrypt procedure in ryptographyManagement.Codeunit.al as well

### DIFF
--- a/Modules/System/Cryptography Management/src/CryptographyManagement.Codeunit.al
+++ b/Modules/System/Cryptography Management/src/CryptographyManagement.Codeunit.al
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------------------------------------------
+// ------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


mazhelez:
@sirhc101 The way you designed the solution won't work because `Encrypt` is now ambiguous.
As a solution, you can rename `Encrypt(InputString: Text[215])` to `EncryptInput(InputString: Text[215])` (or somthing like that).

Throwing an error in the original `Encrypt` is fine as well. But do you want the "overload" with Text[215] because you prefer to get a compilation error/warning or you're fine with having the runtime error?

mazhelez:
@sirhc101 The way you designed the solution won't work because `Encrypt` is now ambiguous.
As a solution, you can rename `Encrypt(InputString: Text[215])` to `EncryptInput(InputString: Text[215])` (or somthing like that).

Throwing an error in the original `Encrypt` is fine as well. But do you want the "overload" with Text[215] because you prefer to get a compilation error/warning or you're fine with having the runtime error?

sirhc101:
@mazhelez Yes, you are absolutely right, I just quickly edit the source code on GitHub and did not saw the obvious error.
Sorry for that.

Anyway, I personally would prefer a compiler error/warning instead of an error message during execution, because this would be the same as passing more than 215 characters.
But I guess just change `Text` to `Text[215]` is not possible, due to a breaking change, isn't?

How could we solve this, so others will not fall into this trap?

PR #17882, opened 2022-05-09
Bank Deposit App install. Break company initialize.

vlkov:
Source Code Setup must be created only in company-initialize codeunit. So, when we use My Company in cicd process the source codes are not created in base application, because the record in source code setup already exists.


dcenic:
We are already looking to fix this issue in Microsoft's version of Bank Deposits app and roll it out as a hotfix to Business Central 20.0 and Business Central 20.1. This is because it currently breaks the upgrade to Business Central 20.0 for 6 Live Site tenants.
Please hold off with this pull request - Microsoft will roll out a hotfix that addresses this.

PR #18086, opened 2022-05-18
[Azure BLOB Storage] Fix BLOB property values when listing BLOBs (Lis…

### LINKED ISSUE #17520: Azure blob list not returning Content Length in ABS Container Content table
Hello,

currently I'm using the "ABS Blob Client" codeunit to get a list of all blobs.
Everything works except that I don't receive the Content-Length and Content-Type of a file.

Screen Shot 2022-05-03 at 15 45 01

Can it be a problem in the codeunit 9054 "ABS Container Content Helper". Where the field is searched by returned properties.
Screen Shot 2022-05-03 at 15 51 13

And if I'm reading correctly here https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs. The return property is Content-Type and the field name is Content Type. So without he -.

Thanks in advance,
Gunter

mazhelez:
_**Problem:**_
When BLOB properties are filled in, the convention is to use the name of the corresponding fields in table "ABS Container Content". However, the fields in the table do not match the BLOB properties as defined in https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs#response-body

_**Solution:**_
To avoid breaking changes, keep the fields as they are and use the fields captions to determine the corresponding properties. 

Fixes #17520


Modules/System/Azure Blob Services API/src/ABSContainerContent.Table.al
```diff
@@ -49,23 +49,27 @@ table 9043 "ABS Container Content"
...
             DataClassification = SystemMetadata;
-            Caption = 'Last Modified';
+            Caption = 'Last-Modified';
+            Description = 'Caption matches the corresponding property as defined in https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs#response-body';
         }
         field(13; "Content Length"; Integer)
         {
             DataClassification = SystemMetadata;
-            Caption = 'Content Length';
+            Caption = 'Content-Length';
```
pri-kise, line 58: Shouldn't the Captions beeing set to `Locked = true` ?

mazhelez: Very true :) Thanks!

darjoo (approved):
Initial changes were ported in with the 20.2 update

PR #18678, opened 2022-06-16
Migrate GP Vendor Classes

jaymckinney-enavate:
This change enhances the GP cloud migration to include migrating GP Vendor Classes to BC Vendor Posting Groups.

- For each GP vendor class that has related accounts a new Vendor Posting Group will be created in BC.
- Any vendor migrated from GP that is part of a vendor class will have the related Vendor Posting Group assigned.
- A configuration option has been added to give the ability to not migrate Vendor Classes, disabled by default.

### DIFF
--- a/Apps/W1/HybridGP/app/Permissions/HybridGPEdit.PermissionSet.al
+++ b/Apps/W1/HybridGP/app/Permissions/HybridGPEdit.PermissionSet.al
@@ -67,9 +67,11 @@ permissionset 4031 "HybridGP - Edit"
                     tabledata "GP Migration Errors" = IMD,
                     tabledata "GP Segment Name" = IMD,
                     tabledata "GP Company Additional Settings" = IMD,
-                    tabledata "GP SY40100" = RIMD,
-                    tabledata "GP SY40101" = RIMD,
-                    tabledata "GP CM20600" = RIMD,
+                    tabledata "GP SY40100" = IMD,
+                    tabledata "GP SY40101" = IMD,
+                    tabledata "GP CM20600" = IMD,
                     tabledata "GP MC40200" = IMD,
-                    tabledata "GP SY06000" = IMD;
+                    tabledata "GP SY06000" = IMD,
+                    tabledata "GP PM00100" = IMD,
+                    tabledata "GP PM00200" = IMD;
 }
--- a/Apps/W1/HybridGP/app/src/Migration/Vendors/GPVendorMigrator.codeunit.al
+++ b/Apps/W1/HybridGP/app/src/Migration/Vendors/GPVendorMigrator.codeunit.al
@@ -455,4 +455,79 @@ codeunit 4022 "GP Vendor Migrator"
             Vendor.Modify(true);
         end;
     end;
+
+    procedure MigrateVendorClasses()
+    var
+        GPCompanyAdditionalSettings: Record "GP Company Additional Settings";
+        GPPM00200: Record "GP PM00200";
+        GPPM00100: Record "GP PM00100";
+        VendorPostingGroup: Record "Vendor Posting Group";
+        Vendor: Record Vendor;
+        HelperFunctions: Codeunit "Helper Functions";
+        ClassId: Text[20];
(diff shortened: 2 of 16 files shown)


Apps/W1/HybridGP/app/Permissions/HybridGPEdit.PermissionSet.al
```diff
@@ -67,9 +67,11 @@ permissionset 4031 "HybridGP - Edit"
                     tabledata "GP Migration Errors" = IMD,
                     tabledata "GP Segment Name" = IMD,
                     tabledata "GP Company Additional Settings" = IMD,
-                    tabledata "GP SY40100" = RIMD,
-                    tabledata "GP SY40101" = RIMD,
-                    tabledata "GP CM20600" = RIMD,
+                    tabledata "GP SY40100" = IMD,
```
nikolakukrika, line 70: This could be a breaking change if we have shipped with it.  If it is needed and you feel comfortable that nobody could have taken dependency on it we can change the permissions

Apps/W1/HybridGP/app/src/Migration/Vendors/GPVendorMigrator.codeunit.al
```diff
@@ -455,4 +455,79 @@ codeunit 4022 "GP Vendor Migrator"
...
+        VendorPostingGroup: Record "Vendor Posting Group";
+        Vendor: Record Vendor;
+        HelperFunctions: Codeunit "Helper Functions";
+        ClassId: Text[20];
+        AccountNumber: Code[20];
+        MigrateVendorClasses: Boolean;
+    begin
+        MigrateVendorClasses := GPCompanyAdditionalSettings.GetMigrateVendorClasses();
+
+        if not MigrateVendorClasses or not GPPM00200.FindSet() then
```
nikolakukrika, line 473: It is a good practice to split IFs if there is a transaction within it.
AL will execute both conditions. So it could be better to extract Findset to a separate If statement.

nikolakukrika (approved):

Apps/W1/HybridGP/app/src/tables/GPCompanyAdditionalSettings.table.al
```diff
@@ -17,6 +16,11 @@ table 40105 "GP Company Additional Settings"
             InitValue = true;
             DataClassification = SystemMetadata;
         }
+        field(11; "Migrate Vendor Classes"; Boolean)
+        {
+            InitValue = true;
```
nikolakukrika, line 21: Same comment here - Check with Kasey if the value should be true by default.

nikolakukrika (approved):

Apps/W1/HybridGP/app/Permissions/HybridGPEdit.PermissionSet.al
jaymckinney-enavate: This corrects a mistake from a previous commit. All of the permissions here are IMD, and this permissionset includes 'HybridGP - Read', so having 'R' in this file is redundant.

Apps/W1/HybridGP/app/src/tables/GPCompanyAdditionalSettings.table.al
```diff
@@ -37,4 +41,15 @@ table 40105 "GP Company Additional Settings"
 
         exit(MigrateInactiveCheckbooks);
     end;
+
+    procedure GetMigrateVendorClasses(): Boolean
+    var
+        MigrateVendorClasses: Boolean;
+    begin
+        MigrateVendorClasses := true;
```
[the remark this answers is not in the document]
jaymckinney-enavate: Confirmed, she said they should be false by default. Good catch.

nikolakukrika (approved):

JesperSchulz:
Availability update: We will publish a fix for this issue in the next update for release 20.

Build ID to track: 43275.

JesperSchulz:

Build ID: 43303.

PR #18723, opened 2022-06-20
:bug: Fix Bank Deposit PermissionSet Name

pri-kise:
Current PermissionSet Name _"Bank Deposits - Objects"_ didn't respect the [rule AL0305](https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/diagnostics/diagnostic-al305) and did cause errors


Apps/W1/BankDeposits/app/src/permissions/BankDepositsObj.PermissionSet.al
```diff
@@ -3,7 +3,7 @@
 /// do not include this permission set in any other permission set
 /// and do not change the Access and Assignable properties
 /// </summary>
-permissionset 1690 "Bank Deposits - Objects"
+permissionset 1690 "Bank Deposits - Obj."
```
mazhelez, line 6: The permission set is public, so this would be a breaking change.

The permission set is not assignable so the length should not be a problem.

stkillen: @pri-kise Thanks for reporting this. However as Maria said, fixing this on the application side by renaming the permission set is a breaking change. We'll look into the issue on the platform side.

mazhelez:
As per Stig's last comment, we will address the issue in the platform.

PR #18724, opened 2022-06-20
Update pull_request_target Branch Name to main

pri-kise:
My Latest PR didn't trigger the CI.
I updated PR Branch Name from old name master to main.

This change will only be used for new PR, as soon as this change is merged to the main branch.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -2,7 +2,7 @@ name: Build
 
 on:
   pull_request_target:
-    branches: [ master ]
+    branches: [ main ]
 
 jobs:
   SystemApplication:


JesperSchulz (approved):

JesperSchulz:
We're actually working on adopting the CI process from AL:Go for GitHub (Freddy's project). Should very soon be ready :-)

pri-kise:
> We're actually working on adopting the CI process from AL:Go for GitHub (Freddy's project). Should very soon be ready :-)

I had this idea yesterday, too and hoped that this might happen some time. I'm really looking  forward to it.
Now we only need some way to Run the AL-Validation to avoid breaking changes.

PR #18960, opened 2022-06-30
Migrate GP Customer Classes

jaymckinney-enavate:
This change enhances the GP cloud migration to include migrating GP Customer Classes to BC Customer Posting Groups.

- For each GP customer class that has related accounts a new Customer Posting Group will be created in BC.
- Any customer migrated from GP that is part of a customer class will have the related Customer Posting Group assigned.
- A configuration option has been added to give the ability to not migrate Customer Classes, disabled by default.

### DIFF
--- a/Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
+++ b/Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
@@ -375,4 +375,80 @@ codeunit 4018 "GP Customer Migrator"
         HelperFunctions.UpdateFieldWithValue(RecordVariant, MigrationGPCustTrans.FieldNo(GLDocNo), DocumentNo);
     end;
 
+    procedure MigrateCustomerClasses()
+    var
+        GPCompanyAdditionalSettings: Record "GP Company Additional Settings";
+        GPRM00101: Record "GP RM00101";
+        GPRM00201: Record "GP RM00201";
+        CustomerPostingGroup: Record "Customer Posting Group";
+        Customer: Record Customer;
+        HelperFunctions: Codeunit "Helper Functions";
+        ClassId: Text[20];
+        AccountNumber: Code[20];
+        MigrateCustomerClasses: Boolean;
+    begin
+        if not GPRM00101.FindSet() then
+            exit;
+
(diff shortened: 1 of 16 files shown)


Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
```diff
@@ -375,4 +375,80 @@ codeunit 4018 "GP Customer Migrator"
...
+
+        repeat
+            Clear(GPRM00201);
+            Clear(CustomerPostingGroup);
+
+            ClassId := GPRM00101.CUSTCLAS.Trim();
+            if ClassId <> '' then
+                if Customer.Get(GPRM00101.CUSTNMBR) then begin
+                    if not CustomerPostingGroup.Get(ClassId) then
+                        if GPRM00201.Get(ClassId) then begin
```
nikolakukrika, line 405: Suggestion - is it better to move the customer posting groups first, regardless, then to assign them as we create customers? We could save on the SQL queries.

jaymckinney-enavate: Good suggestion, I made the change.

Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
```diff
@@ -375,4 +375,86 @@ codeunit 4018 "GP Customer Migrator"
...
+                        CustomerPostingGroup.Validate("Payment Tolerance Debit Acc.", AccountNumber);
+                        CustomerPostingGroup.Validate("Payment Tolerance Credit Acc.", AccountNumber);
+                    end;
+
+                    CustomerPostingGroup.Insert();
+                end;
+        until GPRM00201.Next() = 0;
+
+        // Assign the Customer Posting Groups to the Customers
+        repeat
```
nikolakukrika, line 451: Suggestion - could we move this code to the place where we are creating customers? We should ensure that Posting Groups are created before. 

It would save us one loop on the customers and 1 modify query on the customer.

nikolakukrika (approved):

JesperSchulz:
Availability update: We will publish a fix for this issue in the next update for release 20.

Build ID to track: 43275.

PR #19505, opened 2022-07-29
Added bitwise operations to Math

PeterConijn:
The math codeunit helps a great deal with mathematical operations, but the one thing that Dynamics has always lacked is bitwise operations that are sometimes used in e.g. communication and validation with external systems.


Modules/System/Math/src/Math.Codeunit.al
```diff
@@ -288,5 +288,82 @@ codeunit 710 Math
...
+    begin
+        InitBitwiseValues(BitPosition, Result);
+        for BitIndex := to 32 do begin
+            if (IntegerValue1 mod 2) + (IntegerValue2 mod 2) = 1 then
+                UpdateBitwiseResult(BitPosition, Result);
+            UpdateBitwiseValues(IntegerValue1, IntegerValue2, BitPosition);
+        end;
+    end;
+
+    local procedure InitBitwiseValues(var BitPosition: Integer; var Result: Integer)
```
darjoo, line 351: Create an implementation codeunit and move these local functions there. It wasn't created before because the previous functions either called a dotnet(implementation) or direct value.

Modules/System/Math/src/Math.Codeunit.al
```diff
@@ -288,5 +288,82 @@ codeunit 710 Math
...
+    /// </summary>
+    /// <param name="IntegerValue1">The first integer for the evaluation</param>
+    /// <param name="IntegerValue2">The second integer for the evaluation</param>
+    /// <returns>The resulting integer of the bitwise exclusive OR (XOR) operaion</returns>
+    procedure BitwiseXOr(IntegerValue1: Integer; IntegerValue2: Integer) Result: Integer
+    var
+        BitPosition: Integer;        
+        BitIndex: Integer;
+    begin
+        InitBitwiseValues(BitPosition, Result);
```
darjoo, line 343: Once the implementation codeunit is created, these functions should call the same function there.

darjoo (commented):
Add some tests for these new functions :)

Modules/System/Math/src/Math.Codeunit.al
PeterConijn: I will rewrite using the implementation codeunit pattern, but writing tests for a mathematical operation that cannot otherwise be calculated in Dynamics is going to be tricky. This is probably why there is currently no test for the math codeunit that I can see.

The only way I can reasonably do that is by creating a reasonably large set of integers and their outcomes and see if they match. It's limited, but if you're ok with that, I will use that approach.

Modules/System/Math/src/MathImpl.Codeunit.al
```diff
@@ -0,0 +1,63 @@
...
+    procedure BitwiseAnd(IntegerValue1: Integer; IntegerValue2: Integer) Result: Integer
+    var
+        BitPosition: Integer;
+        BitIndex: Integer;
+    begin
+        InitBitwiseValues(BitPosition, Result);
+
+        for BitIndex := to 32 do begin
+            if (IntegerValue1 mod 2 = 1) and (IntegerValue2 mod 2 = 1) then
+                UpdateBitwiseResult(BitPosition, Result);
```
darjoo, line 14: For readability, it'd be preferred to change Result to a local variable and exit(Result) at the end.

Modules/System/Math/src/Math.Codeunit.al
darjoo: These math functions are deterministic, so despite the tests going to be rudimentary, that's alright.
I'm not sure why tests weren't written for the others, but my assumption is that they're using the system.math in dotnet or constants, so that should've been tested thoroughly, and we would not need to.

Modules/System/Math/src/MathImpl.Codeunit.al
darjoo: Once the tests and this change are in, I'll process this today.

darjoo: One more thing, I think these functions will only work for positive numbers. If that's the case, either add validation that the parameters cannot be negative or ensure these will work for negative numbers too.

PeterConijn: I will add that to my test and add validation as needed.

PeterConijn:
Added a test codeunit; not sure whether you wanted me to add an app.json for that. Secondly, I could not determine a codeunit number for the test, since I am not sure which logic is behind that.

mazhelez:
@PeterConijn that's great effort from your side and we are grateful to have contributors like you!

However, for this particular functionality it is wise to wait for it to be available in the language as the current solution isn't performant and does not cover all the cases.

That said, it's also crucial to acknowledge that we cannot share any details on when bitwise operations will be available in the AL language. What we are also willing to do in the meantime, however, is to expose bitwise operations in some of the .net libraries we use and then expose these in System Application via .net calls. How does that sound?

Of course, if you feel too strong about your contribution, we will reconsider it :)

PeterConijn:
Hi Maria,

That sounds like a wonderful idea and will solve a lot of challenges. Thank you!


Peter Conijn
Application Developer
t: +31742051199 | m: +31627332165
e: ***@***.*** | w: www.dysel.com

IhorHandziuk:
Hi @PeterConijn, thanks for your contribution!

I just have some concerns about the performance of the current implementation (as well as potentially unhandled edge cases).
Would you be open to change the implementation to just wrap the existing C# functionality?

Here is an example:
```
  procedure BitwiseOr(Left: Integer; Right: Integer): Integer
  var
      Result: DotNet BigInteger;
      DotNetLeft: DotNet BigInteger;
      DotNetRight: DotNet BigInteger;
      ResultInt: Integer;
  begin
      DotNetLeft := DotNetLeft.BigInteger(Left);
      DotNetRight := DotNetLeft.BigInteger(Right);
      Result := Result.op_BitwiseOr(DotNetLeft, DotNetRight);
      Evaluate(ResultInt, Result.ToString());
      exit(ResultInt);
  end
```

PR #19202, opened 2022-07-15
Limit number of Historical GL years being migrated

ToddE123:
The GP migration wizard will include a field where users can enter the oldest history year they want to migrate. If nothing is entered, the migration will work as it does today. If they enter a valid year, a beginning balance for all balance sheet accounts will be created in BC for that year along with subsequent GL transactions.

PR #19931, opened 2022-08-23
Intermittent beginning balance creation

ToddE123:
The Account beginning balance is not always getting properly created due to Get() being used instead of FindFirst(). This PR changes Get() to FindFirst().


Apps/W1/HybridGP/app/src/Migration/Accounts/GPAccountMigrator.codeunit.al
```diff
@@ -74,6 +74,65 @@ codeunit 4017 "GP Account Migrator"
...
 
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"GL Acc. Data Migration Facade", 'OnCreateOpeningBalanceTrx', '', true, true)]
+    procedure OnCreateOpeningBalanceTrx(VAR Sender: Codeunit "GL Acc. Data Migration Facade"; RecordIdToMigrate: RecordId)
+    var
+        GPAccount: Record "GP Account";
+    begin
+        if RecordIdToMigrate.TableNo() <> Database::"GP Account" then
+            exit;
+
+        GPAccount.Get(RecordIdToMigrate);
```
nikolakukrika, line 85: Should we have if not GPAccount.Get(RecordIdToMigrate) then 
                               exit;

Just to be on the safe side?

Apps/W1/HybridGP/app/src/Migration/Accounts/GPAccountMigrator.codeunit.al
```diff
@@ -74,6 +74,65 @@ codeunit 4017 "GP Account Migrator"
...
+        GPGL10111.SetRange(PERIODID, 0);
+        GPGL10111.SetRange(YEAR1, InitialYear);
+        if not GPGL10111.FindFirst() then
+            exit;
+
+        BeginningBalance := GPGL10111.PERDBLNC;
+
+        PostingGroupCode := PostingGroupCodeTxt + format(InitialYear) + 'BB';
+        GPFiscalPeriods.SetRange(YEAR1, InitialYear);
+        if GPFiscalPeriods.FindFirst() then begin
```
nikolakukrika, line 117: Minor if not findfrist then
exit.

Then we have less nesting

nikolakukrika (approved):

Apps/W1/HybridGP/app/src/Migration/Accounts/GPAccountMigrator.codeunit.al
ToddE123: There appears to be an issue with the merge I did from main as there was only to be a single line changed. I believe we covered this same question in a previous PR (https://github.com/microsoft/ALAppExtensions/pull/19202). Perhaps I was mistaken as to when that PR was going into BC. This PR is only needed to fix that PR.

JesperSchulz:

Build ID: 45998.

Issue #19756, opened 2022-08-08
ABS Container Content - Folder Display


JulianTillmann:
There is something wonky with table 9043 "ABS Container Content" and from the Azure Blob Storage API. 

I've created a few test folders and files
![image](https://user-images.githubusercontent.com/35929603/183361277-b300f4d3-c59f-4e64-9ea0-56801bb58211.png)

and created subfolders 
![image](https://user-images.githubusercontent.com/35929603/183361327-b68b4988-3f9a-4b25-97e2-d66b49f0b069.png)

and subsubfolders
![image](https://user-images.githubusercontent.com/35929603/183361411-73a3d606-3822-4d39-ab95-fce2b98a51b2.png)
![image](https://user-images.githubusercontent.com/35929603/183361432-9c491969-7c31-44d0-b397-e9004be6dbf2.png)

then I created a Page on the ABS Container Content table with a filter to the content Type "Directory" and ShowAsTree with IndenationColumn "Level"
![image](https://user-images.githubusercontent.com/35929603/183361631-f16138f2-ef18-4714-aab9-0db96a706b4f.png)

unfortunately it seems like the first folder name is repeated for the following folders:
![image](https://user-images.githubusercontent.com/35929603/183361953-0d80ff6a-eb4d-4fda-97b2-66036000ec41.png)

could you please check if that's a bug or I'm doing something wrong?

JulianTillmann:
Additionally it would be great if Directories got a value for "Full Name" as well. Right now it would be difficult to know which folder is selected if the name is used elsewhere as well

AlexanderYakunin:
@JesperSchulz

JesperSchulz:
We'll look into it. It might take us a little while, though...

JulianTillmann:
That's too bad, we wanted to use this for creating an app :-/

mazhelez:
Hey @JulianTillmann nice catch!

The code for Azure Blob Storage API is all in this repo and, to be perfectly honest, the fastest way to get this issue fixed is... well, to do it yourself⭐ 
Just create a pull request with a fix and we'll gladly uptake it!

JulianTillmann:
I created PR #19976 that should hopefully fix it.

JesperSchulz:
That is awesome! Progress will be tracked in the new PR. Closing issue 🥳

PR #19976, opened 2022-08-25
Changed ABS Container API Directory Identification

JulianTillmann:
Fix for Issue #19756 

Directories were not properly identified


mazhelez:
Great :) The fix looks good enough!

Is there a way to write a test that would fail with the previous version, but succeeds with your changes?

JulianTillmann:
@mazhelez 
I don't know, if you have multiple folders in a test azure container or a mock you could maybe create a scenario.

mazhelez:
> @mazhelez I don't know, if you have multiple folders in a test azure container or a mock you could maybe create a scenario.

Hey, I'm not sure I understand the question...?

In any case, for automated tests (although they do not run automatically in the gates), we use [azurite](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio) and it should be possible to 
 write a test that creates blob in folders and subfolders and then retrieves them with the correct type and everything.

JulianTillmann:
ok, feel free to do it with azurite or any other way :-)

mazhelez:
> ok, feel free to do it with azurite or any other way :-)

@JulianTillmann the idea is that you (as a contributor) add the test 🦸

JulianTillmann:
> > ok, feel free to do it with azurite or any other way :-)
> 
> @JulianTillmann the idea is that you (as a contributor) add the test 🦸

I have enough on my plate already, I will not have the time to 

and since there has not been any tests for this before or the bug would have been noticed, this does not seem to be a priority

mazhelez:
Hey @JulianTillmann 

I did some testing (with azurite) and here are the results:
I used the following hierarchy:
```
|   rootblob1
|   rootblob2
|
\---folder1
    |   folderblob1
    |   folderblob2
    |
    +---subfolder1
    |   |   subfolderblob1
    |   |   subfolderblob2
    |   |
    |   \---subsubfolder
    |           subsubfolderfolderblob1
    |           subsubfolderfolderblob2
    |
    \---subfolder2
        \---subsubfolder
                subsubfolderfolderblob1
                subsubfolderfolderblob2
```
and created a sample explorer page.

Currently, this is how it looks (let's agree that is bit messy):
image


Applying you changes, it looks like the following image: only BLOBs in the result, `IsDirectory` doesn't seem to work.
image

I added some changes as well (haven't pushed them yet) and with my changes it looks like this:
image

Let me know what you think. It's a quite interesting case. Maybe @SimonOfHH can share an opinion as well.

SimonOfHH:
> It's a quite interesting case. Maybe @SimonOfHH can share an opinion as well.

I also had this problem in one of my early implementations of this and was unsatisfied with the result 😅 Your result looks good to me @mazhelez and the fix looks also simple enough.
I wonder how one would structure a test for this, though 🤔

JulianTillmann:
@mazhelez 
I'm pretty sure that my IsDirectory function worked but I cannot see what you changed, so I'm happy to go with your solution :-)

mazhelez:
@JulianTillmann just created a PR to your branch with my additions: https://github.com/JulianTillmann/ALAppExtensions/pull/1

If you approve with the changes, just merge it and I'll continue with the uptake.

JulianTillmann:
looks good

mazhelez (approved):

JesperSchulz:

Build ID: 47660.

JesperSchulz:

Build ID to track: 47679.

JulianTillmann:
> Availability update: We will publish a fix for this issue in the next update for release 21.
> 
> Build ID to track: 47679.

so this fix will not be applied to BC20?

JulianTillmann:
@mazhelez @JesperSchulz 
unfortunately this doesn't work in BC20 CU1. 

image

image

we also need a fix for BC20 as well

JulianTillmann:
> Issues addressed on GitHub will only get implemented in the latest released version - in this case 21.x. If we had to backport to all supported versions, it would be too time consuming.

so I would need to create a support case to get a fix for BC20? 

and it's not fixed for BC21 as I wrote 27 days ago

JesperSchulz:
To me it looks like this is fixed from 21.1 and newer? Try looking again - it really should be fixed. You can even see the changes here in the repo.
If you really need this for 20.x, I can backport the change for you as an exception. But usually, you should not expect changes coming in through GitHub to be backported. If you've got bugs of that nature, support cases would indeed be the way to go.
I'll do the backport to 20.x right away.

JulianTillmann:
> To me it looks like this is fixed from 21.1 and newer? Try looking again - it really should be fixed. You can even see the changes here in the repo. 

I just tried again with BC21 CU2. The content-type is "application/octet-stream" instead of "directory" or "Directory" 

image

JesperSchulz:
So this doesn't even work in 21.2? That's bad news. We need a new PR here on GitHub for that, as my team unfortunately doesn't have capacity to fix bugs in open source PRs.

mazhelez:
@JulianTillmann is there another issue other than the content type?

It's a weird one as there is a test that tests the content type for directories 🤔

mazhelez:
@JulianTillmann I did some testing on our side and I cannot repro the issue you experience.

image

This time I tested against an actual _Azure Blob Storage_ account (typically we run tests using azurite).
I tested with BLOBs created with `PutBlobBlockBlobText`. E.g.:
```
  BlobName := 'Folder1/Subfolder1/' + Format(CreateGuid());
  if ABSBlobClient.PutBlobBlockBlobText(BlobName, BlobName).IsSuccessful() then
      Message('Blob %1 created successfully', BlobName);
```

I also tested manually creating BLOBs from the Azure portal:
image

This is how it looks in the portal:
image

And this is how this looks in BC:
image


How are your BLOBs created? Can you at least try provide repro steps? Otherwise, there is very little we can do.

JulianTillmann:
@mazhelez 
that's extremely strange

I mostly created the folders manually in azure portal:
image

but I now also tried PutBlobBlockBlobText() and Upload a file from azure portal into a specific folder like you showed. but then the content-type is just empty for me:
image

I'm not sure on how to provide a better repro :-(  could we maybe do a teams session somehow sometime?

mazhelez:
That's indeed very weird 🤔 

I can't even see the "Add Directory" action on my container (and I am the owner of the subscription, so I should have god powers).

I just tried adding folders using the _Azure Storage Explorer_ and the result is as expected: the content type is "Directory" for all the folders.

The only thing that's different is that I'm running the latest version of the ABS module in _System Application_. I don't think anything has changes since the fix for the hierarchy, but I might be wrong. Any chance you can run the latest version?

Otherwise, I'd gladly help out with a Teams session, where I can let you test with my ABS and vice versa.

PR #20445, opened 2022-09-27
License information

paulojcdev:
We need a way of getting the license details stored on the License Information table while making our App universal code compliant.
This table is not accessible on SaaS, hence we propose this changes to be able to get license details for OnPrem environments.
It should be possible to reference this new codeunit on Apps where target = cloud, but it will only provide the details if the environment is OnPrem.

### DIFF
--- a/Modules/System Tests/Environment Information/src/EnvironmentInformationTest.Codeunit.al
+++ b/Modules/System Tests/Environment Information/src/EnvironmentInformationTest.Codeunit.al
@@ -12,6 +12,7 @@ codeunit 135091 "Environment Information Test"
         Assert: Codeunit "Library Assert";
         EnvironmentInformation: Codeunit "Environment Information";
         EnvironmentInfoTestLibrary: Codeunit "Environment Info Test Library";
+        LicenseInformation: Codeunit "License Information";
 
     [Test]
     [Scope('OnPrem')]
@@ -77,6 +78,40 @@ codeunit 135091 "Environment Information Test"
         // [When] Poll for IsSaaS
         // [Then] Should return false
         Assert.IsFalse(EnvironmentInformation.IsSaaS(), 'Testability should have dictacted a non- SaaS environment');
+    end;
+
+    [Test]
+    procedure TestGetLicenseErrorWhenTestabilitySaaSIsSet()
+    var
+        LicenseDetails: text;
+    begin
+        // [SCENARIO] Set the SaaS testability to true. GetLicenseDetails returns error, only for OnPrem.
+
+        // [Given] Environment is SaaS
+        EnvironmentInfoTestLibrary.SetTestabilitySoftwareAsAService(true);
+
+        // [When] Get license details
+        asserterror LicenseDetails := LicenseInformation.GetLicenseDetails();
+
+        // [Then] Expected Error
+        Assert.ExpectedError('Only for OnPrem environments. In SaaS environments use Entitlements.');
+    end;
+
+    [Test]
+    procedure TestGetLicenseDetails()
+    var
+        LicenseDetails: text;
+    begin
+        // [SCENARIO] Set the SaaS testability to false. GetLicense returns the license details.
+
+        // [Given] Environment is OnPrem
+        EnvironmentInfoTestLibrary.SetTestabilitySoftwareAsAService(false);
+
+        // [When] Get license details
+        LicenseDetails := LicenseInformation.GetLicenseDetails();
+
+        // [Then] Should return true
+        Assert.IsTrue(LicenseDetails <> '', 'License details should not be empty.');
     end;
 }
 
--- a/Modules/System/Environment Information/permissions/EnvironmentInfoObjects.PermissionSet.al
+++ b/Modules/System/Environment Information/permissions/EnvironmentInfoObjects.PermissionSet.al
@@ -12,5 +12,7 @@ permissionset 417 "Environment Info. - Objects"
     Permissions = Codeunit "Environment Information Impl." = X,
                   Codeunit "Environment Information" = X,
                   Codeunit "Tenant Information Impl." = X,
-                  Codeunit "Tenant Information" = X;
+                  Codeunit "Tenant Information" = X,
+                  codeunit "License Information Impl." = X,
+                  codeunit "License Information" = X;
 }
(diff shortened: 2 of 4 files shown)


paulojcdev:
@microsoft-github-policy-service agree [company="LS Retail"]

paulojcdev:
Hi,

is there something I can change/fix? I see two errors on the CI/CD.

Thanks.

JesperSchulz:
@paulojcdev, the build has now completed successfully internally, but before I merge it, I have one question: Now that you get the license information returned as one large string, what do you intend to do with it? Do you now parse that string, to extract some information out of it? Or do you simply display it in its entirety? I wonder if we could provide a more capable API, than returning everything as a string... 🤔

paulojcdev:
> @paulojcdev, the build has now completed successfully internally, but before I merge it, I have one question: Now that you get the license information returned as one large string, what do you intend to do with it? Do you now parse that string, to extract some information out of it? Or do you simply display it in its entirety? I wonder if we could provide a more capable API, than returning everything as a string... 🤔

Hi! Yes, the ideia was to parse it and fetch the relevant information for us. 
Probably, if we could instead return a json object or a json formatted text would be already better. 

Or we can look at creating procedures to get the different values individually? (but keep this too, to get the raw data)

Speaking of that, it would be amazing if we could also get the information if a license is "Dual Rights".

Thanks for looking into this.

SBalslev:
@paulojcdev, I'd like to understand more about the information that is relevant from the license information table that cannot be derrived from other logic.
As soon as you use data from this table your solution is no longer cloud compatible and cannot be moved to a cloud environment. That means it is no longer tageting cloud.
So I'd like to talk about the functionality/data that is missing so your solution target cloud, regardless of where it is deployed.

paulojcdev:
> @paulojcdev, I'd like to understand more about the information that is relevant from the license information table that cannot be derrived from other logic. As soon as you use data from this table your solution is no longer cloud compatible and cannot be moved to a cloud environment. That means it is no longer tageting cloud. So I'd like to talk about the functionality/data that is missing so your solution can target cloud, regardless of where it is deployed.

Hi,

the information we need/use is:
- Voice Account Number
- Amount of 
	- Business Central Device
	- Business Central Team Members
	- etc
- Dual Rights

We only use this information if the app is installed on-prem, for SaaS installations we will use the Tenant Id and entitlements (Dual rights does not apply).

Hope this help.

SBalslev:
What's the scenario where you need to know such data? And not just see how many users/devices etc are registered in the other tables. Is it for an additional/custom licensing story on top of the license file?

paulojcdev:
Hi,

yes, we need this information to do our won licensing checking for our modules.

thanks,

ti-jalopez:
> Probably, if we could instead return a json object or a json formatted text would be already better.

@JesperSchulz, @SBalslev ,
Maybe to use as a return value of  type **List of [Text]** simplifies the process to decode the content of text license lines.
A process than nowadays iterate over lines of table of "License Information" table is very similar to iterate over the list of lines of license.

What do you think about?

SBalslev:
The challenge I see is that this would only work for on premises, so it would not be a cloud compatible solution.
If we are to make a cloud compatible solution, we should agree on a contract that can be used on-premises as well as in a cloud deployment. 
It could be tenant id (account no), number of seats pr. license, expiry and perhaps other things. Naturally it is going to be difficult to align subscriptions to license file but if we're just going to give raw access to the data in the table without making it cloud compatible then there is really not point in having the target=onprem on the table.

paulojcdev:
> The challenge I see is that this would only work for on premises, so it would not be a cloud compatible solution. If we are to make a cloud compatible solution, we should agree on a contract that can be used on-premises as well as in a cloud deployment. It could be tenant id (account no), number of seats pr. license, expiry and perhaps other things. Naturally it is going to be difficult to align subscriptions to license file but if we're just going to give raw access to the data in the table without making it cloud compatible then there is really not point in having the target=onprem on the table.

This seems like a good idea.

Our app is already checking on what environment type we are (On-prem or SaaS) and depending on it, will access and use the correct information.
But with the goal of making the app Universal Code compliant we need to have a way of still getting the same information that we currently get from the License Information table, when On-prem.
We are open to any kind of solution, we just want to be able to make our app Universal Code compliant without losing this feature.
For us the most important is to be able to get the voice account number. We will use this value for On-prem installations only. For SaaS we will instead use the tenant id which we already get from the existing TenantId() function.

ti-jalopez:
> The challenge I see is that this would only work for on premises, so it would not be a cloud compatible solution.
> If we are to make a cloud compatible solution, we should agree on a contract that can be used on-premises as well as in a cloud deployment.

The origin of switching to target cloud is that an application that is installed on OnPrem meets the "Universal Code" requirement.
In this context, they should be allowed to access all the tables of the database either OnLine or OnPrem (previously checking the environment with IsSaas()), since there is no technical limitation that prevents access to the database.

JesperSchulz:
We just discussed this case internally and reached the agreement, that the PR in its current shape cannot be approved. The idea behind "universal code", is that you can take your on-prem app, and install it in a cloud environment without modifications. If we were to allow access to the raw license data, that would not work, as the license doesn't exist in the shape it does on-prem.

So, the suggestion is to go back to the drawing board, and to change this PR to return the values you need. The code should work both on-prem AND cloud. E.g., if you are on-prem, the license account no. would return the "Voice Account Number" from the license, while it would return e.g. the tenant ID, when you are online.

Would you be willing to re-engineer this PR to live up to these guidelines? The platform should supply everything you need to create such system application module. What do you say?

JesperSchulz (changes requested):
PR cannot be approved in its current shape, as per comment in the discussion. Solutions for target "cloud" must work both on-prem as well as in the cloud - that's the idea behind universal code (it's universal ;-)). This solution would break in the cloud. We would set ourselves on a wrong path.

ti-jalopez:
> We just discussed this case internally and reached the agreement, that the PR in its current shape cannot be approved. The idea behind "universal code", is that you can take your on-prem app, and install it in a cloud environment without modifications. If we were to allow access to the raw license data, that would not work, as the license doesn't exist in the shape it does on-prem.
> Would you be willing to re-engineer this PR to live up to these guidelines? The platform should supply everything you need to create such system application module. What do you say?

I'm thinking about how to agree with "Universal Code":

- **Why not create the virtual "Licence Permissions" table in Cloud platform -without data- and remove scope=OnPrem from this table?**
- Currently there are other tables like "License Permissions" that contain OnPrem license data (perpetual or pay per use), and they don't have scope=OnPrem. This table can be used with target Cloud or target OnPrem without any problems. Why not do the same for the "License Information" table?

JesperSchulz:
We'll do as suggested :-) "License Information" will from 21.3 be target "cloud". On prem, the table contains data, in the cloud, it is empty. It's not the perfect solution, but it will do.
Thank you for your contribution though! I was a good discussion.

If you're not happy with the solution, please re-open this PR.

PR #20510, opened 2022-09-30
Delete a range of values from temp blob list

adrogin:
Currently, there is no way to efficiently delete a range from a Temp Blob List. If I need to clear a list, I can call RemoveAt in a loop, but this method copies a bulk of elements on each iteration.
This pull request introduces the RemoveRange operation, identical to the List data type, to remove a range in a single operation.


Modules/System/BLOB Storage/src/TempBlobList.Codeunit.al
```diff
@@ -67,6 +67,18 @@ codeunit 4102 "Temp Blob List"
         exit(TempBlobListImpl.RemoveAt(Index));
     end;
 
+    /// <summary>
+    /// Remove a number of elements starting from a specified location from a non-empty list.
+    /// </summary>
+    /// <error>The index is larger than the number of elements in the list or less than one.</error>
+    /// <param name="Index">The index of the first element to be deleted.</param>
+    /// <param name="ElementCound">The number of elements to be deleted.</param>
```
mazhelez, line 75: Typo

```suggestion
    /// <param name="ElementCount">The number of elements to be deleted.</param>
```

Modules/System/BLOB Storage/src/TempBlobListImpl.Codeunit.al
```diff
@@ -44,13 +44,20 @@ codeunit 4109 "Temp Blob List Impl."
...
     var
         TempBlobList: Codeunit "Temp Blob List";
     begin
         if not TempBlobRec.Get(Index) then
             Error(ObjectDoesNotExistErr, Index);
-        if TempBlobRec.Get(Index + 1) then
-            GetRange(Index + 1, Count() - Index, TempBlobList);
+
+        if TempBlobRec.Get(Index + ElementCount) then
+            GetRange(Index + ElementCount, Count() - Index - ElementCount, TempBlobList);
```
mazhelez, line 59: This PR about is parametrizing the `1`. 

However, `Count() - Index` became `Count() - Index - ElementCount`. Why is that? Was this an issue before?

Modules/System/BLOB Storage/src/TempBlobListImpl.Codeunit.al
```diff
@@ -8,9 +8,9 @@ codeunit 4109 "Temp Blob List Impl."
     Access = Internal;
 
     var
-    #pragma warning disable AA0073
+#pragma warning disable AA0073
```
mazhelez, line 11: Formatting is weird here

Modules/System/BLOB Storage/src/TempBlobList.Codeunit.al
mazhelez: Also, what if negative value is passed? Is this supported?

mazhelez (commented):
Minor comments to address. But overall: great contribution. 👍

Modules/System/BLOB Storage/src/TempBlobList.Codeunit.al
adrogin: Thanks for the comment. Added an explicit error and test cases for a negative count.

Modules/System/BLOB Storage/src/TempBlobListImpl.Codeunit.al
adrogin: That's a bug, indeed. Existing tests did not cover this verification. Fixed and added tests.

Modules/System/BLOB Storage/src/TempBlobListImpl.Codeunit.al
adrogin: That's AL Language extension autoformatting. And as far as I can see, the pragma instruction is always aligned on the left without indentation.

pri-kise: That formatting was reported on the AL Language Repository, but it wasn't solved yet.
See Issue https://github.com/microsoft/AL/issues/6530

JesperSchulz (approved):

JesperSchulz:

Build ID: 49407.

JesperSchulz:

Build ID to track: 49535.

PR #20592, opened 2022-10-11
Add "Locked = true" to HttpHeader Labels

MattiLE:
In the module Azure Blob Service Api it is possible to set HTTP Header via the predefined codeunit "ABS Optional Parameters". When using the "Range" function, a label is used that does not have locked = true. This means that the translation is always used when setting the HTTP header. The problem is that in some XLF files the translation is different from the caption, so the HTTP request runs into an error. (e.g. in the German translation)


mazhelez (approved):
Nice catch!
Thanks for the fix :)

JesperSchulz (approved):

JesperSchulz:

Build ID: 48783.

PR #20751, opened 2022-10-25
GP - PO Non-Inventory missing UofM fix

jaymckinney-enavate:
This change fixes an issue where the GP migration would fail if an Open PO contains a non-inventory item that has a unit of measure not set up in GP (added via free text on the PO line).

- Moved the PO migration code out of the tables and to a new Codeunit.
- It creates the Unit of Measure if it doesn't exist.


nikolakukrika (approved):

JesperSchulz:

Build ID: 48508.

JesperSchulz:

Build ID to track: 48686.

PR #20979, opened 2022-11-12
[DRAFT] [New Module] Background Task Framework

StefanMaron:
# Background Task Framework

## Overview

- low entry barrier for developers
- centralized log/overview list for users
- in System App (contrary to "Job Queue" in Base App)
- additional developer tooling to handle background tasks

## Misc
Ask for Opinion:
- MS
- Mads Grams?
- Vjeko? 
- Waldo?
- Tobias Fenster?
- Henrik Helgesen?

## Description
The idea is to create a framework that wrapps the task scheduler, to improve the way developers can run code in the background. 

### Story
I think the majority of AL developers writes procedual code either all the time or at least most of the time. Reasons for that may be individual, but I think most of the time its just due to old habbits. The hurdle to run some task in the background is just too high to just try it or get started.

And yet it brings so many benefits to run a task in the background. And this is not about splitting a task into multiple smaller tasks to run them in background and in parallel, its just about running one task in the background instead of the UI session of the user.

1. The user session will be free immediately. This improves the perceived performance A LOT!
2. The result is, that the user can start with his next task right away and gets more productive
3. The service of BC can do much better load balancing when there are more sessions. Sessions get distributed between mutliple NSTs. If there is just one session, nothing can get balanced.
4. With the implementation of the Await() function, the developer can get an overall performance increase if the code.
5. Once we have the framework in place, more helper functions can be added easily. Maybe a CallBack function would make sence? Easy implemented.

So why the heck is this not used more among AL developers?

### Problem 1:
Many AL developers dont touch this area too much because its a huge unknown. Not really sure how it behaves, not really sure how to implement it the right way. Especially when it comes to error handling. Most developers dont have the time to invest into learning such a technology in their daily work routine and can not afford to build a buggy solution.
So better stick with the traditional way of developing.

### Solution 1:
The task framework should solve that problem by introducing a wrapper for the task scheduler. The framework will intoduce some helper functions:

- Await
- Callback?
- Parameters for Background Tasks

And along with that some examples how to implement them correctly.
Additionally, the framework will implement default error handling and logging. This will give the certainty to the developer that there is proper error handling without the need to come up with a solution themselfes.

### Problem 2:
When looking at BC from a users perpective, there are not many tasks that run background currently and in the past. Only the job queue but thats rather recurring tasks, and run a report in background. Thats it. 

That means, if we (developers) suddenly start to run tasks in background, the users will be confused. The UI session does not "hang" so the task must be finished. But there was no message or confirmation. No feedback at all - very confusing.

Without any aditional development, when using the task scheduler, the user does not get any information that the action he clicked, did start, and is not yet finished. And there is no place to check the current status of tasks and to check if a task already did finish.
Also any error handling would need to be implemented manually, and therefore would differ from solution to solution.

### Solution 2: 
The framework should include a log table where all the tasks are logged to. The user can open this log to check currently running tasks, tasked planned for future execution, already finished tasks, and task that failed and view the error details.

It also needs to implement notifications to the user, that a task has been scheduled for background execution. The notification should include a link to open up a status/log list to check and view the current status of this task.

## Job queue
When asking various people for feedback, the no 1 question I got was "But why not just use the Job Queue". Most of all, I feel like, this kind of module should not be part of the base app, its a clear candiate for the system app. But then also, the job queue module has many featues and is mainly used for reocurring tasks. 
The main purpose of this framework should be to make it easier to just dispatch tasks to the background.

## References
https://twitter.com/StefanMaron/status/1529038705641631745
https://thedoubleh.dev/posts/performancevsexperience/#real-world-issue

### DIFF
--- a/Modules/System/Background Task Framework/app.json
+++ b/Modules/System/Background Task Framework/app.json
@@ -1,0 +1,22 @@
+{
+    "id": "b53bcdb4-4264-44ac-9dc0-a98fc652b128",
+    "name": "Background Task Framework",
+    "publisher": "Microsoft",
+    "brief": "Background Task Framework provides functionality to easily dispatch functionality to get executed in the background.",
+    "description": "Background Task Framework provides functionality to easily dispatch functionality to get executed in the background.",
+    "version": "21.1.0.0",
+    "privacyStatement": "https://go.microsoft.com/fwlink/?linkid=724009",
+    "EULA": "https://go.microsoft.com/fwlink/?linkid=2009120",
+    "help": "https://go.microsoft.com/fwlink/?linkid=2103698",
+    "url": "https://go.microsoft.com/fwlink/?linkid=724011",
+    "contextSensitiveHelpUrl": "https://docs.microsoft.com/dynamics365/business-central/",
+    "platform": "21.0.0.0",
+    "dependencies": [],
+    "target": "OnPrem",
+    "idRanges": [
+        {
+            "from": 9203,
+            "to": 9224
+        }
+    ]
+}


StefanMaron:
This is a first draft. I needed to create the app.json already in order to be able to create the PR.
With me working on this is @SimonOfHH and Kristof Klein.

We tried to give a decent description to get some feedback and to start a discussion first, before we get into coding.

So let us know what you think!

mazhelez:
@darjoo should really be in these discussions 😎

pri-kise:
Right now we are having a similiar structure to schedule processes, that should be done, after a posting of documents.
This could be refactored to a decent system module.

Therefore it is really convient to add a some delay to this tasks to avoid locking issues.
And some kind of automatic rescheduling depending on specific conditions, e.g. reschedule if error was caused by a locking issue.

StefanMaron:
> Therefore it is really convient to add a some delay to this tasks to avoid locking issues. And some kind of automatic rescheduling depending on specific conditions, e.g. reschedule if error was caused by a locking issue.

I believe this is even done automatically by the task scheduler
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-task-scheduler#retriable-exceptions-in-the-main-codeunit

But it seems like the docs do not really specify which error types are acutally covered.

BernhardKloibmueller:
Hello,

I am not sure if feedback from partners are wanted but since this open to all I will try to add productive feedback.

I am not sure if it is goo to add more strain on the scheduled task table. I know of circumstances where the scheduled task table is locked and not accessible. 

And I would have a question where you would want to use this kind of module? What do you want to achieve?
Do you wish something like the page backgroundtasks for overall logon? But you speak of freeing user sessions and putting work to background, so when the user logs of, then the execution should continue, am I right?

Do you want to make independent sessions (this is how the task scheduler works) or do you want to control the sessions just like with start session? I can tell you that I have once created a parallel tasks tool also working over multiple companies so it is possible with the tools we have at hand. Unfortunately that one never got any production phase.
What technical basis you had in mind?

But we have some technical hindrances:
 - If Task Scheduler: if all sessions are in use then the execution will wait, could take much time on heavy use databases
 - if StartSession; there is a maximum number of child sessions setup in the server instance, I don't know the value in saas though
 - Unfortunately I know that the Scheduled Task table can be put at much strain so that it could be inaccessible for an update (I can share a support request ID for the people of Microsoft if wanted), so adding more ways to increase the usage can be troublesome from my point of view
 - action timeout: a few versions back we got the timeout parameters at reports and the job queue, these should also apply to these tasks, how high do you would put them?
 - How would you define dependencies between tasks? adding a background tasks which ads again a background tasks or even multiple ones which should start in a specified order?

Maybe if there is any usage in our daily business you could explain, it would be better for me to follow your argumentation.

Do you want to read Data parallel?
Do you want to execute tasks parallel? 
Do you want to accessing external systems though web services?
Do you want to manipulate data parallel?

Please don't misunderstand my comment that I would be against it, but I don't see the benefit to the tools at hand at the moment. I am every time the first one looking for new and easiers ways to develop.

Best regards

vhn:
IMO the trouble with all these older asynchronous mechanisms in BC like using task scheduler, job queue or background sessions via StartSession() is that they cannot run any code back on the user thread with their results when they are done.
That is what separates AL from any language with continuations, callbacks, async+await approaches.

But here's the thing: BC finally got the "Page background task" feature which already has the exact API needed here.
It can pass some data to a child session, do some work, send data back to user session when done and run some more code there. It is the first real normal async. API in the platform/language that I am aware of. 

The only problem with the feature is the limitations placed upon it:
It is linked to pages when it could instead have been a "Background task" feature available for codeunits as well.
It blocks writing to DB inside child session.

I don't see why those two limitations couldn't be removed, finally giving AL a proper callback based asynchronous mechanism to speed up any business procedure, either on NAS or on user session, which is what you are after in this PR.

JesperSchulz:
@darjoo, do you think you can find the time to engage in this discussion after 2023 Wave 1 has been sent out the door? This is a highly complex, yet highly interesting area to start a collaboration on! Would you think we're ready to handle that? What would the path forward be?

JesperSchulz:
@StefanMaron, we will soon be retireing this repo for System Application contributions. The new repository will be https://github.com/microsoft/BCApps. As this PRs doesn't have any code, there isn't much to migrate or process. Could you reopen this discussion somewhere else? It doesn't really qualify as a pull request 😉 Thank you 🙏

PR #21117, opened 2022-11-22
Deliver packages with artifacts and source code to nuget.org

mazhelez:
Define custom deliver script to pack the source code in the package as well.

- During app compilation, place the _generated app file_ and the _source code_ (in a `SourceCode` folder) for the current app in the build artifacts folder under folder `Package/<app name>` (to mark that it will be included in the package).
The source code is needed right after compilation as this is when the LCG files are generated (they are not part of the source code in the .app file).
- When creating the package to deliver, place all app folders under `Apps` in the package itself.
E.g.:
```
\---Apps
    +---Any
    |   |   Microsoft_Any_21.3.272.0.app
    |   |
    |   \---SourceCode
    |       |   app.json
    |       |   README.md
    |       |
    |       \---src
    |               Any.Codeunit.al
    |
    +---Library Assert
    |   |   Microsoft_Library Assert_21.3.272.0.app
    |   |
    |   \---SourceCode
    |       |   app.json
    |       |   README.md
    |       |
    |       \---src
    |               LibraryAssert.Codeunit.al
```


.github/workflows/CICD.yaml
```diff
@@ -637,7 +634,7 @@ jobs:
 
   Deliver:
     needs: [ Initialization, Build ]
-    if: always() && needs.Build.result == 'Success' && github.event_name != 'workflow_run' && github.ref_name == 'main' && needs.Initialization.outputs.deliveryTargetCount > 0
+    if: always() && needs.Build.result == 'Success' && github.event_name != 'workflow_run' && needs.Initialization.outputs.deliveryTargetCount > 0
```
mazhelez, line 637: Revert! :)

.github/DeliverToNuGet.ps1
```diff
@@ -0,0 +1,137 @@
...
+    Copy-Item -Path (Join-Path $env:GITHUB_WORKSPACE $project) -Destination (Join-Path $packageFolder 'SourceCode') -Recurse -Force
+    
+    #Create .nuspec file(Join-Path $packageFolder 'manifest.nuspec')
+    $nuspecFilePath = (Join-Path $packageFolder 'manifest.nuspec')
+    $nuspec.Save($nuspecFilePath)
+    
+    $outputDirectory = Join-Path $env:GITHUB_WORKSPACE 'out'
+    New-Item -Path $outputDirectory -ItemType Directory -Force | Out-Null
+
+
```
mazhelez, line 107:
```suggestion
```

.github/DeliverToNuGet.ps1
```diff
@@ -0,0 +1,137 @@
...
+function GenerateNuspec
+(
+    [Parameter(Mandatory=$true)]
+    $PackageId,
+    [Parameter(Mandatory=$true)]
+    $Version
+)
+{
+    [xml] $template = '<?xml version="1.0" encoding="utf-8"?>
+    <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
```
mazhelez, line 14: Separate file maybe?

.github/DeliverToNuGet.ps1
```diff
@@ -0,0 +1,137 @@
...
+    {
+        throw "Generating the nuget pack failed with exit code $LASTEXITCODE"
+    }
+}
+finally {
+    Remove-Item $packageFolder -Recurse -Force
+}
+
+# Get the newly created package
+$nugetPackageFile = gci -Path $outputDirectory -Filter "$($packageParameters.packageId)*.nupkg"
```
mazhelez, line 124:
```suggestion
$nugetPackageFile = gci -Path $outputDirectory -Filter "$packageId*.nupkg"
```

.github/DeliverToNuGet.ps1
```diff
@@ -0,0 +1,137 @@
...
+
+# Get the newly created package
+$nugetPackageFile = gci -Path $outputDirectory -Filter "$packageId*.nupkg"
+
+if(-not $nugetPackageFile) {
+    throw "Cannot find nupkg file"
+}
+
+Write-Host "Push package $($nugetPackageFile.FullName) to $nuGetServerUrl" -ForegroundColor Magenta
+$nugetOutput =  Invoke-Expression -Command "./nuget.exe push $($nugetPackageFile.FullName) -ApiKey $nuGetToken -Source $nuGetServerUrl"
```
mazhelez, line 131:
```suggestion
$nugetOutput =  Invoke-Expression -Command "$outputDirectory/nuget.exe push $($nugetPackageFile.FullName) -ApiKey $nuGetToken -Source $nuGetServerUrl"
```

.github/DeliverToNuGet.ps1
```diff
@@ -0,0 +1,156 @@
...
+    [Parameter(Mandatory=$true)]
+    $Owners
+)
+{
+    [xml] $template = '<?xml version="1.0" encoding="utf-8"?>
+    <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+        <metadata>
+            <id></id>
+            <version></version>
+            <title>D365 Business Central — System Modules</title>
```
mazhelez, line 22: Just shouldn't be hardcoded to System Modules, but for now it is only that module.

Configuring the nuget package needs to be handled per project.

.github/workflows/CICD.yaml
```diff
@@ -461,23 +461,20 @@ jobs:
 
       - name: Publish artifacts - apps
         uses: actions/upload-artifact@v3
-        if: github.ref_name == 'main' || startswith(github.ref_name, 'release/')
```
mazhelez, line 464: Revert 🗡️

.github/DeliverToNuGet.ps1
```diff
@@ -0,0 +1,124 @@
...
+    $testAppsPackage = Join-Path $testAppsFolder 'Package'
+    if(Test-Path -Path "$testAppsPackage") {
+        Copy-Item -Path "$testAppsPackage" -Destination "$packageFolder/Tests/" -Recurse -Container -Force
+    }
+    
+    #Create .nuspec file
+    $nuspecFilePath = (Join-Path $packageFolder 'manifest.nuspec')
+    $nuspec.Save($nuspecFilePath)
+
+    Write-Host "Download nuget CLI" -ForegroundColor Magenta
```
jonasblunck, line 96: Should we move the download of Nuget.exe into a separate action instead, and reference that one? We might need it in more places?

Modules/.AL-Go/CompileAppInBcContainer.ps1
```diff
@@ -2,40 +2,43 @@ Param(
...
+# Extract app name from app.json
+$appName = (gci -Path $appProjectFolder -Filter "app.json" | Get-Content | ConvertFrom-Json).name
+
+Write-Host "Current app name: $appName; app folder: $appProjectFolder"
+
+# Create an archive with the current source code in the build artifacts folder
+
+$archiveFile = "$env:TEMP/$appName.Source.zip"
+Write-Host "Archive the current source code for app: $appName as $archiveFile"
+Compress-Archive -Path "$appProjectFolder/**" -DestinationPath $archiveFile -Force
```
elbardel, line 24: Have you considered storing the source unzipped?
- The nupkg format is already zipped, so zip files inside do not save any space.
- It would be simpler to consume.

elbardel (approved):

Modules/.AL-Go/CompileAppInBcContainer.ps1
mazhelez: This step is to transfer the source code to GitHub artifacts so that it can be used in the _Deliver_ task. So, it's more about saving space there and not have to deal with files and folders but with a simple .zip.

We'll re-think which approach suits us best.

.github/DeliverToNuGet.ps1
mazhelez: This is the part where AL-Go is a bit limited.
YAML files are only to be edited from AL-Go itself.

I see this as an obstacle, because as much as we try, we might not be able to have "one solution fits them all" scenario.

Modules/.AL-Go/CompileAppInBcContainer.ps1
```diff
@@ -2,40 +2,37 @@ Param(
...
-                }
-                Copy-Item -Path $appPath -Destination (Join-Path $packagesFolder $name)
-                Copy-Item -Path $appPath -Destination (Join-Path "c:\run\my" $name)
-            }
-
-            CopyApp -localAppPath "C:\Applications.*\Microsoft_Base Application_*.*.*.*.app"    -w1AppPath "C:\Applications\BaseApp\Source\Microsoft_Base Application.app"
-            CopyApp -localAppPath "C:\Applications.*\Microsoft_Application_*.*.*.*.app"         -w1AppPath "C:\Applications\Application\Source\Microsoft_Application.app" 
-        } -argumentList (Get-BcContainerPath -ContainerName $parameters.ContainerName -path $Parameters.appSymbolsFolder)
-    }
+$appProjectFolder = $parameters.appProjectFolder
```
mazhelez, line 13: This part needs to be executed only on certain branches.

.github/workflows/CICD.yaml
```diff
@@ -461,23 +461,20 @@ jobs:
...
         with:
           name: ${{ env.appsArtifactsName }}
           path: '${{ matrix.project }}/.buildartifacts/Apps/'
           if-no-files-found: ignore
 
       - name: Publish artifacts - dependencies
         uses: actions/upload-artifact@v3
-        if: github.ref_name == 'main' || startswith(github.ref_name, 'release/')
```
mazhelez, line 472: Needs to be reverted :)

Modules/.AL-Go/CompileAppInBcContainer.ps1
```diff
@@ -2,40 +2,42 @@ Param(
...
+if(($branchName -eq 'main') -or $branchName.StartsWith('release/')) {
+    $appProjectFolder = $parameters.appProjectFolder
+
+    # Extract app name from app.json
+    $appName = (gci -Path $appProjectFolder -Filter "app.json" | Get-Content | ConvertFrom-Json).name
+
+    Write-Host "Current app name: $appName; app folder: $appProjectFolder"
+
+    # TODO Use Al-Go project settings to determine if the app is a test app or not.
+    $holderFolder = 'TestApps'
```
freddydk, line 25: Just had a look - there is actually a boolean variable called $app, which is true if this is an app.
There is a variable called $testApp, which is true is this is a test App.

If we make major restructuring of Run-AlPipeline, then this might change, but we have absolutely no plans of doing this.

Modules/.AL-Go/CompileAppInBcContainer.ps1
```diff
@@ -2,40 +2,42 @@ Param(
     [Hashtable] $parameters
 )
 
+# Setup compiler features to generate captions and LCGs
 if (!$parameters.ContainsKey("Features")) {
     $parameters["Features"] = @()
 }
-$parameters["Features"] = @("lcgtranslationfile", "generateCaptions")
+$parameters["Features"] += @("lcgtranslationfile", "generateCaptions")
```
freddydk, line 9: Only if $app ?

Modules/.AL-Go/CompileAppInBcContainer.ps1
```diff
@@ -2,40 +2,42 @@ Param(
...
-                }
-                Copy-Item -Path $appPath -Destination (Join-Path $packagesFolder $name)
-                Copy-Item -Path $appPath -Destination (Join-Path "c:\run\my" $name)
-            }
-
-            CopyApp -localAppPath "C:\Applications.*\Microsoft_Base Application_*.*.*.*.app"    -w1AppPath "C:\Applications\BaseApp\Source\Microsoft_Base Application.app"
-            CopyApp -localAppPath "C:\Applications.*\Microsoft_Application_*.*.*.*.app"         -w1AppPath "C:\Applications\Application\Source\Microsoft_Application.app" 
-        } -argumentList (Get-BcContainerPath -ContainerName $parameters.ContainerName -path $Parameters.appSymbolsFolder)
+$branchName = $ENV:GITHUB_REF_NAME
+
```
freddydk, line 14: You should include this code here:

$workflowName = "$env:GITHUB_WORKFLOW".Trim()
And then check that workflowName -eq "CI/CD"

Else the code will also trigger on other workflows.

freddydk: Maybe also check $ENV:GITHUB_REPOSITORY is microsoft/ALAppExtensions (to ensure the code isn't run on forks)

mazhelez: I'd imagine if no deliver is needed in the fork, there won't be a delivery context.

freddydk: You are right on that - unless you have the secret in a shared KeyVault - but probably a non-problem

freddydk:
I think you should use DeliverToMSFT instead of DeliverToNuGet (and rename the secret to MSFTContext).

DeliverToNuGet will become a standard AL-Go construct and might confuse people that you use the same, but change the functionality (and the package).

freddydk:
Did we find any solution on the test runner?

mazhelez:
> Did we find any solution on the test runner?

@freddydk what is this about? I am missing some context here.

mazhelez:
> I think you should use DeliverToMSFT instead of DeliverToNuGet (and rename the secret to MSFTContext).
> 
> DeliverToNuGet will become a standard AL-Go construct and might confuse people that you use the same, but change the functionality (and the package).

Changed to `DeliverToBCFeed`.

.github/DeliverToBCFeed.ps1
```diff
@@ -0,0 +1,135 @@
...
+    $packageId += "-$projectName"
+}
+
+if ($type -eq 'CD') 
+{
+    $packageId += "-preview"
+}
+
+# Extract version from the published folders (naming convention)
+$packageVersion = $appsFolder.SubString($appsFolder.LastIndexOf("-Apps-") + "-Apps-".Length) #version is right after '-Apps-'
```
elbardel, line 70: It might be a matter of taste, but I find regexes more readable than this.
This would be
``` $packageVersion = $appsFolder -replace '.*-Apps-','' #version is ... ```

File: .github/ALAppExtensions.nuspec [hunk 16]
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id></id>
```
elbardel, line 4: Consider making it clearer that this file is a template and not the actual nuspec. I can see there is a script to fill the missing fields, but looking only at this one is a bit confusing.

Maybe call it ALAppExtensions.template.nuspec? Or add a comment at the top?

elbardel (approved):

.github/ALAppExtensions.nuspec
mazhelez: Renamed to ALAppExtensions.template.nuspec

.github/DeliverToBCFeed.ps1
mazhelez: For me, it is also clearer using the regex :) Changed it. Thanks.

Modules/.AL-Go/CompileAppInBcContainer.ps1
mazhelez: Done

aholstrup1 (approved):

PR #21123, opened 2022-11-22
Updated AL-Go System Files

mazhelez:
No release notes available!

### DIFF
--- a/.github/workflows/NextMajor.yaml
+++ b/.github/workflows/NextMajor.yaml
@@ -15,7 +15,7 @@ env:
 
 jobs:
   Initialization:
-    runs-on: [ self-hosted,1ES.Pool=ALAppExt ]
+    runs-on: [ windows-latest ]
     outputs:
       telemetryScopeJson: ${{ steps.init.outputs.telemetryScopeJson }}
       settings: ${{ steps.ReadSettings.outputs.SettingsJson }}
@@ -209,7 +209,7 @@ jobs:
 
   PostProcess:
     if: always()
-    runs-on: [ self-hosted,1ES.Pool=ALAppExt ]
+    runs-on: [ windows-latest ]
     needs: [ Initialization, Build ]
     steps:
       - name: Checkout
--- a/.github/workflows/NextMajor.yaml
+++ b/.github/workflows/NextMajor.yaml
@@ -10,11 +10,12 @@ defaults:
   run:
     shell: PowerShell
 
+env:
+  workflowDepth: 1
+
 jobs:
   Initialization:
     runs-on: [ self-hosted,1ES.Pool=ALAppExt ]
-    env:
-      workflowDepth: 1
     outputs:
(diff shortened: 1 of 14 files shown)


.github/workflows/NextMinor.yaml
```diff
@@ -125,6 +127,24 @@ jobs:
           settingsJson: ${{ env.Settings }}
           SecretsJson: ${{ env.RepoSecrets }}
 
+      - name: Upload thisbuild artifacts - apps
```
ankosiorek, line 130:
```suggestion
      - name: Upload this build artifacts - apps
```

.github/workflows/PullRequestHandler.yaml
```diff
@@ -41,11 +41,13 @@ jobs:
           Add-Content -Path $env:GITHUB_OUTPUT -Value "prfolder=$prfolder"
           $prPath = Join-Path $location $prFolder
           New-Item -Path $prPath -ItemType Directory | Out-Null
+          $prfilesChanged = @()
```
ankosiorek, line 44:
```suggestion
          $prFilesChanged = @()
```

.github/workflows/NextMajor.yaml
```diff
@@ -125,6 +127,24 @@ jobs:
           settingsJson: ${{ env.Settings }}
           SecretsJson: ${{ env.RepoSecrets }}
 
+      - name: Upload thisbuild artifacts - apps
```
ankosiorek, line 130:
```suggestion
      - name: Upload build artifacts - apps
```

.github/workflows/NextMinor.yaml
ankosiorek: Is that some kind of convention I don't know: _thisbuild_?

mazhelez: Yeah, this comes from AL-Go. It was updated in [this PR](https://github.com/microsoft/AL-Go/pull/297/files#diff-01f4ef19a96cdb7663a0199b5880386ffc66ae75cf8696fbc655bcc2eb5cc316).

.github/workflows/PullRequestHandler.yaml
mazhelez: Again, this needs to be fixed in AL-Go itself :)

freddydk (approved):

.github/workflows/NextMinor.yaml
freddydk: The thisbuild artifacts are artifacts from this build, which only is used for this build to carry over to the next project (when running multiprojects with useProjectDependencies = true) - a fix for this issue: https://github.com/microsoft/AL-Go/issues/292

.github/workflows/PullRequestHandler.yaml
freddydk: Yeah - we can fix this in AL-Go

PR #21172, opened 2022-11-23
Update Listing-translation.en-us.json

AndreiPanko:


JesperSchulz:
Processed internally already. No need to merge. Will get updated with next push.

PR #21203, opened 2022-11-25
Fix SharePoint AcquireToken

ShareTEC:
- IsSuccess never received a value from OAuth2.AcquireTokenByAuthorizationCode.
- Changed "Prompt Interaction"::Login to "Prompt Interaction"::None. Otherwise, you always need to login again. Now, you can select an account already signed in.


JesperSchulz:
Processing internally.

JesperSchulz:

Build ID to track: 50796.

JesperSchulz:

Build ID: 50848.

PR #21206, opened 2022-11-27
Delete record links from a recordset

adrogin:
Adding a function RemoveLinks in the codeunit Record Link Management to delete record links from all records within a filtered recordset. This is a helper function to fix the issue 138 in BaseApp.
https://github.com/microsoft/BusinessCentralApps/issues/138


TKapitan (approved):

Modules/System Tests/Record Link Management/src/RecordLinkMgtTest.Codeunit.al
```diff
@@ -85,10 +85,7 @@ codeunit 132508 "Record Link Mgt. Test"
 
         // [GIVEN] A new record is created to set record links on
         FromRecordLinkRecordTest.DeleteAll();
-        FromRecordLinkRecordTest.Init();
-        FromRecordLinkRecordTest.PK := 1;
-        FromRecordLinkRecordTest.Field := 'Rec A';
-        FromRecordLinkRecordTest.Insert();
+        CreateRecordLinkRecTest(FromRecordLinkRecordTest, 1, 'Rec A');
```
grobyns, line 88: I dislike passing constants to helper methods in test code. While this case seems benign, it leads to paremeterized tests (one-liners) where it because very hard to know what is actually being tested. 
In cases like this, what I do is create a well named additional helper in which the constants are passed to the 'general' helper. Something like:
CreateFirstRecordLink(FromRecordLinkRecordTest);

and the second hardcode one would then obviously be:
CreateSecondRecordLink(FromRecordLinkRecordTest);

Another alternative to this is to create variables like RecordLinkPrimaryKeyValue and RecordLinkFieldFieldValue, assign values to those and pass the variables but that would kind of defeat the purpose of your helper here.

Modules/System/Record Link Management/src/RecordLinkImpl.Codeunit.al
```diff
@@ -65,6 +65,18 @@ codeunit 4470 "Record Link Impl."
         Note := BinReader.ReadString();
     end;
 
+    procedure RemoveLinks(RecVariant: Variant)
```
PeterConijn, line 68: Since the facade procedure is public, perhaps you might want to add a check to ensure that the passed argument is a record variant.

Then this might also be something you would want to include in your tests.

adrogin: I think I can throw a very generic error message in this case, because it's not something that users can fix themselves anyway.

JesperSchulz:
This PR will ship with 21.4.

JesperSchulz (approved):

JesperSchulz:

Build ID: 51294.

JesperSchulz:

Build ID to track: 51599.

PR #21244, opened 2022-11-30
Copy filters on test suite update

adrogin:
Action "Update Test Methods" in AL Test Tool resets filters on the page, mixing test codeunit from different suites

Steps to reproduce

Run AL Test Tool and pull a test codeunit into the Default test suite
Create another test suite and get a test codeunit
Run Update Test Methods action
Result: Test codeunits from all suites are displayed, page filters are lost.
Expected result: The action should keep the filters.

Issue: Functions GetView/SetView which are used to copy filters between record instances only pick filters in the active filter group (currently 0), while the suite filter is applied in group 2.
Solution: Use CopyFilters instead, which copies filters in all groups.


JesperSchulz (approved):

JesperSchulz:

Build ID to track: 50796.

JesperSchulz:

Build ID: 50848.

PR #21246, opened 2022-11-30
Added time zone management module

PeterConijn:
This module aims to provide methods to correctly calculate the offset between a datetime in a time zone from UTC and the offset between time zones. Other than the functions in the Type Helper codeunit in the base app, this module takes daylight saving time into account when applicable.

Additionally, the module provides indicators on whether a time zone supports daylight saving time or whether a given datetime falls within the daylight saving time period for a given time zone.


Modules/System/Time Zone Management/src/TimeZoneInfoInitializer.Codeunit.al
```diff
@@ -0,0 +1,23 @@
...
+    local procedure TryInstantiateTimeZoneInfo(TimeZoneId: Text; var TimeZoneInfoDotNet: DotNet TimeZoneInfo)
+    begin
+        TimeZoneInfoDotNet := TimeZoneInfoDotNet.FindSystemTimeZoneById(TimeZoneId);
+    end;
+
+    local procedure ThrowInvalidTimeZoneIdError(TimeZoneId: Text)
+    var
+        InvalidTimeZoneIdErr: Label 'You have passed an invalid timezone ID (%1). Please reference the time zone list for supported time zone IDs.', Comment = '%1 = The invalid time zone ID passed to the procedure.';
+    begin
+        Error(InvalidTimeZoneIdErr);
```
pri-kise, line 21: The TimeZoneId is missing in this error

PeterConijn: Fixed missing TimeZoneId.

pri-kise:
@PeterConijn your pull request looks very promising.

It would be really useful in my opinion to provide a procedure that uses the current time zone to calculate the offset.
Additionally, have you thought about adding a wrapper for `GetSystemTimeZones` to give access to the available time zone ids?
(https://learn.microsoft.com/en-us/previous-versions/windows/embedded/ms912391(v=winembedded.11)?redirectedfrom=MSDN)
If I read the documentation correctly these names must not be always the same.
https://learn.microsoft.com/de-de/dotnet/api/system.timezoneinfo.getsystemtimezones?view=net-7.0

Now waiting for someone of microsoft.

PeterConijn:
> It would be really useful in my opinion to provide a procedure that uses the current time zone to calculate the offset.

That is a good idea. I will look into it and add it to the PR

> Additionally, have you thought about adding a wrapper for `GetSystemTimeZones` to give access to the available time zone ids?

A list of time zones is already handled by the Time Zone Selection module in the System App so it would be superfluous.

pri-kise:
@PeterConijn thanks for clarification. I must have missed this System Feature.

I guess you have to include your permissionset in on the system application basic permissonset

NKarolak:
@pri-kise, as Peter indicated: Once this PR is merged, please remember to obsolete/move TypeHelper functions to the new module. Old code:
https://github.com/StefanMaron/MSDyn365BC.Code.History/blob/bcbdcfc434259c1bfc3708b55d62752c5b5fca26/BaseApp/Source/Base%20Application/TypeHelper.Codeunit.al#L328

When you search the Base app for further occurrences of "TimeZone", you will find even more procedures to obsolete, such as https://github.com/StefanMaron/MSDyn365BC.Code.History/blob/1c682b0af15e55f729253392b821e64244712e68/BaseApp/Source/Base%20Application/TransformationRule.Table.al#L378 (provided such function exists in the new module; I did not check).

Modules/System/Time Zone Management/src/DateTimeOffset.Codeunit.al
```diff
@@ -0,0 +1,38 @@
...
+    begin
+        exit(DateTimeConversionImpl.GetUtcOffset(SourceDateTime, TimeZoneId));
+    end;
+
+    /// <summary>
+    /// Retrieves the offset from the user time zone at the time of the requested datetime. This takes into account any daylight saving time conditions that may apply.
+    /// </summary>
+    /// <param name="SourceDateTime">The datetime that will be used as the basis for the difference calculation.</param>
+    /// <returns>A duration that indicates the offset between UTC and the user time zone for the provided datetime.</returns>
+    procedure GetUtcOffsetForUserTimeZone(SourceDateTime: DateTime; TimeZoneId: Text): Duration
```
pri-kise, line 19: Signature of procedure should be updatet to only need the SourceDateTime

PeterConijn: Shows you should never code while on the phone. 

I fixed the signature and added some other minor fixes to the PR.

PeterConijn:
The automated builds seem to fail on the object nos (I used the default 50100-50149 range). Maybe a noobie question, but is this something Microsoft will do or is that something expected of me? If the latter, I would like to know which object numbers I can use.

pri-kise:
The best way is to wait for someone of microsoft and update the id's for the Tests and the new Objects afterwards.

pri-kise:
@JesperSchulz could you please assign someone to this pull request?
To give feedback if this module can be accepted and to provide some new Ids, that can be used for this new module?

JesperSchulz:
I was hoping for other members of the community to review this code too ;-) I know this particular repository has a history of "partner develops, Microsoft reviews", but in the long run that won't scale. This PR is something that many partners over time have requested, and I was hoping to see some involvement from them.
Oh yeah, and it's year-end, which always means "go-time" for all Business Central developers - that may also play into, why we haven't looked at this PR yet ourselves ;-) But I promise to do so today/tomorrow! The ID problem is an easy one to fix - that I can do for sure. Stay tuned!

pri-kise:
@JesperSchulz do you have some Ids for the Test Codeunits too?

JesperSchulz:
132979 seems to be available, both for the PS and test :-)

PeterConijn:
> @PeterConijn, if you grant me permissions to your repo, I can push the new IDs :-) Alternatively, you can update them yourself. I've reserved 8720 - 8724 for you.

Thanks @JesperSchulz. I have done both. You have contributor access to my fork and I have updated the object numbers in the app, test, and test permissions.

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
```diff
@@ -0,0 +1,10 @@
+permissionset 132979 TimeZoneMgtObjects
```
pri-kise, line 1: This PermissionSet should probably use the Id of the normal app `8720`.
The ID Range: 
```json
{
      "from": 132979,
      "to": 132979
    }
```
is only meant for testapps

JesperSchulz:
Great! Now let's see if this builds, and then it's time for code review 🥳

JesperSchulz:
Build failed. Here's the error: AL0185 Codeunit 'Library - Test Initialize' is missing

Please make build pass before we continue code review.

PeterConijn:
Thanks for the reminder! 

Fixed the bug. Build succeeded.

adrogin (approved):

pri-kise (dismissed):

Modules/System/Time Zone Management/src/DateTimeOffsetImpl.Codeunit.al
```diff
@@ -0,0 +1,37 @@
...
+        Offset := TimeZoneInfoDotNet.GetUtcOffset(SourceDateTime);
+        exit(Offset);
+    end;
+
+    procedure GetUtcOffsetForUserTimeZone(SourceDateTime: DateTime): Duration
+    var
+        Session: SessionSettings;
+    begin
+        Session.Init();
+        if Session.TimeZone <> '' then
```
miljance, line 21: Can the Session have empty in TimeZone? If yes what happens in this case - Duration will be 0? If no is this line really needed?

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
```diff
@@ -0,0 +1,10 @@
+permissionset 8720 TimeZoneMgtObjects
```
miljance, line 1: I think that the permission set should be included in objects PermissionSet 219 "System Application - Objects". Otherwise this permission set is not accessible nor assignable.

miljance (changes requested):
Could you please include the permission set as noted in the feedback?

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
```diff
@@ -0,0 +1,10 @@
...
+{
+    Caption = 'Time Zone Management - Objects';
+    Assignable = false;
+    Permissions = codeunit "DateTime Offset" = X,
```
JesperSchulz, line 5: Let's add a blank line here between assignable and permissions - just to make the permissions portion stand out a little more.

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
JesperSchulz, line 1: This permission set should probably be called "Time Zone Management - Objects", just like the caption. This one should be internal. You could then create a new permission set, which isn't internal and has the name "Time Zone Management - Execute". For now, that permission set would only contain the objects one...

Modules/System/Time Zone Management/src/DateTimeOffset.Codeunit.al
```diff
@@ -0,0 +1,38 @@
+codeunit 8720 "DateTime Offset"
```
JesperSchulz, line 1: As this is a public codeunit you must add:
```
/// <summary>
/// write me
/// <example>
write me
/// </example>
/// </summary>
```

Also, you need to add the open source header to all objects:
```
// ------------------------------------------------------------------------------------------------
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License. See License.txt in the project root for license information.
// ------------------------------------------------------------------------------------------------
```

Modules/System/Time Zone Management/src/DateTimeOffsetImpl.Codeunit.al
```diff
@@ -0,0 +1,37 @@
+codeunit 8721 "DateTime Offset Impl."
+{
+    Access = Internal;
+
+    procedure GetUtcOffset(SourceDateTime: DateTime; TimeZoneId: Text): Duration
+    var
+        TimeZoneInfoInitializer: Codeunit "TimeZoneInfo Initializer";
+        Offset: Duration;
+        TimeZoneInfoDotNet: DotNet TimeZoneInfo;
```
JesperSchulz, line 9: Looks like we need a dependency to the dotnet module.

Modules/System/Time Zone Management/src/TimeZoneInfoInitializer.Codeunit.al
```diff
@@ -0,0 +1,23 @@
+codeunit 8724 "TimeZoneInfo Initializer"
```
JesperSchulz, line 1: We usually don't CamelCase object names.

Modules/System/Time Zone Management/app.json
```diff
@@ -0,0 +1,33 @@
+{
+  "id": "4c315253-b024-4fdc-9c3d-794036768af5",
+  "name": "Time Zone Management",
```
JesperSchulz, line 3: I wonder if we really need a module for "Time Zone Management", or if it maybe would make sense to make the module a tad wider? We've got a lot of DateTime conversion functions in the BaseApp - maybe these could find a new home in this module too?

Modules/System/Time Zone Management/app.json
```diff
@@ -0,0 +1,33 @@
+{
+  "id": "4c315253-b024-4fdc-9c3d-794036768af5",
```
JesperSchulz, line 2: We've tried to stop calling things "management", as it really doesn't add much information to module. "Time Zone" would be fine, I guess?

Modules/System Tests/Time Zone Management/src/TimeZoneManagementTest.Codeunit.al
```diff
@@ -0,0 +1,332 @@
...
+        // [THEN] DateTime shows as not being in DST
+        VerifyDateTimeShowsAsNotBeingInDST(DateTimeIsInDST);
+    end;
+
+    var
+        PermissionsMock: Codeunit "Permissions Mock";
+        DateTimeOffset: Codeunit "DateTime Offset";
+        DaylightSavingTimeInfo: Codeunit "Daylight Saving Time Info";
+        LibraryAssert: Codeunit "Library Assert";
+        IsInitialized: Boolean;
```
JesperSchulz, line 205: Let's get all of the global variables moved to the top. It's kind of a standard we have.

Modules/System Tests/Time Zone Management/src/TimeZoneManagementTest.Codeunit.al
```diff
@@ -0,0 +1,332 @@
...
+        ExpectedOffset := -9 * GetHourConversionFactor();
+        LibraryAssert.AreEqual(ExpectedOffset, Offset, OffsetMismatchErr);
+    end;
+
+    local procedure CheckWhetherTimeZoneSupportsDaylightSavingTime(TimeZoneId: Text): Boolean
+    begin
+        exit(DaylightSavingTimeInfo.TimeZoneSupportsDaylightSavingTime(TimeZoneId));
+    end;
+
+    local procedure VerifyDaylightSavingTimeIsNotSupported(SupportsDST: Boolean)
```
JesperSchulz, line 266: These local helper methods seem only the called once in the tests - why have them in the first place?

Modules/System Tests/Time Zone Management/src/TimeZoneManagementTest.Codeunit.al
```diff
@@ -0,0 +1,332 @@
...
+            GetTimeZoneIdForDSTTimeZone():
+                begin
+                    if InDaylightSavingPeriod then
+                        exit(2);
+                    exit(1);
+                end;
+        end;
+    end;
+
+    local procedure GetTimeZoneIdForNonDSTTimeZone(): Text
```
JesperSchulz, line 313: If we've being honsest with ourselves, the name of this function should not be GetTimeZoneIdForNonDSTTimeZone(). It should rather be GetTimeZoneIdForUSMountainStandardTime. You are using a lot of hard coded values and "magic numbers" in your test - that could be improved.

JesperSchulz (changes requested):

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
pri-kise: @JesperSchulz  Keep in mind, that PermissionSets should have a max length of 20 chars.

Modules/System/Time Zone Management/src/DateTimeOffsetImpl.Codeunit.al
PeterConijn: I believe it can be empty when called from a web service, for example. See procedures GetTimezoneOffset and GetUserClientTypeOffset for similar patterns.

That said, if I remove this statement, the missing time zone will be handled with an error in the initializer codeunit. That is a choice that we will have to make. Given the code in codeunit 10, I opted for similar patterning as in that codeunit.

Modules/System/Time Zone Management/app.json
pri-kise: @JesperSchulz What are thinking of?
Something like "Date and Time" or "Date & Time Utility", ...
And if this becomes whole Date,DateTime,DateTimeOffset,TimeZone Module. Are the Id's still the correct ones or should they be changed to some where we can easily add several more objects without the need for additional Id Ranges?

Modules/System/Time Zone Management/src/DateTimeOffset.Codeunit.al
PeterConijn: Added the MIT license header and the summary.

Modules/System/Time Zone Management/src/TimeZoneInfoInitializer.Codeunit.al
PeterConijn: I did that in this case, since the DotNet object being initialized is TimeZoneInfo, but I have added spaces.

Modules/System/Time Zone Management/app.json
PeterConijn: Perhaps, but given your previous comment where you would like to add codeunit 10 BaseApp functionality to it, would it be better to resolve this once agreement has been reached there?

Modules/System Tests/Time Zone Management/src/TimeZoneManagementTest.Codeunit.al
PeterConijn: It keeps the test procedures better structured in the given-when-then pattern, thus increasing readability.

Modules/System Tests/Time Zone Management/src/TimeZoneManagementTest.Codeunit.al
PeterConijn: True, but the problem with this sort of functionality is that it does not exist yet. In a good test, you would have two ways of reaching the same result: the one you want to test and the control where you know the answer is correct.

The problem here is that there is no real alternative method to reach the correct offset; if there were, I would not have needed to create this module. The only thing I can do is use the same method using the TimeZoneMgt dotnet module, which would reliably use the same  time.

A possible alternative is using several time zones in the set, but I don't really see a way around using hard-coded information.

Modules/System Tests/Time Zone Management/src/TimeZoneManagementTest.Codeunit.al
PeterConijn: Yes, you are right, of course. My test codeunit template puts them in that spot and I forgot to move them. Did that now.

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
JesperSchulz: 30, I believe? We've got plenty with a length of more than 20 chars (e.g. permissionset 9010 "AAD User Management - Objects").

Modules/System/Time Zone Management/app.json
JesperSchulz: Exactly, a "Date & Time" module. That'd be lovely. And no, we would not have to change the ID's. While ranges are preferable, they're not a requirement. We can pick individual IDs.

Modules/System/Time Zone Management/app.json
JesperSchulz: Absolutely.

JesperSchulz: So, what should we agree upon? Should we make a "Date & Time" module, which contains a Timezone facade and implementation codeunit, containing both DateTimeOffset and DaylightSavingTimeInfo? It would be a more versatile module.
Input to that idea?

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
pri-kise: I recieve the following warning in VS Code.
image
Maybe this warning is outdated.
But there has been some limitation.
At least the Role ID in`table 2000000167 "Aggregate Permission Set"` is limited to Code[20]
and the Name is limited to Text[30].
Therefore we always add a maxlength = 30 to the PermissionSet Caption

```al
table 2000000167 "Aggregate Permission Set"
{
    DataPerCompany = false;
    Scope = Cloud;

    //WriteProtected=True;
    fields
    {
        field(1; Scope; Option)
        {
            OptionMembers = System,Tenant;
        }
        field(2; "App ID"; Guid)
        {
        }
        field(3; "Role ID"; Code[20])
        {
        }
        field(4; Name; Text[30])
        {
        }
        field(5; "App Name"; Text[250])
        {
        }
    }
    keys
    {
        key(pk; Scope, "App ID", "Role ID")
        {
        }
    }

    fieldgroups
    {
        fieldgroup("DropDown"; "Role ID", Name, "App Name")
        {

        }
    }
}
```

Modules/System/Time Zone Management/app.json
PeterConijn: I like the Date & Time module idea. 

To summarize and make sure we're on the same page: it will have a single timezone facade and implementation codeunit that merges the functionality that is now contained in the two separate facade/implementation pairs.

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
PeterConijn: That would make sense in and of itself. @JesperSchulz, you indicated below that this should have its own public permission set.

> This permission set should probably be called "Time Zone Management - Objects", just like the caption. This one should be internal. You could then create a new permission set, which isn't internal and has the name "Time Zone Management - Execute". For now, that permission set would only contain the objects one...

Waiting for feedback on this one.

Modules/System/Time Zone Management/app.json
PeterConijn: Since we are having two similar conversations on this topic, I am marking this one as resolved and we can add any details on the one below this.

Modules/System/Time Zone Management/app.json
JesperSchulz: Correct. We're exactly on the same page! I like that idea too. It has even grown on me since yesterday :-) Let's do that!

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
PeterConijn: My compiler keeps crashing on the app.json (accessviolation error) so my code analyzer breaks like mad, but I have updated the name.

Modules/System/Time Zone Management/src/DateTimeOffsetImpl.Codeunit.al
PeterConijn: Added it.

Modules/System/Time Zone Management/src/DateTimeOffsetImpl.Codeunit.al
PeterConijn: I have added the permission set to "System Application - Objects"; this seems the most logical place.

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
PeterConijn: After thinking this over, I agree with @miljance that adding it to the "System Application - Objects" would suffice. A separate assignable time zone permission seems impractical.

Modules/System/Time Zone Management/app.json
PeterConijn: Renamed the module and merged the face and implementation as discussed.

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
JesperSchulz: About permission sets... from 2023 wave 1, we're going to implement inherent permissions on these kinds of modules (permissions defined in the objects, which aren't controllable). Nobody wants to control permissions on "Time Zone" calculations - everyone should just be able to use that code. Hence, please do put it into "System Application - Objects", and I will convert the permissions to inherent permissions in the objects, when porting the code to 2023 Wave 1.

PeterConijn: Makes sense. I have added it to "System Application - Objects".

Modules/System/Permissions/SystemApplicationObjects.PermissionSet.al
```diff
@@ -88,5 +88,6 @@ PermissionSet 219 "System Application - Objects"
                              "Web Service Management - Obj.",
                              "Word Templates - Objects",
                              "XML Validation - Objects",
-                             "XmlWriter - Objects";
+                             "XmlWriter - Objects",
+                             "Date & Time - Objects";
```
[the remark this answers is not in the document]
PeterConijn: The weird thing is that I remember staring at that list of permissions for that specific reason and my brain somehow forgot how the alphabet worked.

Fixed!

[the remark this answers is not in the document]
SHermann73: That was fast! :-)
Great, thanks.

Modules/System/Time Zone Management/Permissions/TimeZoneMgtObjects.PermissionSet.al
PeterConijn: Permissions added to System Application - Objects

Modules/System Tests/Time Zone Management/src/TimeZoneManagementTest.Codeunit.al
PeterConijn: I have made some changes to the test feature to introduce a wider array of time zones to test and a certain degree of randomness in order to make the test more robust.

JesperSchulz:
Alright, that should do it. I made a few changes to the module locally (e.g. removed permissions and used inherent permissions instead). This module will ship with v. 22 in April!

PeterConijn:
> Alright, that should do it. I made a few changes to the module locally (e.g. removed permissions and used inherent permissions instead). This module will ship with v. 22 in April!

That's great news! Thank you very much for your collaboration in this module.

JesperSchulz:

Build ID: 52216.

JesperSchulz:
Merging the PR, even though it means the changes will get reverted with 21.4, as the module doesn't release until 22.0. But don't worry! The module will be back 🥳

JesperSchulz (approved):
Approved. I've made some further changes in the internal PR, but most of the code stays as it is in this PR :-)

PR #21494, opened 2022-12-22
ItemJnlLine being intialized after the event

azadhussainster:
If we populate anything in ItemJnlLine by subscribing to OnBeforeSubcontractComponentSendPost event, it is lost in the next line "ItemJnlLine.Init();".

### DIFF
--- a/Apps/IN/INGST/app/GSTSubcontracting/src/Codeunit/SubcontractingPost.Codeunit.al
+++ b/Apps/IN/INGST/app/GSTSubcontracting/src/Codeunit/SubcontractingPost.Codeunit.al
@@ -99,9 +99,10 @@ codeunit 18466 "Subcontracting Post"
         TrackingQtyToHandle: Decimal;
         QuantitySent: Decimal;
     begin
-        OnBeforeSubcontractComponentSendPost(ItemJnlLine, DeliveryChallanHeader, SubOrderCompList);
-
         ItemJnlLine.Init();
+         
+        OnBeforeSubcontractComponentSendPost(ItemJnlLine, DeliveryChallanHeader, SubOrderCompList);
+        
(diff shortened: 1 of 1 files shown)


Apps/IN/INGST/app/GSTSubcontracting/src/Codeunit/SubcontractingPost.Codeunit.al
```diff
@@ -99,9 +99,10 @@ codeunit 18466 "Subcontracting Post"
         TrackingQtyToHandle: Decimal;
         QuantitySent: Decimal;
     begin
-        OnBeforeSubcontractComponentSendPost(ItemJnlLine, DeliveryChallanHeader, SubOrderCompList);
-
         ItemJnlLine.Init();
+         
+        OnBeforeSubcontractComponentSendPost(ItemJnlLine, DeliveryChallanHeader, SubOrderCompList);
```
adrogin, line 104: Maybe this is just due to my lack of knowledge of the local GST functionality, but to me, moving the event does not seem to solve the issue, at least in general, for all partners that possibly use it. Even if the event is triggered after the initialization of the journal line, the fields are overwritten by subsequent code.
If we think about a wider context, could another event before `ItemJnlPostLine.Run` be a solution?

azadhussainster: You are correct when updating the values of base fields in ILE. However,  extended fields can be filled and values will be preserved if the event is moved after init. 

Design wise, as you said, an event just before posting is the best as it can be used to fill base as well as extended fields.

adrogin (approved):

JesperSchulz:

Build ID to track: 54497.

JesperSchulz:

Build ID to track: 54495.

JesperSchulz:

Build ID: 10162.

PR #21626, opened 2023-01-04
FIX: correct PI

sigttou:
looking at various sources, the 17th position should be 2, not 1, the following digits would be 3846.

https://decimal.info/digits-of-pi/value-of-pi-to-20-decimal-places.html http://www.math.com/tables/constants/pi.htm
https://www.piday.org/million/

### DIFF
--- a/Modules/System/Math/src/Math.Codeunit.al
+++ b/Modules/System/Math/src/Math.Codeunit.al
@@ -19,7 +19,7 @@ codeunit 710 Math
     /// <returns>Value of pi.</returns>
     procedure Pi(): Decimal
     begin
-        exit(3.1415926535897932);
+        exit(DotnetMath.PI);
     end;
 
     /// <summary>
@@ -28,7 +28,7 @@ codeunit 710 Math
     /// <returns>Value of E.</returns>
     procedure E(): Decimal
     begin
-        exit(2.7182818284590451);
+        exit(DotnetMath.E);
     end;
 
     /// <summary>


PeterConijn (approved):

mazhelez (approved):

mazhelez:
> Huh, maybe some kind of a rounding error, I guess it comes from the DotNet class itself: https://learn.microsoft.com/en-us/dotnet/api/system.math.pi?view=net-7.0 Some discussion can be found on threads like Stackoverflow where different precisions seem to have been used in the source code through different versions. Then I wonder, why doesn't the procedure do `exit(DotnetMath.PI);` (and do the same for `E()`)? which also seem to have the wrong last character)

That's a very good point! @sigttou what's your opinion?

darjoo (approved):

JesperSchulz:
Let's get this one in :-) Will process right away.

JesperSchulz:
Turns out, the code does not compile. Investigating. But as the result of DotnetMath.PI _is_ 3.1415926535897931, I don't feel like spending too much time on this.

JesperSchulz:
@sigttou, out of curiosity (before I start spending a lot of time on investigation): has the rounding of PI caused you any problem, or is it just the perfectionist in you wanting to fix this? ;-)

JesperSchulz:
@darjoo couldn't help himself and did some investigation 😋 Turns out, it's a compiler issue. I'm not sure where this lands on the compiler team priority-wise, but eventually this will get fixed. Until then, this PR is blocked (which again I believe/hope should be fine, as the end-result will be identical).

JesperSchulz:
Unfortunately, the compiler team replied that this is "By Design" for now. As it isn't trivial to add support for fields, this is not something which is going to happen as part of this request. I'm afraid we will have to live with this constant for now. But I think it'll be alright. Closing this PR, as there isn't any action to take at this point in time.

PR #21661, opened 2023-01-09
[Shopify] Delivery of part 1 of M3 cycle

FrankAuwers:


Apps/W1/Shopify/app/src/Base/Tables/ShpfyShop.Table.al
```diff
@@ -507,33 +553,46 @@ table 30102 "Shpfy Shop"
...
+    internal procedure GetLastSyncTime(Type: enum "Shpfy Synchronization Type"): DateTime
     var
-        SyncInfo: Record "Shpfy Synchronization Info";
+        ShpfySynchronizationInfo: Record "Shpfy Synchronization Info";
     begin
-        if SyncInfo.Get(Rec.Code, Type) then
-            exit(SyncInfo."Last Sync Time");
+        if Type = "Shpfy Synchronization Type"::Orders then begin
+            if Rec."Shop Id" = 0 then begin
+                CalcShopId();
```
grobyns, line 562: now that we have these implicit with warnings some of the code becomes very confusing. 
here we have:
CalcShopId()
Rec.Modify()

now, potentially this would act on two different instances... Well, at least it's unclear what CalcShopId() could possibly do to the Rec that it deserves a modify.

Then you look at that method and there's not a rec in sight which adds to the confusion.

So:
first, Rec. should be added everywhere. simple.

Should the code be:
Rec.CalcShopId();
Rec.Modify();

well, yes... 

and yes i know it makes no difference to the execution but it is much easier on the eyes...

Apps/W1/Shopify/app/src/Base/Tables/ShpfyShop.Table.al
```diff
@@ -507,33 +553,46 @@ table 30102 "Shpfy Shop"
...
     end;
 
-    internal procedure SetLastSyncTime(Type: Enum "Shpfy Synchronization Type")
+    internal procedure SetLastSyncTime(Type: enum "Shpfy Synchronization Type")
     begin
         SetLastSyncTime(Type, CurrentDateTime);
     end;
 
-    internal procedure SetLastSyncTime(Type: Enum "Shpfy Synchronization Type"; ToDateTime: DateTime)
+    internal procedure SetLastSyncTime(Type: enum "Shpfy Synchronization Type"; ToDateTime: DateTime)
```
grobyns, line 578: Shouldn't ToDateTime be named LastSyncDateTime?

Apps/W1/Shopify/app/src/Base/Tables/ShpfyShop.Table.al
```diff
@@ -507,33 +553,46 @@ table 30102 "Shpfy Shop"
...
     var
-        SyncInfo: Record "Shpfy Synchronization Info";
+        ShpfySynchronizationInfo: Record "Shpfy Synchronization Info";
```
grobyns, line 580: There's inconsistent naming throughout the app. Sometimes the variable has the Shpfy prefix, sometimes it doesn't. Sometimes the prefix is Shopify.
It think it's fine to drop the Shpfy prefix. In most cases it doesn't add anything. In some cases, dropping the prefix makes the name a little confusing so there, adding Shopify seems the right thing to do.
ex: Record "Shpfy Customer" would be confusing as just Customer, but perfectly ok as ShopifyCustomer.

What is worse though is when the same object has different names in different places. For example, I have seen "Shpfy Shop" named as ShpfyShop, ShopifyShop and just Shop.

please make the naming consistent throughout the code.

Apps/W1/Shopify/app/src/Customer/Codeunits/ShpfyCustByBillto.Codeunit.al
```diff
@@ -54,28 +54,28 @@ codeunit 30111 "Shpfy Cust. By Bill-to" implements "Shpfy ICustomer Mapping"
...
-        Address.SetRange("Customer Id", CustomerId);
-        Address.SetFilter("Address 1", FilterMgt.CleanFilterValue(JsonHelper.GetValueAsText(JCustomerInfo, 'Address'), MaxStrLen(Address."Address 1")));
-        Address.SetFilter("Address 2", FilterMgt.CleanFilterValue(JsonHelper.GetValueAsText(JCustomerInfo, 'Address2'), MaxStrLen(Address."Address 2")));
-        Address.SetRange(Zip, JsonHelper.GetValueAsText(JCustomerInfo, 'PostCode'));
-        Address.SetRange(City, JsonHelper.GetValueAsText(JCustomerInfo, 'City'));
-        Address.SetRange("Country/Region Code", JsonHelper.GetValueAsText(JCustomerInfo, 'CountryCode'));
-        if Address.IsEmpty then
+        CustomerAddress.SetAutoCalcFields("Customer No.");
+        CustomerAddress.SetRange("Customer Id", CustomerId);
+        CustomerAddress.SetFilter("Address 1", FilterMgt.CleanFilterValue(JsonHelper.GetValueAsText(JCustomerInfo, 'Address'), MaxStrLen(CustomerAddress."Address 1")));
```
grobyns, line 59: this is interesting. the clean does this:
        exit('@' + Value.Replace('(', '?').Replace(')', '?').Replace('*', '?').Replace('.', '?').Replace('<', '?').Replace('>', '?').Replace('=', '?'));


So basically make a case insensitive wildcard filter. But, this could get you wrong results. The ? will accept any character, not only the ones that mess with the filterstring. 

Also, there is a good way of filtering that works even when the value contains these chars and this is not it.

Say the address is 
'BC (AL)',

this filter thing will return anything that has 'BC ?AL?' (case insensitive). like, BC HALO but that's not correct is it? 

What I wonder about is whether there was actually an issue or its this way because you thought you had to avoid one. 
If the cases insensitive thing isn't truly required, a SetRange should have done the trick. 

it seems to me the loopy code below is to weed out all the addresses this finds that are not the ones you're looking for... 

I'd be interested in seeing an actual case where this is needed and then trying to find an easier solution to the filtering.

Apps/W1/Shopify/app/src/Customer/Codeunits/ShpfyCustByBillto.Codeunit.al
```diff
@@ -35,8 +35,8 @@ codeunit 30111 "Shpfy Cust. By Bill-to" implements "Shpfy ICustomer Mapping"
     internal procedure DoMapping(CustomerId: BigInteger; JCustomerInfo: JsonObject; ShopCode: Code[20]; TemplateCode: Code[10]; AllowCreate: Boolean): Code[20];
```
grobyns, line 35: this function could really use being split into several smaller bits. It's very long and does multiple things

Apps/W1/Shopify/app/src/Order handling/Codeunits/ShpfyOrdersAPI.Codeunit.al
```diff
@@ -191,34 +197,37 @@ codeunit 30165 "Shpfy Orders API"
...
-                    JsonHelper.GetValueIntoField(JNode, 'fullyPaid', RecRef, OrderToImport.FieldNo("Fully Paid"));
-                    JsonHelper.GetValueIntoField(JNode, 'unpaid', RecRef, OrderToImport.FieldNo(Unpaid));
-                    JsonHelper.GetValueIntoField(JNode, 'subtotalLineItemsQuantity', RecRef, OrderToImport.FieldNo("Total Quantity of Items"));
-                    JsonHelper.GetValueIntoField(JNode, 'totalPriceSet.shopMoney.amount', RecRef, OrderToImport.FieldNo("Order Amount"));
-                    JsonHelper.GetValueIntoField(JNode, 'totalPriceSet.shopMoney.currencyCode', RecRef, OrderToImport.FieldNo("Currency Code"));
-                    RecRef.SetTable(OrderToImport);
-                    OrderToImport."Risk Level" := ConvertToRiskLevel(JsonHelper.GetValueAsText(JNode, 'riskLevel'));
-                    OrderToImport."Financial Status" := ConvertToFinancialStatus(JsonHelper.GetValueAsText(JNode, 'displayFinancialStatus'));
-                    OrderToImport."Fulfillment Status" := ConvertToFulfillmentStatus(JsonHelper.GetValueAsText(JNode, 'displayFulfillmentStatus'));
+                    // OrderToImport.SetCurrentKey(Id, "Shop Id");
```
grobyns, line 200: let's not keep commented out code

Apps/W1/Shopify/app/src/Order handling/Tables/ShpfyOrderTaxLine.Table.al
```diff
@@ -44,6 +44,21 @@ table 30122 "Shpfy Order Tax Line"
             Caption = 'Currency Code';
             DataClassification = SystemMetadata;
             Editable = false;
+            ObsoleteState = Pending;
+            ObsoleteTag = '21.3.0.0';
+            ObsoleteReason = 'Is available in Shopify Order Header table';
+        }
+        field(7; "Presentment Amount"; Decimal)
```
grobyns, line 51: is Presentment really the correct term?

grobyns (changes requested):

Apps/W1/Shopify/app/src/Order handling/Tables/ShpfyOrderTaxLine.Table.al
MathijsSchols: Presentment is the correct name, this is the name used in Shopify.

grobyns: 👍

Apps/W1/Shopify/app/src/Order handling/Codeunits/ShpfyImportOrder.Codeunit.al
```diff
@@ -15,120 +15,325 @@ codeunit 30161 "Shpfy Import Order"
...
+        ShpfyShippingCharges.UpdateShippingCostInfos(ShpfyOrderHeader);
+        ShpfyTransactions.UpdateTransactionInfos(ShpfyOrderHeader."Shopify Order Id");
+        if IsNew then begin
+            ShpfyOrderTransaction.SetRange("Shopify Order Id", OrderId);
+            ShpfyOrderTransaction.SetFilter(Status, '%1|%2', "Shpfy Transaction Status"::Pending, "Shpfy Transaction Status"::Success);
+            ShpfyOrderTransaction.SetFilter(Type, '%1|%2', "Shpfy Transaction Type"::Sale, "Shpfy Transaction Type"::Capture);
+            ShpfyOrderTransaction.SetCurrentKey(Amount, "Shopify Order Id", Status, Type);
+            ShpfyOrderTransaction.SetAscending(Amount, false);
+            if ShpfyOrderTransaction.FindFirst() then
+                ShpfyOrderHeader.Gateway := ShpfyOrderTransaction.Gateway;
```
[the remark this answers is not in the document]
grobyns: My bad, must have misread the code...

FrankAuwers:
Comments where solved with Fix-PR-comments upload.

onbuyuka:
We have merged this PR into our repo. Please get latest changes from this repo to your fork, create a new branch and a PR out of it for new changes.

PR #21716, opened 2023-01-11
Add Release workflow 

aholstrup1:


mazhelez (commented):
Looks alright to me.
Few more comments:
- Let's organize the `PackNuget.ps1` and the template nuspec in a folder somewhere. Otherwise, .`github` folder will get crowded.
- What about cleaning up `.github\DeliverToBCFeed.ps1` and somehow turning off the _Deliver_ job from the CI/CD workflow?
- Ideally, we also want to clean up in _Modules\.AL-Go\CompileAppInBcContainer.ps1_ as well.

Build/Scripts/PackNuget.ps1
```diff
@@ -0,0 +1,74 @@
...
+    $PackageId,
+    [Parameter(Mandatory=$true)]
+    $Version,
+    [Parameter(Mandatory=$true)]
+    $Authors,
+    [Parameter(Mandatory=$true)]
+    $Owners
+)
+{
+    [xml] $template = Get-Content "$RepoRoot\Build\ALAppExtensions.template.nuspec"
```
mazhelez, line 26: Instead of using `$RepoRoot` and hardcoding the path to the template manifest, why don't we simply add a parameter `$ManifestTemplate` or `$ManifestTemplatePath`?

Build/Scripts/PackNuget.ps1
```diff
@@ -0,0 +1,74 @@
...
+    $appsToPackage = Join-Path $_ 'Package'
+    
+    if(Test-Path -Path $appsToPackage) 
+    {
+        Copy-Item -Path "$appsToPackage/*" -Destination "$OutputPackageFolder/Apps" -Recurse -Force
+    }
+}
+
+# Copy over the license file
+Copy-Item -Path "$RepoRoot/LICENSE" -Destination "$OutputPackageFolder" -Force
```
mazhelez, line 72: Similar to the manifest template, I suggest a `$LicensePath` parameter or something like that.

Build/Scripts/PackNuget.ps1
```diff
@@ -0,0 +1,74 @@
...
+    $template.package.metadata.authors = $Authors
+    $template.package.metadata.owners = $Owners
+
+    return $template
+}
+
+New-Item -Path $OutputPackageFolder -ItemType Directory | Out-Null
+
+$appsFolders = Get-ChildItem $BuildArtifactsPath -Directory 
+$packageVersion = ($appsFolders -replace ".*-Apps-","" | Select-Object -First 1).ToString()
```
mazhelez, line 39: This line is better suited under the `# Generate Nuspec file` comment

mazhelez (commented):
What about splitting the _Prepare Package_ step into 3 steps:

-- Organize/Prepare package
-- Generate manifest
-- (Nuget) Pack

That way it'll be easier to test the different steps (the ones that need testing).

mazhelez (approved):

PR #21831, opened 2023-01-19
Add unusedALGoSystemFiles setting

mazhelez:


.github/workflows/CICD.yaml
```diff
@@ -146,14 +146,14 @@ jobs:
 
       - name: Initialize the workflow
         id: init
-        uses: microsoft/AL-Go-Actions/WorkflowInitialize@preview
+        uses: microsoft/AL-Go-Actions/WorkflowInitialize@v2.3
```
mazhelez, line 149: Why not _preview_ anymore?

mazhelez: @freddydk Do you if this is intended change?

aholstrup1 (approved):

.github/workflows/CICD.yaml
mazhelez: Seems that this will be addressed in the next AL-Go version.

PR #21909, opened 2023-01-27
Add documentation that filter tokens need to start with a % sign

StefanMaron:
I just struggled a few hours with the filter tokens for a code field. 

It seems like the filter token event implementation on the service it self is not consistent with how the events are raised.

For Date fields it works if the filter token is just some string.
But for Code/Text fields the events are not raised when the filter token does not start with a `%` sign.

I added this note to all procedure documentation and at the top of the readme to save this time for others :)

### DIFF
--- a/Modules/System/Filter Tokens/README.md
+++ b/Modules/System/Filter Tokens/README.md
@@ -12,7 +12,7 @@ You can add more filter tokens by subscribing to the following events:
 - `OnResolveTimeTokenFromDateTimeFilter`
 
 > **Warning**  
-> Every filter token needs to start with a `%` sign as first charakter. Otherwise the events listed above might not get raised at all.
+> Every filter token needs to start with a `%` sign as first character. Otherwise, the events listed above will not get raised at all.
 
 # Public Objects
 ## Filter Tokens (Codeunit 41)
--- a/Modules/System/Filter Tokens/README.md
+++ b/Modules/System/Filter Tokens/README.md
@@ -11,6 +11,8 @@ You can add more filter tokens by subscribing to the following events:
 - `OnResolveDateTokenFromDateTimeFilter`
 - `OnResolveTimeTokenFromDateTimeFilter`
 
+> **Warning**  
+> Every filter token needs to start with a `%` sign as first charakter. Otherwise the events listed above might not get raised at all.
 
 # Public Objects
 ## Filter Tokens (Codeunit 41)
--- a/Modules/System/Filter Tokens/README.md
+++ b/Modules/System/Filter Tokens/README.md
@@ -24,7 +24,7 @@ You can add more filter tokens by subscribing to the following events:
 
  Turns text that represents date formats into a valid date filter expression with respect to filter tokens and date formulas.
  Call this function from onValidate trigger of page field that should behave similar to system filters.
- The text from which the date filter should be extracted passed as VAR. For example: "%YESTERDAY", or " 01-01-2012 ".
+ The text from which the date filter should be extracted passed as VAR. For example: "%YESTERDAY", or "01-01-2012".
 
 #### Syntax
 ```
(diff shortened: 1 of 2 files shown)


JesperSchulz (approved):

JesperSchulz:
Much appreciated, thank you! I'll get this change reviewed and implemented for 2023 Wave 1!

Modules/System/Filter Tokens/README.md
```diff
@@ -11,6 +11,8 @@ You can add more filter tokens by subscribing to the following events:
 - `OnResolveDateTokenFromDateTimeFilter`
 - `OnResolveTimeTokenFromDateTimeFilter`
 
+> **Warning**  
+> Every filter token needs to start with a `%` sign as first charakter. Otherwise the events listed above might not get raised at all.
```
NKarolak, line 15: "character" instead of "charakter".
"Might not" or "will not/won't"?

Modules/System/Filter Tokens/README.md
```diff
@@ -22,7 +24,7 @@ You can add more filter tokens by subscribing to the following events:
 
  Turns text that represents date formats into a valid date filter expression with respect to filter tokens and date formulas.
  Call this function from onValidate trigger of page field that should behave similar to system filters.
- The text from which the date filter should be extracted passed as VAR. For example: "YESTERDAY", or " 01-01-2012 ".
+ The text from which the date filter should be extracted passed as VAR. For example: "%YESTERDAY", or " 01-01-2012 ".
```
NKarolak, line 27: `" 01-01-2012 "` really with the leading and trailing spaces? If not, then please also change all other occurrences further below.

NKarolak (changes requested):

Modules/System/Filter Tokens/README.md
StefanMaron: Thanks! Will correct this tomorrow. But its actually "might". At least without making this sense much longer. 

From what I saw, date filter work without % and code text filter do not. 

Maybe someone from the server team can check and verify?
@JesperSchulz

mazhelez:
```suggestion
> Every filter token needs to start with a `%` sign as first character. Otherwise, the events listed above will not get raised at all.
```

Using _suggestions_ here is quite convenient.

mazhelez (commented):
Can you please verify is using `%` sign in the XML doc is okay?

I think `%` needs to be escaped with `&percent;`

Modules/System/Filter Tokens/README.md
StefanMaron: I dont think so. Removed them

StefanMaron:
% seem to work in the documentation without problems
image

Modules/System/Filter Tokens/README.md
NKarolak: @mazhelez Next time I'll use suggestions for sure :) I just could not find where to add them. The 
image
 symbol was unknown for me. The UI in DevOps is so much more intuitive for me when reviewing PRs ...

JesperSchulz (approved):

JesperSchulz:
Processing for 2023 Wave 1. Will merge once completed.

darjoo:
Hi @StefanMaron,

I have to ask for some clarification, why do the filter tokens need to start with %?
What did you mean by code field? Could you give us an example? and which case did the event not trigger?

I ran the following example
----
I tested just MakeDateFilter for now.

Input: "YESTERDAY"
Output: "02/13/2023"
Event (OnResolveDateFilterToken) does trigger

Input: "%YESTERDAY"
Output: "%YESTERDAY"
Event (OnResolveDateFilterToken) does not trigger

StefanMaron:
Hi @darjoo 

That's very interesting!

I did the same tests for a fiel of type code[20] and I got the exact opposite results

darjoo:
The only information I could gather from colleagues about this is that it should only be required in the UI. https://learn.microsoft.com/en-us/dynamics365/business-central/ui-enter-criteria-filters

Which makes it really interesting how you're getting the opposite results from the tests conducted.. I'm not sure how that's happening.

JesperSchulz:
What do we do about this PR? Something seems off here.

StefanMaron:
> What do we do about this PR? Something seems off here.

It seems like the logic for filter tokens does not behave consistently.
I could do some more tests, but it would only be blackbox tests.
Or do you want to give this to the compiler team to verify the behaviour?

StefanMaron:
> @StefanMaron, I think that's the right path forward. I'll create a bug for the platform team based on this PR and have them investigate. I don't think we will need to process this PR though. I'm hence closing it once I've created the bug. Agreed?

Yep, all good. My intention was to improve either documentation or the logic behind it, since I felt like something was off.

PR #22026, opened 2023-02-03
GP Migration - Vendor Name 2 logic update and some code cleanup.

jaymckinney-enavate:
This change fixes an issue with the GP cloud migration and contains some code cleanup.

- If the GP Vendor Name and GP Vendor Check Name are the same, don't set the Name 2 field to the GP Vendor Check Name, otherwise the name is printed twice on checks.
- Code cleanup in the Vendor Migrator codeunit since I was working in that file.

Microsoft ADO feature number to reference when submitting PR to Microsoft: **463636**

### DIFF
--- a/Apps/W1/HybridGP/app/src/Migration/Support/HelperFunctions.codeunit.al
+++ b/Apps/W1/HybridGP/app/src/Migration/Support/HelperFunctions.codeunit.al
@@ -2026,4 +2026,9 @@ Codeunit 4037 "Helper Functions"
         Currency.Validate("Invoice Rounding Type", Currency."Invoice Rounding Type"::Nearest);
         Currency.Insert(true);
     end;
+
+    procedure StringEqualsCaseInsensitive(Text1: Text; Text2: Text): Boolean
+    begin
+        exit(UpperCase(Text1) = UpperCase(Text2));
+    end;
 }
--- a/Apps/W1/HybridGP/app/src/Migration/Vendors/GPVendorMigrator.codeunit.al
+++ b/Apps/W1/HybridGP/app/src/Migration/Vendors/GPVendorMigrator.codeunit.al
@@ -147,54 +147,74 @@ codeunit 4022 "GP Vendor Migrator"
     local procedure MigrateVendorDetails(GPVendor: Record "GP Vendor"; VendorDataMigrationFacade: Codeunit "Vendor Data Migration Facade")
     var
         CompanyInformation: Record "Company Information";
+        Vendor: Record Vendor;
+        GenBusinessPostingGroup: Record "Gen. Business Posting Group";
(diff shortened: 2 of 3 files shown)


Apps/W1/HybridGP/app/src/Migration/Support/HelperFunctions.codeunit.al
```diff
@@ -2026,4 +2026,9 @@ Codeunit 4037 "Helper Functions"
         Currency.Validate("Invoice Rounding Type", Currency."Invoice Rounding Type"::Nearest);
         Currency.Insert(true);
     end;
+
+    procedure StringEqualsCaseInsensitive(Text1: Text; Text2: Text): Boolean
```
nikolakukrika, line 2030: I would put this function inline or rename it to be more descriptive on why this is needed (e.g. we need case insensitive because it is uppercased in BC, then it would be GPStringEqualBCCode or something like this...

Text1 and Text2 could get a bit better name

Apps/W1/HybridGP/app/src/Migration/Vendors/GPVendorMigrator.codeunit.al
```diff
@@ -147,55 +147,75 @@ codeunit 4022 "GP Vendor Migrator"
...
+        Address1: Text[50];
+        Address2: Text[50];
+        City: Text[30];
+        State: Text[30];
     begin
-        VendorName := CopyStr(GPVendor.VENDNAME, 1, 50);
-        ContactName := CopyStr(GPVendor.VNDCNTCT, 1, 50);
-        if not VendorDataMigrationFacade.CreateVendorIfNeeded(CopyStr(GPVendor.VENDORID, 1, 20), VendorName) then
+        VendorNo := CopyStr(GPVendor.VENDORID, 1, MaxStrLen(Vendor."No."));
+        VendorName := CopyStr(GPVendor.VENDNAME.TrimEnd(), 1, MaxStrLen(VendorName));
```
nikolakukrika, line 169: Minor, you can pragma disable this one, no need to copy

nikolakukrika (approved):

Apps/W1/HybridGP/app/src/Migration/Support/HelperFunctions.codeunit.al
jaymckinney-enavate: This is meant to be more generic, something that could be reused when needed. The BC Code uppercase scenario is a common one that this could benefit from in both the migration code and in the unit tests.

The use case here is even if the casing is different for the vendor check name, it should still be considered the same name and shouldn't be set to Name 2.

JesperSchulz:

Build ID to track: 53590.

JesperSchulz:

Build ID: 53598.

PR #22109, opened 2023-02-10
Intrastat breaks when a Posted Document is deleted

griosdp:
Intrastat Module Codeunit first gets the Country/Region Code from the Item Ledger Entry as it used to do in the past when the app didn't exist.

However, now, based on the setup, it tries to update the country and re-read from the corresponding Posted Document, based on the setup field.

The problem is, BC allows you to delete those documents. Therefore, to use a Get without an if statements makes the Codeunit to break if the document doesn't exist anymore. Not only that, but there is no way to fix this functionally.

As a solution, BC needs to only update the country if it can find the document.

Also this logic is also executed in the functions for FA Ledger Entries and Job Ledger Entries


griosdp:

JesperSchulz (approved):
Makes sense. Approved.

JesperSchulz:
I will process this change for 2023 Wave 1.

JesperSchulz:

Build ID: 53526.

PR #22155, opened 2023-02-14
Remove Direct Posting flag from migration code.

ToddE123:
This will update the migration tool to no longer migrate the Direct Posting flag. All accounts migrated over from GP to BC will have the Direct Posting set to yes, the default BC behavior. With this change, the migration will have no effect on that setting.

PR 463635


JesperSchulz:

Build ID to track: 53566.

JesperSchulz:

Build ID: 53561.

Issue #2240, opened 2019-04-29
Event Request Table 414 - OutboxCheckSend()


maheshroyal34:
Event - OnBeforeOutboxCheckSend(), 
>>New Code Start
onBeforeOutboxCheckSend (Var ICOutboxTransaction2: “IC OutBound transaction”;Var IsHandled: Boolean)
<<New Code End

IF NOT IsHandled THEN >>New Code
  IF ICOutboxTransaction2.FINDFIRST THEN
  	IF NOT ConfirmManagement.ConfirmProcess(
STRSUBSTNO(Text001,ICOutboxTransaction2."Transaction No.","Transaction No."),TRUE)
  THEN
    ERROR('');

maheshroyal34:
The requested Event - OnBeforeOutboxCheckSend() to be the below position Table - 414 - "IC Outbox Transaction" in function OutboxCheckSend()

![image](https://user-images.githubusercontent.com/49781885/57704729-1f07b100-7680-11e9-8e8f-f12ab0b554b6.png)

PR #22402, opened 2023-02-28
SharePointClient.DeleteFile()

jwikman:
Adds the DeleteFile() function to the SharePoint client

### DIFF
--- a/Modules/System/SharePoint/src/SharePointClient.Codeunit.al
+++ b/Modules/System/SharePoint/src/SharePointClient.Codeunit.al
@@ -266,6 +266,15 @@ codeunit 9100 "SharePoint Client"
     procedure DownloadFileContent(OdataId: Text; FileName: Text): Boolean
     begin
         exit(SharePointClientImpl.DownloadFileContent(OdataId, FileName));
+
+    /// <summary>
+    /// Delete a file.
+    /// </summary>
+    /// <param name="OdataId">The odata.id parameter of the file entity.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure DeleteFile(OdataId: Text): Boolean
+    begin
+        exit(SharePointClientImpl.DeleteFile(OdataId));
     end;
 
     /// <summary>
--- a/Modules/System/SharePoint/src/SharePointClient.Codeunit.al
+++ b/Modules/System/SharePoint/src/SharePointClient.Codeunit.al
@@ -269,7 +269,7 @@ codeunit 9100 "SharePoint Client"
     end;
 
     /// <summary>
-    /// Delete a file.
+    /// Deletes a file.
     /// </summary>
     /// <param name="OdataId">The odata.id parameter of the file entity.</param>
     /// <returns>True if the operation was successful; otherwise - false.</returns>
--- a/Modules/System/SharePoint/src/SharePointClient.Codeunit.al
+++ b/Modules/System/SharePoint/src/SharePointClient.Codeunit.al
@@ -266,6 +266,7 @@ codeunit 9100 "SharePoint Client"
     procedure DownloadFileContent(OdataId: Text; FileName: Text): Boolean
     begin
         exit(SharePointClientImpl.DownloadFileContent(OdataId, FileName));
+    end;
 
     /// <summary>
     /// Delete a file.
--- a/Modules/System/SharePoint/src/SharePointClientImpl.Codeunit.al
+++ b/Modules/System/SharePoint/src/SharePointClientImpl.Codeunit.al
@@ -494,6 +494,21 @@ codeunit 9101 "SharePoint Client Impl."
         exit(true);
     end;
 
+    procedure DeleteFile(OdataId: Text): Boolean
+    var
+        SharePointHttpContent: Codeunit "SharePoint Http Content";
+    begin
+        SharePointUriBuilder.ResetPath(OdataId);
+        SharePointHttpContent.SetXHTTPMethod('DELETE');
+
+        SharePointRequestHelper.SetAuthorization(Authorization);
+        SharePointOperationResponse := SharePointRequestHelper.Post(SharePointUriBuilder, SharePointHttpContent);
+        if not SharePointOperationResponse.GetDiagnostics().IsSuccessStatusCode() then
+            exit(false);
+
+        exit(true);
+    end;
+
     procedure GetDocumentLibraryRootFolder(OdataID: Text; var SharePointFolder: Record "SharePoint Folder"): Boolean
     var
         SharePointFolderParser: Codeunit "SharePoint Folder";
--- a/Modules/System/SharePoint/src/SharePointClientImpl.Codeunit.al
+++ b/Modules/System/SharePoint/src/SharePointClientImpl.Codeunit.al
@@ -498,8 +498,9 @@ codeunit 9101 "SharePoint Client Impl."
     var
         SharePointHttpContent: Codeunit "SharePoint Http Content";
     begin
+        //POST https://{site_url}/_api/web/GetFileByServerRelativeUrl('/Folder Name/{file_name}')
         SharePointUriBuilder.ResetPath(OdataId);
-        SharePointHttpContent.SetXHTTPMethod('DELETE');
+        SharePointHttpContent.SetXHttpMethod('DELETE');
 
         SharePointRequestHelper.SetAuthorization(Authorization);
         SharePointOperationResponse := SharePointRequestHelper.Post(SharePointUriBuilder, SharePointHttpContent);
--- a/Modules/System/SharePoint/src/Diagnostics/HTTPDiagnostics.Interface.al
+++ b/Modules/System/SharePoint/src/Diagnostics/HTTPDiagnostics.Interface.al
@@ -11,7 +11,7 @@ interface "HTTP Diagnostics"
     Access = Public;
 
     /// <summary>
-    /// Gets reponse details.
+    /// Gets response details.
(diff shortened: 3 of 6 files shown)


JesperSchulz:
@kb-dynaway / @dynaway-kmi, what's your take on this PR? Does it look good to you?

jwikman:
I added a few small changes that we made in our internal code review.

dynaway-kmi:
@JesperSchulz Yes. The Delete does not return any data so there is nothing really to create tests for. 
Option to move file to recycle bin instead of permanent removal could be nice improvement for future PR.

Modules/System/SharePoint/src/SharePointClient.Codeunit.al
```diff
@@ -266,6 +266,15 @@ codeunit 9100 "SharePoint Client"
     procedure DownloadFileContent(OdataId: Text; FileName: Text): Boolean
     begin
         exit(SharePointClientImpl.DownloadFileContent(OdataId, FileName));
```
dynaway-kmi, line 268: end;

jwikman: My aplologies, I tried to multitask while doing this. Generally a bad idea. ;-)

An update is coming.

Modules/System/SharePoint/src/SharePointClientImpl.Codeunit.al
```diff
@@ -494,6 +494,22 @@ codeunit 9101 "SharePoint Client Impl."
         exit(true);
     end;
 
+    procedure DeleteFile(OdataId: Text): Boolean
+    var
+        SharePointHttpContent: Codeunit "SharePoint Http Content";
+    begin
+        //POST https://{site_url}/_api/web/GetFileByServerRelativeUrl('/Folder Name/{file_name}')
```
pri-kise, line 501: @jwikman I don't quite understand this approach to delete a file.
I know it's documented on the following page: https://learn.microsoft.com/en-us/sharepoint/dev/sp-add-ins/working-with-folders-and-files-with-rest

But isn't it possible to delte a sharepoint list item by using the graph API:
If it is possible then this restful API call should be added as option as well in my opinion.
https://learn.microsoft.com/en-us/graph/api/listitem-delete?view=graph-rest-1.0&tabs=http

If I understand the current code correctly, then they make use of both API right now..?
Or is it the same API?

pri-kise (approved):

glachana:
Hi All,

Is there a chance to see these changes merged into Master and published soon?
I am very interested in this feature.

Thanks

jwikman:
@JesperSchulz do you know if we can expect this to be merged anytime soon?

Modules/System/SharePoint/src/SharePointClientImpl.Codeunit.al
glachana: @pri-kise @jwikman 

For your information, this request also works perfectly to delete a file:
//DELETE https://{site_url}/_api/web/GetFileByServerRelativeUrl('/Folder Name/{file_name}')

It may be better to implement this method rather than using POST with the XHttpMethod header.

glachana: @jwikman 

There is no example of this kind of code in the documentation. But it is clearly stated that we can use these REST endpoints to perform typical CRUD operations on SharePoint entities.

documentation here : 
- https://learn.microsoft.com/en-us/sharepoint/dev/sp-add-ins/get-to-know-the-sharepoint-rest-service?tabs=csom

<html>
<body>
<div class="heading-wrapper" data-heading-level="h3" style="box-sizing: inherit; outline-color: inherit; position: relative; color: rgb(230, 230, 230); font-family: &quot;Segoe UI&quot;, SegoeUI, &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(23, 23, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><h3 id="use-http-commands-with-the-sharepoint-rest-service" class="heading-anchor" style="box-sizing: inherit; outline-color: inherit; margin: 30px 0px 18px; padding: 0px; font-size: clamp(1.3125rem, 15.4737px + 1.15132vw, 1.75rem); font-weight: 600; line-height: 1.3; margin-inline-start: -1.875rem; padding-inline-start: 1.875rem;">Use HTTP commands with the SharePoint REST service</h3></div><p style="box-sizing: inherit; outline-color: inherit; margin: 1rem 0px 0px; padding: 0px; overflow-wrap: break-word; color: rgb(230, 230, 230); font-family: &quot;Segoe UI&quot;, SegoeUI, &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(23, 23, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">To use the REST capabilities that are built into SharePoint, you construct a RESTful HTTP request by using the OData standard, which corresponds to the client object model API you want to use. The client.svc web service handles the HTTP request and serves the appropriate response in either Atom or JSON format. The client application must then parse that response.</p><p style="box-sizing: inherit; outline-color: inherit; margin: 1rem 0px 0px; padding: 0px; overflow-wrap: break-word; color: rgb(230, 230, 230); font-family: &quot;Segoe UI&quot;, SegoeUI, &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(23, 23, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The endpoints in the SharePoint REST service correspond to the types and members in the SharePoint client object models. By using HTTP requests, you can use these REST endpoints to do typical CRUD operations against SharePoint entities, such as lists and sites.</p><div class="has-inner-focus" style="box-sizing: inherit; outline-color: inherit; color: rgb(230, 230, 230); font-family: &quot;Segoe UI&quot;, SegoeUI, &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(23, 23, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">

If you want to do this to an endpoint | Use this HTTP request | Keep in mind
-- | -- | --
Read a resource | GET |  
Create or update a resource | POST | Use POST to create entities such as lists and sites.The SharePoint REST service supports sending POST commands that include object definitions to endpoints that represent collections.For POST operations, any properties that aren't required are set to their default values.If you attempt to set a read-only property as part of a POST operation, the service returns an exception.
Update or insert a resource | PUT | Use PUT and MERGE operations to update existing SharePoint objects.Any service endpoint that represents an object property set operation supports both PUT requests and MERGE requests.For MERGE requests, setting properties is optional; any properties that you don't explicitly set keep their current property.For PUT requests, if you don't specify all required properties in object updates, the REST service returns an exception.In addition, any optional properties you don't explicitly specify are set to their default properties.
Delete a resource | DELETE | Use the HTTP DELETE command against the specific endpoint URL to delete the SharePoint object represented by that endpoint.There are recyclable objects, such as lists, files, and list items, this results in a Recycle operation.

</div>
</body>
</html>

glachana: Here is my fork with the addition of the Delete Http method in SharePoint Request Helper.
It also adds SharePointClient.DeleteFile(), SharePointClient.DeleteFolder() and SharePoint.DownloadFileContent() overloads based on your PR. 

https://github.com/glachana/ALAppExtensions/tree/dev/sharepoint-upgrade-client 

Let me know if a PR from my branch might be worth it ;)

jwikman: Thanks for clarifying @glachana 

I don't like the X-Http-Method approach, so as long as the regular DELETE approach works - I think it is better.
Why do they even have the X-Http-Method implementation if DELETE works...? 🤔

I don't mind if this PR is scraped in advantage for yours with DELETE method.

Maybe someone else has an opinion on this? @dynaway-kmi or @pri-kise?

glachana: Here is my PR #22988. It contains the changes included in these two PR #22565 #2240. Regarding the delete function, it uses DELETE HTTP Commands instead of X-Http-Method implementation.

jwikman:
This one is replaced by #22988, since that solution is much cleaner!

Issue #22261, opened 2023-02-24
AcquireTokensWithCertificate() with password


jwikman:
We are working on an addition to the `SharePoint Authorization` module, to also support `Service to Service` / `Client Credentials` authentication for SharePoint.

It turned out that SharePoint does not work with Client Credentials tokens that are generated from Client Id and Client Secret. SharePoint API _only_ supports Client Credential tokens that are generated from Client Id and a **Certificate**.

We have this function to request a token with a certificate:
https://github.com/microsoft/ALAppExtensions/blob/9151fa39d6f6d548de0acdcf67512a30baf297ba/Modules/System/OAuth2/OAuth2Impl.Codeunit.al#L584
BUT, this function do not accept a password for the certificate.
The certificate to use should be a PKCS#12 certificate with private keys, most often saved to file as a .pfx file. It is a security best practice to protect the .pfx file with a password, and in most certificate generators the password is mandatory - which make sense!

I can see that the only usage of `AcquireTokensWithCertificate()` in the Base App is where the certificate is retrieved from a KeyVault, and in that case a password is not necessary - so I understand why the current implementation is as it is.

But I would like to request an overload of `AcquireTokensWithCertificate()` that also has a CertificatePassword (text), that could be passed to `AuthFlow.ALAcquireApplicationTokensWithCertificate()` and further to the X509 Certificate that probably is used in C# somewhere. ;-)

When that is in place, I'll create a PR (with some docs, because this is _not_ straight forward) to add Client Credentials to `SharePoint Authorization` module I don't think that we should add this without the usage of a password protected pfx. We only want to promote security best practices, don't you agree @JesperSchulz?

jwikman:
I've added a draft PR for the functionality we want to add,  #22404, so you can see what we want to accomplish.

jwikman:
I just realized that also the `AcquireTokensFromCacheWithCertificate()` and `AuthFlow.ALAcquireTokensFromCacheWithCertificate()` functions needs to be overloaded with a certificate password.

jwikman:
@IhorHandziuk what do you think about this request? Would it be possible to implement anytime soon? If you don't know, maybe you know who does?

JesperSchulz:
[...]

PR #22404, opened 2023-02-28
[DRAFT] SharePoint Client Credentials Grant Flow

jwikman:
This is a draft PR just to show that we are working on this topic.

Since the OAuth2 module does not yet support password protected certificates, I don't think we want to publish this PR just yet.
Support for password protected certificates are requested in #22261.

Missing pieces (at least):
- More manual tests
- Add certificate password when #22261 is implemented
- Tests
- Updated README


jwikman:
> @jwikman is there any chance that we get this PR ready for BC23? The issue for the AquireTokensWithCertificate seems to be fixed now.

Hi @pri-kise, 
I will probably not be able to do this any time soon, unfortunately. :(

So if no-one else pick this up, we'll have to live with this being added in some CU later on.

JesperSchulz:
As this is only a draft PR, I am closing it in this repository, as we're preparing to move the System Application to the new BCApps repository. Please create a new PR in the BCApps repository (from November) when you are ready to take the code out of draft mode.

PR #22435, opened 2023-03-03
Bug: multiple versions of Guided Experience Items displayed

kristjano:
When multiple versions of Guided Experience Items exist then all versions are displayd in pages. E.g. Title in Guided Experience Type of Assisted Setup has been update and new Guided Exp. Item with same Code but with Version 1 is created. When user opens the Assisted Setup page the temporary table is populated and all versions is added.

Solution:
InsertItem variable is initialized to false in every pass in loop over Guided Experience Items before evaluating if item should be added to temp table.


JesperSchulz (approved):
That looks about right to me. Approved.

PR #22660, opened 2023-03-21
[Packaging] Only package needed apps and validate package content

aholstrup1:
Only package needed apps and validate package content


Build/Scripts/Package/PackNuget.psm1
```diff
@@ -0,0 +1,147 @@
...
+        throw "No apps found in $OutputPackageFolder"
+    } 
+
+    if ($appsInPackageFolder.Count -ne $expectedApplications.Count) {
+        Write-Host "Expected $($expectedApplications.Count) apps, found $($appsInPackageFolder.Count)"
+    }
+
+    $expectedApplications | ForEach-Object {
+        $applicationName = $_.ApplicationName
+        if ($_.IncludeInPackage -and -not (Test-Path -Path "$OutputPackageFolder/Apps/$applicationName")) {
```
mazhelez, line 98: Don't `$expectedApplications` already have `IncludeInPackage` set to _true_?

Build/Scripts/CompileAppInBcContainer.ps1
```diff
@@ -36,7 +36,7 @@ $CICDBuild = $env:GITHUB_WORKFLOW -and ($($env:GITHUB_WORKFLOW).Trim() -eq 'CI/C
 
 if($CICDBuild) {
     # Create the artifacts folder for the app to place in the package
-    . $PSScriptRoot\CreateAppPackageOutput.ps1 -AppProjectFolder $parameters["appProjectFolder"] -BuildMode $appBuildMode -AppFile $appFile -ALGoProjectFolder $currentProjectFolder -IsTestApp:$(!$app)
+    . $PSScriptRoot\Package\CreateAppPackageOutput.ps1 -AppProjectFolder $parameters["appProjectFolder"] -BuildMode $appBuildMode -AppFile $appFile -ALGoProjectFolder $currentProjectFolder -IsTestApp:$(!$app)
```
mazhelez, line 39: Shouldn't that be called based on the info in `Package.json`?

As in... if the value for `IncludeInPackage` is _true_ for the current app, only then run `CreateAppPackageOutput.ps1`.

Build/Scripts/Package/PackNuget.psm1
```diff
@@ -0,0 +1,147 @@
...
+{
+    $appsInPackageFolder = Get-ChildItem -Path "$OutputPackageFolder/Apps"
+    $expectedApplications = Get-ApplicationsForPackage | Where-Object IncludeInPackage
+
+    if($appsInPackageFolder.Count -eq 0) {
+        throw "No apps found in $OutputPackageFolder"
+    } 
+
+    if ($appsInPackageFolder.Count -ne $expectedApplications.Count) {
+        Write-Host "Expected $($expectedApplications.Count) apps, found $($appsInPackageFolder.Count)"
```
mazhelez, line 93: Wouldn't that be a problem? What if `appsInPackageFolder` has more entries than `expectedApplications`?

Build/Scripts/Package/Package.json
```diff
@@ -0,0 +1,24 @@
...
+        "includeInNuget": true
+    },
+    "Performance Toolkit": {
+        "includeInNuget": false,
+        "_comment": "Not enabled yet"
+    },
+    "Permissions Mock": {
+        "includeInNuget": true
+    },
+    "System Application": {
```
mazhelez, line 18: What about _System Application Tests_ and _System Application Test Libraries_? Don't we need them in the package as well?

mazhelez (commented):
One of the points for having "applications in package" settings was to get rid of the custom CompileAppInBcContainer for the _System Application Modules_.

With the current approach, that isn't possible, as far as I can see.

I have two possible approaches in mind:
1. Package.json holds **AL-Go** projects that need to be included in the package. The packages would include all the AL apps from all the AL-Go projects in Package.json.
So, we download _only_ the artifacts from the relevant AL-Go projects and work with them. Each should have a _Package_ folder, etc.
2. Package.json holds all of the AL apps that need to be included in the package. In that case, we download the artifacts for all AL-Go projects and try to look for the apps there. If we find them all, then all good. Otherwise, throw an error as something is off.

You started with approach 2. but the benefit of it is that we don't need to create the _Package_ folders where they wouldn't be needed. That's the missing part, I think.

Build/Scripts/Package/PackNuget.psm1
aholstrup1: Yes, I think so. It was supposed to throw an error here if that's the case

mazhelez (dismissed):
Beautiful PR. Left some non-blocking comments/suggestions.

mazhelez (approved):

PR #22763, opened 2023-03-30
[Shopify] Resolved issue on Shopify Transaction not pulling in Payout Id if it was synchronized before the Payout 

danilovetatek:
Depending on how often you synchronize Payments in BC, there's a high probability of pulling in a Transaction before a Payout for that transaction has been created. In that scenario, the API will return a null value for the Payout Id.

When the Payout is created, this data will be updated within Shopify ecosystem, however it won't replicate to BC because the code is skipping any line that has already been imported.

By adding this fix, it will check whether the transaction's Payout Id value has been populated, and update if available.


Apps/W1/Shopify/app/src/Payments/Codeunits/ShpfyPayments.Codeunit.al
```diff
@@ -189,6 +189,15 @@ codeunit 30169 "Shpfy Payments"
             else
                 if Transaction."Payout Id" > 0 then
                     SinceId := Math.Min(SinceId, Transaction."Payout Id");
+        end else begin
+            Transaction.Get(Id);
+            if Transaction."Payout Id" = 0 then begin
+                PayoutId := JHelper.GetValueAsBigInteger(JTransaction, 'payout_id');
```
onbuyuka, line 195: Does this compile? PayoutId is not defined...

danilovetatek:
Hi @onbuyuka,

My bad, due to internal code diffs I haven't committed the full file. Fixed now.

JesperSchulz:

Build ID: 10814.

PR #22771, opened 2023-03-31
Nextmarker And ListBlobs overload

evjrr:
Return of NextMarker so it is possible to use paging an select more than the first 5000 Blobs

Preformance optimized so collection of 5000 Blobs goes from 45 sec. to  around 9 sec. when using "ABS Container Content" table

Using new overload on ListBlobs will return a dictionary instead of "ABS Container Content" table and the time goes down to 2 sec.

All measurements are based on a local Sandbox Container version 21.5.53619.54870

### DIFF
--- a/Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
+++ b/Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
@@ -74,6 +74,31 @@ codeunit 9053 "ABS Blob Client"
     end;
 
     /// <summary>
+    /// Lists the blobs in a specific container.
+    /// see: https://go.microsoft.com/fwlink/?linkid=2210588
+    /// </summary>    
+    /// <param name="BlobList">Collection of the result (BlobList of [Text, XmlNode]).</param>
+    /// <returns>An operation reponse object</returns>
+    procedure ListBlobs(var BlobList: Dictionary of [Text, XmlNode]): Codeunit "ABS Operation Response"
+    var
+        ABSOptionalParameters: Codeunit "ABS Optional Parameters";
+    begin
+        exit(ABSClientImpl.ListBlobs(BlobList, ABSOptionalParameters));
+    end;
+
+    /// <summary>
+    /// Lists the blobs in a specific container.
+    /// see: https://go.microsoft.com/fwlink/?linkid=2210588
+    /// </summary>    
+    /// <param name="BlobList">Collection of the result (BlobList of [Text, XmlNode]).</param>
+    /// <param name="ABSOptionalParameters">Optional parameters to pass.</param>
+    /// <returns>An operation reponse object</returns>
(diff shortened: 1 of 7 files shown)


evjrr:

Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
```diff
@@ -73,6 +73,31 @@ codeunit 9053 "ABS Blob Client"
         exit(ABSClientImpl.ListBlobs(ABSContainerContent, ABSOptionalParameters));
     end;
 
+    /// <summary>
+    /// Lists the blobs in a specific container.
+    /// see: https://go.microsoft.com/fwlink/?linkid=2210588
+    /// </summary>    
+    /// <param name="BlobList">Collection of the result (BlobList of [Text, XmlNode]).</param>
```
pri-kise, line 80: Could you please add more details about the BlobList Result. e.g. what is the key of the dictionary.

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
```diff
@@ -334,14 +334,24 @@ codeunit 9047 "ABS Optional Parameters"
     end;
 
     /// <summary>
-    /// Specifies the maximum number of blobs to return
+    /// Specifies the maximum number of blobs to be fetch, from 1 to 5000 if not defined it will fetch 5000
+    /// If more than 5000 blobs then use NextMarker to fetch the next set of blobs
```
pri-kise, line 338: Maybe it's worth adding the full details of the documentation to this parameter:

https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs?tabs=azure-ad#uri-parameters
> In certain cases, the service might return fewer results than specified by maxresults, and also return a continuation token.

> Setting maxresults to a value less than or equal to zero results in error response code 400 (Bad Request).

Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
evjrr: Is now updated

Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
evjrr: Is now updated

pri-kise (approved):
looks good to me.
Now maybe someone else to look at it too.

JesperSchulz:
We're on it! Let's get this PR processed!

JesperSchulz (approved):

JesperSchulz:

Build ID: 10731.

JesperSchulz:

Build ID to track: 56517.

PR #22811, opened 2023-04-03
Fix length of Product Group Code in Shopify Connector

bc4all:
The length of the field Product Group Code in the table  "Shpfy Shop Collection Map" was 10. Which gave the error with longer VAT Product Posting Groups.

It was possible to add the VAT Product Posting Group more than 10 chars but there was an issue when renaming the value.

image

I checked where the field is used and looks like there are no references in Shopify Connector that would break anything else.

### DIFF
--- a/Apps/W1/Shopify/app/src/Products/Tables/ShpfyShopCollectionMap.Table.al
+++ b/Apps/W1/Shopify/app/src/Products/Tables/ShpfyShopCollectionMap.Table.al
@@ -40,7 +40,7 @@ table 30128 "Shpfy Shop Collection Map"
             DataClassification = CustomerContent;
         }
 
-        field(4; "Product Group Code"; Code[10])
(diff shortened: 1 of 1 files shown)


pri-kise:
If it would be an public table I'd say this was a breaking change, but since the table is set to internal this should be possible to change.

pri-kise (approved):

dannoe (changes requested):
I am not quite sure, but I think this is still a breaking change (in terms of database schema) and has to be solved another way:
Obsoleting the old field, introducing a new field with a bigger size and then writing upgrade code to transfer the data from the old field to the new field.

Things would look different if the table was a temporary table or only used temporarily. But I do not have enough insight into Shopify to know.

JesperSchulz:
I'll try to process this internally and see what our build system thinks about this 😉

JesperSchulz (approved):

PR #22939, opened 2023-04-06
Encoding.Convert() overloads

glachana:
Adds Convert() overload to Encoding. The goal is to be able to work directly on streams rather than texts


glachana:
@microsoft-github-policy-service agree company="Dynamics International"

glachana:
@microsoft-github-policy-service agree company="Dynamics International"

PR #22973, opened 2023-04-13
Cache baseline packages 

aholstrup1:
Cache baseline packages so they're only downloaded once per job. 

Right now we download the same baseline package every time we invoke CompileAppInBcContainer. This happens 5 times in the "Build Test Framework" task, so this PR would reduce that to downloading it once.


mazhelez (approved):
What about caching the baselines artifact so that they can be reused for different jobs in the same workflow run or even in multiple workflows?

Might be possible using `actions/cache` but for now I don't see a way to integrate in AL-Go. What do you think?

PR #22974, opened 2023-04-13
Various bug fixes in the GP migration

jaymckinney-enavate:
This change fixes the following bugs found in the GP migration.

- Blank screen in the wizard for some users.
- Added better IBAN pre-validation. IBAN codes that are clearly invalid will be skipped instead of going through the BC validation, which halts the migration if this validation fails.
- If migrating Customer or Vendor classes, and the entity has a class that has an account, use this account instead of the default Payables or Receivables account.
- Remove the "Auto Posting Errors" report functionality because of false positives and performance issues that cause timeouts, which halt the migration.

### DIFF
--- a/Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
+++ b/Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
@@ -53,9 +53,11 @@ codeunit 4018 "GP Customer Migrator"
             CopyStr(PostingGroupDescriptionTxt, 1, 20),
             HelperFunctions.GetPostingAccountNumber('ReceivablesAccount')
         );
-        // Set the other two accounts here?
+
         Sender.SetCustomerPostingGroup(CopyStr(PostingGroupCodeTxt, 1, 5));
         Sender.ModifyCustomer(true);
+
+        CreateCustomerPostingGroupIfNeeded(Sender, RecordIdToMigrate);
     end;
 
 #if not CLEAN22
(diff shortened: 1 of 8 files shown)


Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
```diff
@@ -172,6 +176,86 @@ codeunit 4018 "GP Customer Migrator"
                 Sender.SetGeneralJournalLineExternalDocumentNo(CopyStr(MigrationGPCustTrans.DOCNUMBR, 1, 20) + '-' + CopyStr(MigrationGPCustTrans.GLDocNo, 1, 14));
             until MigrationGPCustTrans.Next() = 0;
     end;
+#pragma warning restore AA0207
+
+    local procedure CreateCustomerPostingGroupIfNeeded(var Sender: Codeunit "Customer Data Migration Facade"; RecordIdToMigrate: RecordId)
```
nikolakukrika, line 181: Don't name the codeunit Sender, but Customer Data Migration. It is needed in the subscriber, but it should not be needed here.

nikolakukrika (approved):

Apps/W1/HybridGP/app/src/Migration/Vendors/GPVendorMigrator.codeunit.al
```diff
@@ -497,6 +587,32 @@ codeunit 4022 "GP Vendor Migrator"
         until GPSY06000.Next() = 0;
     end;
 
+    local procedure IsValidIBANCode(IBANCode: Code[100]): Boolean
```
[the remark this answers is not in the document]
jaymckinney-enavate: This is a "pre-validation" check, because the IBAN validation trigger does indeed use CompanyInfomration.CheckIBAN, and it will throw an error if the check fails. The fix here is to not go through the full validation unless the code is fairly confident that the IBAN will pass the full validation.

Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerMigrator.codeunit.al
jaymckinney-enavate: Updated for both the Customer and Vendor migrators. Thanks

JesperSchulz:

Build ID: 10644.

PR #23213, opened 2023-05-04
permissions caption and name

miljance:


Modules/System/Permission Sets/src/PermissionImpl.Codeunit.al
```diff
@@ -156,6 +156,22 @@ codeunit 9864 "Permission Impl."
         end;
     end;
 
+    procedure GetObjectionCaptionAndName(var ExpandedPermission: Record "Expanded Permission"; var ObjectCaption: Text; var ObjectName: Text)
+    var
+        AllObj: Record AllObj;
+    begin
+        if ExpandedPermission."Object ID" <> 0 then begin
+            ExpandedPermission.CalcFields("Object Name");
+            ObjectCaption := ExpandedPermission."Object Name";
```
pri-kise, line 165: I'm a little bit confused by this code.
You're assigne the "Object Name" to the "Object Caption"

miljance: I think this is normal question as I was also confused. It is wrong on other places also.
In this Codeunit there three such functions. The one I created is the third one but all three of them do assign a value from a FlowField called "Object Name" to a ObjectCaption variable. The background is that in all of those three tables the fields names are misleading:
- "Tenant Permission"
- "Metadata Permission"
- "Expanded Permission"

The goal of this pull request is to align the output on all pages. I am attaching the current state as I think you are from Germany and you can instantly spot the inconsistence.

image
image
image

If it happens that there is a better idea how to unify the UI, please let me know. I would be happy to seek for the best possible output.

miljance: While you are there, I would ask you to check one more thing in regard to the page "All Objects with Caption" as it seems it contain further inconsistencies.
I have a W1 SaaS tenant and I open this page using three different languages and then filtering on pages:

1. english
image

2. german (german language app by Microsoft installed)
image


3. serbian (serbian language app by third party installed)
image



The serbian app has all of the page's captions translated but for some reason this page shows the English caption.
I am not sure why this happens for serbian app but not for the german one. The issue is limited to pages only as the table, reports and other types of objects do appear translated properly in this page.

Example - Company Information Page properly translated to serbian when opened:
image

Example with TableData properly properly translated on serbian.
image


This issue does not have a huge impact but I would like to know the root cause even if it would mean that third party translations do not support page language specific captions in this page.

stkillen: To not cause any breaking changes the existing field will stay as it is for now.

Thus, I think the solution proposed here is good - and hopefully we can get around to cleaning up the confusing "Object Name" field in the permission tables at some point.

As for the page translation issue in Object With Caption that you mentioned above I'm not sure. Can you create a new issue on this and tag me? I'll try to get it assigned to someone who can look into this. 

Resolving this conversation.

stkillen (changes requested):
Looks good, will be useful to get the object name on the expanded permissions page as well!

If we can just fix the function name I believe this is ready for merging.

Modules/System/Permission Sets/src/PermissionImpl.Codeunit.al
```diff
@@ -156,6 +156,22 @@ codeunit 9864 "Permission Impl."
         end;
     end;
 
+    procedure GetObjectionCaptionAndName(var ExpandedPermission: Record "Expanded Permission"; var ObjectCaption: Text; var ObjectName: Text)
```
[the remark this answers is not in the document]
stkillen: I see that we actually have a similar named function already in PermissionImpl. Since this is the only comment I'll just fix both internally.

stkillen (approved):
Will merge this and change the function name.

JesperSchulz:

Build ID to track: 58571.

JesperSchulz:

Build ID: 11397.

PR #23225, opened 2023-05-05
Azure Blob Storage, Azure File Share and SharePoint Online Connector Apps

### LINKED ISSUE #22691: [Feature-Contribution] File Accounts
I would like to contribute a file account module that is structured like the mail account module.
I would connect the Azure Blob Storage Module as a standard connector.

The interface provided would provide the following procedures:
```
interface "IFile Access"
{
    procedure DownloadFile(Path: Text; var TempBlob: Codeunit "Temp Blob");
    procedure UploadFile(Path: Text; var TempBlob: Codeunit "Temp Blob");
    procedure FileExists(Path: Text): Boolean;
    procedure DeleteFile(Path: Text): Boolean;
    procedure CreateDirctory(Path: Text): Boolean;
    procedure DirctoryExists(Path: Text): Boolean;
    procedure DeleteDirctory(Path: Text): Boolean;
    procedure ListDirectory(Path: Text): List of [Text];
    procedure CombinePath(ParentPath: Text; ChildPath: Text): Text;
}
```

Is there interest in there?
Which number range should I use?

IceOnly:
This PR contains three new connector apps, to connect  Azure Blob Storage, Azure File Share and SharePoint Online with the New File System Module in the System App.

File System Module PR:
https://github.com/microsoft/BCApps/pull/663

Here are some Screenshots:
image
image
image
image
image
image

Fixes #22691
Fixes [AB#559148](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/559148)


Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageAccount.Page.al
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            field(StorageAccountNameField; Rec."Storage Account Name")
+            {
+                ApplicationArea = All;
+                Caption = 'Storage Account Name';
+                ToolTip = 'Specifies the Azure Storage name.';
+            }
+
+            field(Password; Password)
```
pri-kise, line 39: I know taht currently only shared key is used, but I can imagine, that there might be more options to authenticate in the future.
Maybe we should already prepare this.
It's not a must, but we should at least think about this.

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,414 @@
...
+    /// <param name="Files">A list with all files stored in the path.</param>
+    procedure ListFiles(AccountId: Guid; Path: Text; var FileAccountContent: Record "File Account Content" temporary)
+    var
+        ABSContainerContent: Record "ABS Container Content";
+        ABSBlobClient: Codeunit "ABS Blob Client";
+        ABSOperationResponse: Codeunit "ABS Operation Response";
+        ABSOptionalParameters: Codeunit "ABS Optional Parameters";
+    begin
+        InitBlobClient(AccountId, ABSBlobClient);
+        if (Path <> '') and not Path.EndsWith('/') then
```
pri-kise, line 30: This should be defintily placed in a procedure an reused on all necessary places.

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,414 @@
...
+    end;
+
+    /// <summary>
+    /// Deletes a directory exists on the provided account.
+    /// </summary>
+    /// <param name="AccountId">The file account ID which is used to send out the file.</param>
+    /// <param name="Path">The directory path inside the file account.</param>
+    procedure DeleteDirectory(AccountId: Guid; Path: Text)
+    begin
+
```
pri-kise, line 258: Missing implementation. of DeleteDirectory

Modules/System/File Access/src/Connector/FileConnector.Interface.al
```diff
@@ -0,0 +1,144 @@
...
+/// </summary>
+interface "File Connector"
+{
+    /// <summary>
+    /// Gets a List of Files stored on the provided account.
+    /// </summary>
+    /// <param name="AccountId">The file account ID which is used to get the file.</param>
+    /// <param name="Path">The file path to list.</param>
+    /// <param name="Files">A list with all files stored in the path.</param>
+    procedure ListFiles(AccountId: Guid; Path: Text; var FileAccountContent: Record "File Account Content" temporary);
```
pri-kise, line 17: I'm not sure if we might need an additional optional parameter codeunit.
e.g. we might need to show only the first 100 files for performance reasons instead  of loading all 10.000 files in one call.
I've seen there some pagination for the sharepoint module.

Modules/System/File Access/src/Connector/FileConnector.Interface.al
```diff
@@ -0,0 +1,144 @@
...
+
+    /// <summary>
+    /// Gets a file from the provided account.
+    /// </summary>
+    /// <param name="AccountId">The file account ID which is used to get the file.</param>
+    /// <param name="Path">The file path inside the file account.</param>
+    /// <param name="Stream">The Stream were the file is read to.</param>
+    procedure GetFile(AccountId: Guid; Path: Text; Stream: InStream);
```
pri-kise, line 25: InStream are always passed by reference, but I'd prefer explicit delcaring it as var.

Modules/System/File Access/src/Connector/FileConnector.Interface.al
```diff
@@ -0,0 +1,144 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+/// <summary>
+/// An file connector interface used to creating file accounts and handle external files.
+/// </summary>
+interface "File Connector"
```
pri-kise, line 9: We might need some kind of rename, too.

pri-kise:
Some general thoughts:
I think we need some kind of path serparated from the name added to the file account content.
It's often very important to know the path.

Additionally we might need some additional kind of top level selection before List Directories:
I'm thinking about a webservice for local file system access, there we should be able to select the Drive "C:" or "D:".
For Azure File Share we should be able to select the File Share on a top level before searching directories and files.

Further more if I take a look at the current sharepoint client. I have no idea how we should map this to the current provided 
`interface "File Connector"`.
https://github.com/microsoft/ALAppExtensions/blob/main/Modules/System/SharePoint/src/SharePointClient.Codeunit.al

Sometimes a path isn't enough and we might need more information to download a file.
Maybe this info could be added to the `table 70005 "File Account Content"` table and the `interface "File Connector"` could accept a "File Account Content" as parameter instead of a path....

Maybe someone else with more experience has some ideas how this access can be unified...

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageAccount.Page.al
IceOnly: I think it is easy to implement new auth types without being destructive.

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageConnectorImpl.Codeunit.al
IceOnly: I'm currently thinking about that. Azure Blob Services are Hirachy free. The API and the Azure Storage Explorer are simulating folders based on the Blob File Name. You can not create a folder only a file taht has the Folder in its Name. The https://orchardcore.net/ framework sobled this issue by creating a dummy file that is ignored by its file api.

But I am not sure if i like this "hack".

IceOnly:
> Some general thoughts: I think we need some kind of path serparated from the name added to the file account content. It's often very important to know the path.
> 
> Additionally we might need some additional kind of top level selection before List Directories: I'm thinking about a webservice for local file system access, there we should be able to select the Drive "C:" or "D:". For Azure File Share we should be able to select the File Share on a top level before searching directories and files.
> 
> Further more if I take a look at the current sharepoint client. I have no idea how we should map this to the current provided `interface "File Connector"`. https://github.com/microsoft/ALAppExtensions/blob/main/Modules/System/SharePoint/src/SharePointClient.Codeunit.al
> 
> Sometimes a path isn't enough and we might need more information to download a file. Maybe this info could be added to the `table 70005 "File Account Content"` table and the `interface "File Connector"` could accept a "File Account Content" as parameter instead of a path....
> 
> Maybe someone else with more experience has some ideas how this access can be unified...

Why not listing C: and D: as directory on the top level?
The path is part of the Fiel Acount Content table. It is splitted in Parent Folder and Name.

However, I am not sure if the module should define the path separator. So all Services uses the same path structure and the implentation need to translate if needed.

Modules/System/File Access/src/Connector/FileConnector.Interface.al
IceOnly: I like to keep code as short as possible. I checked the System app and I find no fix rule. TempBlob Impl. uses var. The Image and Email Attachment not :(

Modules/System/File Access/src/Connector/FileConnector.Interface.al
IceOnly: Good point. I will look if i can implement a generic pattern.

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageConnectorImpl.Codeunit.al
pri-kise: You're right I forgot about this part for the Azure Blob Service.
Maybe there should be at least a small comment about this behaviour.

I thought if it might be helpful to add some kind of `GetSupportedMethods` to the Interface to check if one specific feature is possible for this implementation, but this would make i much more complicated....
If it isn't clear what I mean, I was thinking about the pattern that was used for the Barcode Provider. (GetSupportedBarcodeSymbologies in https://github.com/microsoft/ALAppExtensions/blob/main/Modules/System/Barcode/src/Barcode%20Provider/Font/BarcodeFontProvider.Interface.al)

Maybe someone else has an opinion on this specfic topic, too.

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,414 @@
...
+            FileAccountContent.Insert();
+        until ABSContainerContent.Next() = 0;
+    end;
+
+    /// <summary>
+    /// Creates a directory on the provided account.
+    /// </summary>
+    /// <param name="AccountId">The file account ID which is used to send out the file.</param>
+    /// <param name="Path">The directory path inside the file account.</param>
+    procedure CreateDirectory(AccountId: Guid; Path: Text)
```
pri-kise, line 220: After attending @vjekob  session for Testabilit by the use of interface, i think we maybe should split this interface into more interfaces. One General and at least one for directories.

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageAccount.Table.al
```diff
@@ -0,0 +1,76 @@
...
+            if IsolatedStorage.Delete(Rec."Password Key") then;
+    end;
+
+    [NonDebuggable]
+    procedure SetPassword(Password: Text)
+    begin
+        if IsNullGuid(Rec."Password Key") then
+            Rec."Password Key" := CreateGuid();
+
+        if not IsolatedStorage.Set(Format(Rec."Password Key"), Password, DataScope::Company) then
```
PeterConijn, line 66: Since we store the key in the database, would it be an idea to add a "seed" to the Set/Get for the isolated storage, so that even if someone got the password key, they would not be able to access the password, since they do not have the seed?

The seed can be (part of) a guid or a number (as text) in a non-debuggable local function (or even a separate non-debuggable, internal codeunit).

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageAccount.Table.al
```diff
@@ -0,0 +1,76 @@
...
+        {
+            Caption = 'Storage Account Name';
+        }
+        field(4; "Container Name"; Text[2048])
+        {
+            Caption = 'Container Name';
+        }
+        field(8; "Password Key"; Guid)
+        {
+            DataClassification = SystemMetadata;
```
PeterConijn, line 38: Should we add a non-debuggable to this?

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageAccountWizard.Page.al
```diff
@@ -0,0 +1,161 @@
...
+            }
+        }
+    }
+
+    var
+        BlobStorageAccount: Record "File Account";
+        MediaResources: Record "Media Resources";
+        BlobStorageConnectorImpl: Codeunit "Blob Storage Connector Impl.";
+        [NonDebuggable]
+        [InDataSet]
```
PeterConijn, line 138: Not sure, but are these still needed? I thought InDataSet was no longer used in web clients.

Apps/W1/File - Azure BLOB Storage Connector/app/src/BlobStorageAccountWizard.Page.al
```diff
@@ -0,0 +1,161 @@
...
+        Password: Text;
+        IsNextEnabled: Boolean;
+        TopBannerVisible: Boolean;
+
+    trigger OnOpenPage()
+    begin
+        Rec.Init();
+        Rec.Insert();
+
+        if MediaResources.Get('ASSISTEDSETUP-NOTEXT-400PX.PNG') and (CurrentClientType() = ClientType::Web) then
```
PeterConijn, line 148: Can we place this in a locked label or in a Get-procedure?

Modules/System/File Access/src/Account/FileAccountImpl.Codeunit.al
```diff
@@ -0,0 +1,242 @@
...
+
+        if ConnectorLogoBase64 = '' then
+            exit;
+        if not FileConnectorLogo.Get(FileAccount.Connector) then begin
+            TempBlob.CreateOutStream(OutStream);
+            Base64Convert.FromBase64(ConnectorLogoBase64, OutStream);
+            TempBlob.CreateInStream(InStream);
+            FileConnectorLogo.Connector := FileAccount.Connector;
+            FileConnectorLogo.Logo.ImportStream(InStream, StrSubstNo(ConnectorLogoDescriptionTxt, FileAccount.Connector));
+            if FileConnectorLogo.Insert() then;
```
PeterConijn, line 140: "if (not) Insert then" can cause performance issues when there is a PK violation. I usually use an if Get then Modify else Insert pattern.

PeterConijn (changes requested):
Looks cool! I have a few small performance and security concerns, which I have addressed, and some minor other stuff, but nothing major. I'm excited for this.

JesperSchulz:
Let me amplify this on Twitter to get your initial question answered: "before I go round the whole thing, I'm interested in whether there is any interest in the module at all". Let's see what people have to say! 😊

tinfister:
I support every effort to improve file handling.

dNsl9r:
Wonderful! Please merge this 👍🏻

miljance:
Easily overlooked feature but a needed one. Huge development effort that should not be thrown away.
Perhaps I will find some time to invest in contribution from my side but cannot make promises at the moment.

TheDoubleH:
Bravo! Love seeing more 'helper' features/functions that eventually will benefit all of us!

JesperSchulz:
@IceOnly, this PR seems to have stagnated a little (which is perfectly fine). I was just wondering if you want to push this forward at some point, or if you've changed your mind / cannot find the time? There seems to be plenty of interest for this module in the community!

IceOnly:
> @IceOnly, this PR seems to have stagnated a little (which is perfectly fine). I was just wondering if you want to push this forward at some point, or if you've changed your mind / cannot find the time? There seems to be plenty of interest for this module in the community!

It is the time. I will try to find some free time to make the last changes.

This are my open points:
- I am struggling with the wist from [pri-kise](https://github.com/microsoft/ALAppExtensions/pull/23225#discussion_r1236829294) to split the Main Interface in logic interfaces like one for File, one for Directory and one for Account access. I like the Idea to implement one Interface and you are done.
- I will try to implement a realy simple pagination interface. The Most file services uses realy different methods for getting the next batch. So i will only alow the implemntation to result a pagination codeunit that can store the inforamtion it needs to get the next batch. How big a batch is can only be defined by the implementation.

JesperSchulz:
Feel free to do it at your own pace. I didn't mean to stress you. I was just curious if you still intend to continue work. If you do, we'll just leave this PR open. No problem whatsoever.

IceOnly:
Should now contain everything we need. Next I will upgrade the code to BC23, add Namespaces and the test units.

JesperSchulz:
Hi @IceOnly! Looks like your PR slowly but steadily is going places 🥳 Unfortunately, it's not in a quality yet, where we can merge it, and we hence have to port your PR from this ALAppExtensions repository to the new BCApps repository (https://github.com/microsoft/BCApps), as we'll soon be deleting the System Application from this repo. A lot of things are better on BCApps, so this should be good. Ideally, I'd of course want you to create the new PR on BCApps, as you are the author of this work. Could I persuade you to do so? Here's an example of a similar move: https://github.com/microsoft/BCApps/pull/341.
When you have created the new PR, you may close this one.
Sorry for the inconvenience and thanks for your help!

IceOnly:
> Hi @IceOnly! Looks like your PR slowly but steadily is going places 🥳 Unfortunately, it's not in a quality yet, where we can merge it, and we hence have to port your PR from this ALAppExtensions repository to the new BCApps repository (https://github.com/microsoft/BCApps), as we'll soon be deleting the System Application from this repo. A lot of things are better on BCApps, so this should be good. Ideally, I'd of course want you to create the new PR on BCApps, as you are the author of this work. Could I persuade you to do so? Here's an example of a similar move: [microsoft/BCApps#341](https://github.com/microsoft/BCApps/pull/341). When you have created the new PR, you may close this one. Sorry for the inconvenience and thanks for your help!

I will create the new PR in the next days :)

JesperSchulz:
Uh, I just noticed that this PR consists of both a 1st party app, and a System Application module. The new BCApps repository doesn't support 1st party apps yet 😟 Solution: This PR must be split. The System Application part moves, while the 1st party App part stays here.
Does that work for you?

IceOnly:
Yes, no problem.

IceOnly:
@JesperSchulz I moved the the system app part to BCApps https://github.com/microsoft/BCApps/pull/663

JesperSchulz:
Excellent! I've renamed this PR to avoid confusion. Feel free to rename it once more 😊

IceOnly:
@JesperSchulz I need an three Object Ranges for this Apps. Which one can I used? The highest object type count is three objects.
I'm also not sure which namespace to use. Or simply use none, as in the email apps?

Apps/W1/File - SharePoint Connector/app/src/SharePointAccount.Table.al
```diff
@@ -0,0 +1,86 @@
...
+        }
+        field(4; "SharePoint Url"; Text[2048])
+        {
+            Caption = 'SharePoint Url';
+        }
+        field(5; "Base Relative Folder Path"; Text[2048])
+        {
+            Caption = 'Base Relative Folder Path';
+        }
+        field(6; "Tenant Id"; Text[36])
```
duichwer, line 38: Shouldn't we use Guid Fields for those values that are clearly Guides?

Apps/W1/File - Azure File Service Connector/app/src/FileShareAccount.Table.al
```diff
@@ -0,0 +1,82 @@
...
+    begin
+        if IsNullGuid(Rec."Secret Key") then
+            Rec."Secret Key" := CreateGuid();
+
+        if not IsolatedStorage.Set(Format(Rec."Secret Key"), Secret, DataScope::Company) then
+            Error(UnableToSetSecretMsg);
+    end;
+
+    [NonDebuggable]
+    procedure GetSecret(SecretKey: Guid) Secret: SecretText
```
duichwer, line 77: If we only use SecretText in the getter then the NonDebuggable shouldn't be needed @darjoo

Apps/W1/File - SharePoint Connector/app/src/SharePointConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,444 @@
...
+using System.Integration.Sharepoint;
+using System.Utilities;
+
+codeunit 80300 "SharePoint Connector Impl." implements "File System Connector"
+{
+    Access = Internal;
+    Permissions = tabledata "SharePoint Account" = rimd;
+
+    var
+        ConnectorDescriptionTxt: Label 'Use SharePoint to store and retrieve files.';
```
duichwer, line 17: MaxLength property is missing here

Apps/W1/File - Azure File Service Connector/app/src/FileShareAccount.Table.al
darjoo: Yes, you are correct. `[NonDebuggable]` property isn't needed when we're using `SecretText`.

Apps/W1/File - SharePoint Connector/app/src/SharePointAccount.Table.al
```diff
@@ -0,0 +1,84 @@
...
+/// <summary>
+/// Holds the information for all file accounts that are registered via the SharePoint connector
+/// </summary>
+table 80300 "SharePoint Account"
+{
+    Access = Internal;
+
+    Caption = 'SharePoint Account';
+
+    fields
```
pri-kise, line 17: @IceOnly could you maybe add the 2nd Authentication Type for SharePoint that was added recently .
This would require a few new Fields: Enum SharePointAuthType , a "Certificate Password Key" and a Certificate Password (Blob)

See: https://github.com/microsoft/BCApps/pull/705

Apps/W1/File - SharePoint Connector/app/src/SharePointConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,444 @@
...
+    /// Gets a file from the provided account.
+    /// </summary>
+    /// <param name="AccountId">The file account ID which is used to get the file.</param>
+    /// <param name="Path">The file path inside the file account.</param>
+    /// <param name="Stream">The Stream were the file is read to.</param>
+    procedure GetFile(AccountId: Guid; Path: Text; Stream: InStream)
+    var
+        SharePointFile: Record "SharePoint File";
+        SharePointClient: Codeunit "SharePoint Client";
+        TempBlob, TempBlob2 : Codeunit "Temp Blob";
```
pri-kise, line 66: TempBlob2 is unused.

@IceOnly  regarding your issue with the empty stream.
Take a look at the following issue: [Stream](https://github.com/microsoft/AL/issues/7329) 

Why aren't you directly using the parameter `Stream` in the procedure `DownloadFileContentByServerRelativeUrl`?

Apps/W1/File - SharePoint Connector/app/src/SharePointConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,444 @@
...
+    /// </summary>
+    /// <param name="AccountId">The file account ID which is used to send out the file.</param>
+    /// <param name="SourcePath">The source file path.</param>
+    /// <param name="TargetPath">The target file path.</param>
+    procedure CopyFile(AccountId: Guid; SourcePath: Text; TargetPath: Text)
+    var
+        TempBlob: Codeunit "Temp Blob";
+        Stream: InStream;
+    begin
+        TempBlob.CreateInStream(Stream);
```
pri-kise, line 114: This CreateInStream statement shouldn't be need here.

tinestaric (commented):
Great work on these apps! I hope they get merged in soon, it's exactly what all partners have been implementing in so many different ways!

Apps/W1/File - Azure Blob Service Connector/app/src/BlobStorageConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,488 @@
...
...AF2JACgYsWNAMYOZlUfPwQyAgA7EgBQMZuDQoeMAEAxirsICLNMBwAUo7gAaDQdzKhBgY8EKm4EuD4Vl0nQipVoihuri1tttw+u67sE2BNHC/xsFxcAh9IwvjG3EosFtkuwG8Mc8fBwOT7WlLUpSESBI0BExJ1pIW6fvzFezZfi5Ki8Z6nD5bqhGcYdzUJcF6m49j+i0L0BI/67n9odaSHuGJaXmjArihsBgO4UtzMQ0B0dAFRMAEDFirwLAHRDBwAVEwBQMSMAVEwHABUTAFCxokeA1zfW4tyli73WAFdjcTCMwweu7+39p63hIv8L8PKbp+Onrx6P9S1/BOLaN0xNPPah2+PLh97fdynbFNcBnL24GT965a8xcgGSGbGVx/Hz116JI8vv6rwbuOaeCfjSG6csfmbSC6dP9l3CNsUFwJubG32XAHviXxvrfZewTXEbg4zGvv2ZTVt57LHgQDkEAFSsuLsAMMuMAEAxBABUzAgAHck5F7emdABQMQEAFStuBCitRYI2lfb51gFAxYrrAGCWlbamdABQMQEAFStub8DSWiRoi98BAEURAFAxdwGgQ6WtKR0AVKy4DiBHWQkJbUlR3kX14jqA5bn5vkuAPbE8v9B3CdsUFwCH9/e3iwrspcP7l/suYZviRoCPr7wnvnjTofhjgc9Qh906unJD3HPwUHEjQJFbg331fR+Jz954c/xz7Xycu7TZdzmwa0vDuXjv/uU4cmCl71L+ryIDICLi1n0H4tZ9B/ouA2ZacSMA0J5r7i4A0J3itgYD2qMDAHYkAKBiLgLCDDMCADsSAFAxIwBUTAcAFZv2U+BzOedbOqkEaF3O+dyk49M6gL+1WAvQsZTSxDU8MQBSSsfbLQfo2MQ1PO0i4M/G4/G3U0qL7dYE7LWc81rTNE9POmdiB7C6unq6aZon2y0L6ELTNE+urq6ennTO1OcBrK+v/2BxcfHxiDjaWmXAXntldXX1h9NOmnob8NFHHx0NBoMvRcRzrZQF7LXnBoPBvSmlqT/kuaL/+h47duyxiHgiIu7cbWXAnjkeEd+9//77f3m5L9j1n/2fffbZI03THNrt64Grl3POw+Hw5H333fePvmsBAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIA6/AeyXHTtib4x0gAAAABJRU5ErkJggg==', Locked = true;
```
pri-kise, line 19: Since the platform now supports resources we should get away with the base64 and instead uptake this new feature.
https://learn.microsoft.com/en-us/dynamics365/release-plan/2024wave2/smb/dynamics365-business-central/package-resources-extensions-access-al

IceOnly:
@JesperSchulz I still need three id ranges and app logos here. Is anything else necessary for a review to start?

Apps/W1/File - SharePoint Connector/app/src/SharePointAccount.Table.al
```diff
@@ -11,6 +11,7 @@ namespace System.ExternalFileStorage;
 table 80300 "SharePoint Account"
 {
     Access = Internal;
```
tinestaric, line 13: I'd like to challenge the idea of these tables being internal. 

One simple use case I'm thinking of is storing a file to SharePoint and then giving the user a URL where they can access/find this file. With the table being internal, I have no option of building up the file URL without storing the SharePoint URL somewhere else again.

What are your thoughts on this?

I could also imagine that being an additional interface procedure, and implementations have to define how the fully qualified file identifier is built...

Apps/W1/File - SharePoint Connector/app/src/SharePointConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,439 @@
...
+        NewFileShareAccount.Insert();
+
+        TempFileAccount."Account Id" := NewFileShareAccount.Id;
+        TempFileAccount.Name := NewFileShareAccount.Name;
+        TempFileAccount.Connector := Enum::"Ext. File Storage Connector"::"SharePoint";
+    end;
+
+    local procedure InitSharePointClient(var AccountId: Guid; var SharePointClient: Codeunit "SharePoint Client")
+    var
+        FileShareAccount: Record "SharePoint Account";
```
tinestaric, line 377: This should probably be SharePointAccount. There's a few other places as well.

JesperSchulz:
@IceOnly, let's get this on the road! Please re-ID to IDs from 4560 and up (as needed). I don't have any app icon (for now, anyway). Let's push forward without those.

Apps/W1/File - SharePoint Connector/app/src/SharePointAccount.Table.al
JesperSchulz: If there's a valid reason for tables to be exposed, there is no issue at all in doing so. In this case it seems to make sense to have the table publicly accessible, so let's do that.

Apps/W1/File - Azure Blob Service Connector/app/app.json
```diff
@@ -0,0 +1,30 @@
+{
+    "id":  "c9ce86fe-cb70-4b79-be03-d21856b1a4ca",
+    "name":  "File - Azure Blob Service Connector",
```
JesperSchulz, line 3: Let's align the naming:

```suggestion
    "name":  "External File Storage - Azure Blob Service Connector",
```

Apps/W1/File - Azure Blob Service Connector/app/app.json
```diff
@@ -0,0 +1,30 @@
...
+    "publisher":  "Microsoft",
+    "brief":  "Enable all users to use a Azure Blob Storage to save files from Business Central.",
```
JesperSchulz, line 5: It's not really what this does. It adds the capability to do file operations on the Azure Blob Storage through the External File Storage module. This brief (and the other two) need a bit more work.

Apps/W1/File - Azure Blob Service Connector/app/app.json
```diff
@@ -0,0 +1,30 @@
...
+    "description":  "This app enables all users to store files in the smae azure blob storage in Business Central.",
+    "version":  "25.0.0.0",
```
JesperSchulz, line 7: All versions should be 26 now.

Apps/W1/File - Azure Blob Service Connector/app/app.json
```diff
@@ -0,0 +1,30 @@
...
+    "url":  "https://go.microsoft.com/fwlink/?linkid=724011",
+    "logo":  "ExtensionLogo.png",
+    "application":  "25.0.0.0",
+    "internalsVisibleTo":  [],
+    "dependencies":  [],
+    "screenshots":  [],
+                    "platform":  "25.0.0.0",
+    "idRanges":  [
+                     {
+                         "from":  80100,
```
JesperSchulz, line 20: As mentioned in other comment, let's use 4560 and upwards for the 3 new apps (e.g. 4560 for this, 4570 for the other and 4580 for the last one).

Apps/W1/File - Azure Blob Service Connector/app/app.json
```diff
@@ -0,0 +1,30 @@
...
+                         "to":  80199
+                     }
+                 ],
+    "resourceExposurePolicy":  {
+                                   "allowDebugging":  false,
```
JesperSchulz, line 25: You could allow debugging, if you make sure to use secret text etc. where necessary.

Apps/W1/File - Azure Blob Service Connector/app/permissions/FileStorageAdminBlobStor.PermissionSetExt.al
```diff
@@ -0,0 +1,11 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.ExternalFileStorage;
+
+permissionsetextension 80100 "File Storage - Admin - Blob Stor." extends "File Storage - Admin"
```
JesperSchulz, line 8: Darn, we should have named the permission set External File Storage. File Storage suggests that you through this permission set can control the access to all files.

JesperSchulz (changes requested):
Did a first round of review, mostly looking at the app setups. Let's fix these things first, and then I'll take a 2nd round, reviewing the actual code in the apps :-)

Apps/W1/File - Azure Blob Service Connector/app/permissions/FileStorageAdminBlobStor.PermissionSetExt.al
JesperSchulz: This will have to get fixed in BCApps, if we want to change this. Or we just close both eyes and live with it 😉

JesperSchulz (changes requested):
Coming along nicely! I couldn't find anything serious, but would suggest to revisit the permissions.

Apps/W1/External File Storage - Azure Blob Service Connector/app/src/BlobStorageAccount.Page.al
```diff
@@ -0,0 +1,91 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.ExternalFileStorage;
+
+/// <summary>
+/// Displays an account that was registered via the Blob Storage connector.
+/// </summary>
+page 4560 "Blob Storage Account"
```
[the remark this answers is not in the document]
pri-kise: @JesperSchulz
after your suggestion I re-checke this PR again and compared with other w1 first party apps.

Now I'm questioning if the namespace is correct.

I currently can find two patterns:
- The mail connector apps are using the same top level namespace as the System App .
- The new e-documents connector apps are adding another level for the specific connector like Avalara.

This PR follows the approach with the same namespace.
Is this correct? How can we ensure that we won't add an object with an conflicting name in the system application layer?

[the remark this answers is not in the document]
JesperSchulz: Build would fail, if an object with the same name exists in the same namespace. A problem only occurs if an object with the same name exists in a different namespace - but we're looking into that...

I think it's fine to have the same namespace for the System App module and the connectors. Makes sense to me.

JesperSchulz:
@IceOnly, could you pull from main and then we run CI once more.

JesperSchulz:
It actually looks like your PR failed on some compilation issues related to your changes. Could you take a look?

tinestaric:
I tried using the connectors with our apps, and I must again say that this is a great addition!
One thing that I've noticed that error handling could be improved, but as we agreed earlier, that can be addressed in a later PR.

One final thing I'd try to get sorted out for this PR is to make the Blob Storage and File Share tables public as well. The use case is that I'd store files in either of those storage types from BC and then send the file URI to an external service that can pull it (File based integrations). Right now I have to keep Storage Account, Container, and File Share Name in a separate record to build up a URI.

IceOnly:
> I tried using the connectors with our apps, and I must again say that this is a great addition! One thing that I've noticed that error handling could be improved, but as we agreed earlier, that can be addressed in a later PR.
> 
> One final thing I'd try to get sorted out for this PR is to make the Blob Storage and File Share tables public as well. The use case is that I'd store files in either of those storage types from BC and then send the file URI to an external service that can pull it (File based integrations). Right now I have to keep Storage Account, Container, and File Share Name in a separate record to build up a URI.

I made the other setup tables global as well.

IceOnly:
> Found one more case that would probably need to be dealt with within the connector app. Copying a production environment to a sandbox (mabye also copy companies?). We don't want changes in a sandbox environment to be reflected in production at all, so during the copy part, the settings or at the very least the credentials should be dropped.

Good idea. I added a disabled field to all account tables. It will be set automaticly on enviroment copy.

JesperSchulz:
I am super eager to get this one the road. Community, could we get some sign-offs? Once we've got those (and we've fixed our uptake issues 🤦‍♂️),  we'll move this PR into the next stage! My goal is to have this completed before end of this month 😊

tinestaric (approved):

Apps/W1/External File Storage - Azure Blob Service Connector/app/src/ExtBlobStorageAuthType.Enum.al
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.ExternalFileStorage;
+
+enum 4560 "Ext. Blob Storage Auth. Type"
+{
+    Access = Internal;
+
+    value(0;
```
pri-kise, line 12: This line break looks wrong

Apps/W1/External File Storage - Azure File Service Connector/app/src/ExtFileShareAccountWizard.Page.al
```diff
@@ -0,0 +1,155 @@
...
+            {
+                Editable = false;
+                ShowCaption = false;
+                Visible = TopBannerVisible;
+                field(NotDoneIcon; MediaResources."Media Reference")
+                {
+                    Editable = false;
+                    ShowCaption = false;
+                    ToolTip = ' ';
+                    Caption = ' ';
```
pri-kise, line 38: If ShowCaption = false then you shouldn't have to set a Caption.
Furthermore the ToolTip here could be locked, since we don't want a translation here.

Apps/W1/External File Storage - Azure File Service Connector/app/src/ExtFileShareAccountWizard.Page.al
```diff
@@ -0,0 +1,155 @@
...
+        TopBannerVisible: Boolean;
+
+    trigger OnOpenPage()
+    var
+        AssistedSetupLogoTok: Label 'ASSISTEDSETUP-NOTEXT-400PX.PNG', Locked = true;
+    begin
+        Rec.Init();
+        Rec.Insert();
+
+        if MediaResources.Get(AssistedSetupLogoTok) and (CurrentClientType() = ClientType::Web) then
```
pri-kise, line 142: @JesperSchulz do we already have to prepare for this upcoming change when the Images are removed from the MediaResources table?
https://github.com/microsoft/BCApps/pull/2391

Apps/W1/External File Storage - Azure File Service Connector/app/src/ExtFileShareAccountWizard.Page.al
```diff
@@ -0,0 +1,155 @@
...
+                end;
+            }
+
+            action(Next)
+            {
+                Caption = 'Next';
+                Image = NextRecord;
+                Enabled = IsNextEnabled;
+                InFooterBar = true;
+                ToolTip = 'Next';
```
pri-kise, line 115: The ToolTip for Next and Back could be improved.
e.g. :
- `ToolTip = 'Move to the previous step.';`
- ToolTip = 'Move to the next step.';
 
I know that other wizards are there bad, too.

pri-kise: I reviewed this again. 
Isn't the Naming here wrong?
The first action cancels the wizard and the 2nd Finish the wizards..?

Apps/W1/External File Storage - SharePoint Connector/app/permissions/ExtSharePointEdit.PermissionSet.al
```diff
@@ -0,0 +1,18 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.ExternalFileStorage;
+
+permissionset 4582 "Ext. SharePoint - Edit"
+{
+    Assignable = false;
```
pri-kise, line 10: Whats' the use for permission Sets that are Public but not assignable?
Is this really on purpose?

Apps/W1/External File Storage - SharePoint Connector/app/src/ExtSharePointAccount.Table.al
```diff
@@ -0,0 +1,94 @@
...
+/// </summary>
+table 4580 "Ext. SharePoint Account"
+{
+    DataClassification = CustomerContent;
+    Caption = 'SharePoint Account';
+
+    fields
+    {
+        field(1; "Id"; Guid)
+        {
```
pri-kise, line 19: I would suggest to set the property `AllowInCustomizations` to `Never` for the Id, since this is only used internally and shouldn't be exposed to pages by profile customization. (This applies to the other new tables, too).

Apps/W1/External File Storage - SharePoint Connector/app/src/ExtSharePointConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,458 @@
...
+
+        if TempAccount."Base Relative Folder Path" = '' then
+            exit(false);
+
+        exit(true);
+    end;
+
+    internal procedure CreateAccount(var AccountToCopy: Record "Ext. SharePoint Account"; Password: SecretText; var TempFileAccount: Record "File Account" temporary)
+    var
+        NewileShPointareAccount: Record "Ext. SharePoint Account";
```
pri-kise, line 364: nit: The name of the variable `NewileShPointareAccount` could be fixed.

Apps/W1/External File Storage - Azure Blob Service Connector/app/Entitlements/ExtBlobStorageConnector.Entitlement.al
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.ExternalFileStorage;
+
+entitlement "Ext. Blob Storage Connector"
+{
+    Type = Implicit;
+
+    ObjectEntitlements = "Ext. Blob Stor. - Objects";
```
pri-kise, line 12: in the other entitlements the edit permission set is included.
In this entitlement on the object permission set.

For what do we need thos entitlements here ..?

Apps/W1/External File Storage - Azure Blob Service Connector/app/permissions/FileStorageAdminExtBlobStorage.PermissionSetExt.al
```diff
@@ -0,0 +1,11 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.ExternalFileStorage;
+
```
pri-kise, line 7: Will this permissionset extension allow partners to set this up as delegated admin agent? (@JesperSchulz could you maybe forward this to someone with knowledge on this topic?)

Apps/W1/External File Storage - Azure File Service Connector/app/src/ExtFileShareAccount.Page.al
```diff
@@ -0,0 +1,68 @@
...
+    var
+        SecretEditable: Boolean;
+        [NonDebuggable]
+        Secret: Text;
+
+    trigger OnOpenPage()
+    begin
+        Rec.SetCurrentKey(Name);
+
+        if not IsNullGuid(Rec."Secret Key") then
```
pri-kise, line 60: Shouldn't the trigger `trigger OnAfterGetRecord()` be used for checks on the record?

I would suggest to move the code
```al
if not IsNullGuid(Rec."Secret Key") then
    Secret := '***';
```
to the `OnAfterGetRecord` trigger.
(This applies for the other external file storages, too.)

pri-kise (changes requested):
I added some comments.
You don't necessary need to change everything, but I wanted to give you my hones opinion on the current PR.

nit: if you have the time or use the known VS Code extensions, then you could sort the properties of the pages.

JesperSchulz:
None of the open comments prevent us from creating a first iteration of the connector apps. I will open an internal PR to get reviews from Microsoft started.

JesperSchulz:
@IceOnly, so far, so good. Build on GitHub is passing, internal build is now passing - in theory, I could submit the code!

However, with 0 tests I will probably have a hard time getting any sign-offs. I get that these apps only connect the tested module to tested API helper modules, but we would probably still need tests to verify the basic capabilities. Ideally we'd have mocks, but those aren't ready for showtime yet.

You don't happen to have a test or 10 lying around for this code, do you?

IceOnly:
> @IceOnly, so far, so good. Build on GitHub is passing, internal build is now passing - in theory, I could submit the code!
> 
> However, with 0 tests I will probably have a hard time getting any sign-offs. I get that these apps only connect the tested module to tested API helper modules, but we would probably still need tests to verify the basic capabilities. Ideally we'd have mocks, but those aren't ready for showtime yet.
> 
> You don't happen to have a test or 10 lying around for this code, do you?

@Jesper I added some test. I selected some random ids (not fully random i prefixed the id Ranges of the Apps with 14), maybe I have to move the id of the test app.

JesperSchulz:
> > @IceOnly, so far, so good. Build on GitHub is passing, internal build is now passing - in theory, I could submit the code!
> > However, with 0 tests I will probably have a hard time getting any sign-offs. I get that these apps only connect the tested module to tested API helper modules, but we would probably still need tests to verify the basic capabilities. Ideally we'd have mocks, but those aren't ready for showtime yet.
> > You don't happen to have a test or 10 lying around for this code, do you?
> 
> @jesper I added some test. I selected some random ids (not fully random i prefixed the id Ranges of the Apps with 14), maybe I have to move the id of the test app.

I have changed two IDs, which were conflicting. The other 4 seem fine. Let me run your changes in our internal pipeline and let's see what comes out of that! We're getting there 😊

Apps/W1/External File Storage - Azure Blob Service Connector/app/src/ExtBlobStoConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,508 @@
...
+        NotRegisteredAccountErr: Label 'We could not find the account. Typically, this is because the account has been deleted.';
+        MarkerFileNameTok: Label 'BusinessCentral.FileSystem.txt', Locked = true;
+
+    /// <summary>
+    /// Gets a List of Files stored on the provided account.
+    /// </summary>
+    /// <param name="AccountId">The file account ID which is used to get the file.</param>
+    /// <param name="Path">The file path to list.</param>
+    /// <param name="FilePaginationData">Defines the pagination data.</param>
+    /// <param name="Files">A list with all files stored in the path.</param>
```
JesperSchulz, line 31:
```suggestion
    /// <param name="TempFileAccountContent">A list with all files stored in the path.</param>
```

JesperSchulz (changes requested):
Some minor issues left before we can merge :-)

Apps/W1/External File Storage - Azure Blob Service Connector/app/src/ExtBlobStoConnectorImpl.Codeunit.al
JesperSchulz: Same issue 2 other places - but don't worry, I've fixed those in my internal PR.

tinestaric:
We found another use case that would be nice to cover, but doesn't block this PR. This is more just dropping the idea for the future.

Using a Certificate to authenticate the Sharepoint connector. I see there already is a helper function for it in the SharepointAuth, so it's mainly about a way to store the certificate and password.

JesperSchulz (approved):

aholstrup1 (approved):

Apps/W1/External File Storage - SharePoint Connector/app/src/ExtSharePointConnectorImpl.Codeunit.al
```diff
@@ -0,0 +1,458 @@
...
+    /// <returns>Returns true if the directory exists</returns>
+    procedure DirectoryExists(AccountId: Guid; Path: Text): Boolean
+    var
+        SharePointFolder: Record "SharePoint Folder";
+        SharePointClient: Codeunit "SharePoint Client";
+    begin
+        InitPath(AccountId, Path);
+        InitSharePointClient(AccountId, SharePointClient);
+        if SharePointClient.GetSubFoldersByServerRelativeUrl(Path, SharePointFolder) then
+            exit;
```
tinestaric, line 235: @JesperSchulz Since this is already merged, you want to handle that on your internal repo, but while testing today, I noticed that this function never actually returns true.

tinestaric: After some more digging around, the issue is a bit bigger

The GetFolderByServerRelativeUrl('RelativePath')/Folders API on Sharepoint's side doesn't throw and error if the path doesn't exist, it simply returns an empty array of subfolders.
image

The GetFolderByServerRelativeUrl('RelativePath')/Exists would be the proper call, but we don't have that abstraction for that in the SharePoint Client...

JesperSchulz: @IceOnly, please create issue and PR. We've got plenty of time to address these bugs before showtime in April 😊
@tinestaric, thank you for the elaborate test! Nice to get these issues ironed out prior to release! 🙏

PR #23227, opened 2023-05-05
default values for Permissions with Type=TableData

miljance:
**🕮 Describe the issue**

When user creates new permission for a certain permission set, the default values for Table Data are to generous. Only "read permission" should be defaulted to Yes during record initialization. Currently insert, modify and delete are also defaulted to Yes.

**🔧 To Reproduce**

1. Go to Permission Sets
2. Click on any user defined permission set or create a new one.
3. Open page permissions
4. Click on "New Line" on permissions subpage.
5. enter some table id, for example 3 (Payment Terms)
6. Observe that all permission types - read, insert, modify and delete are defaulted to Yes

**📣 Expected behavior**

Only Read Permission should be defaulted to Yes and the user can set other permission types to Yes if needed. Similar to the function "Add Read Permission to Related Tables"

**Screenshots**

image


JesperSchulz (approved):

JesperSchulz:

Build ID: 10905.

JesperSchulz:

Build ID to track: 57721.

PR #23231, opened 2023-05-05
remove filters on TableData when objects for permission are being selected

miljance:
Our consultants reported that they do not see a value to have filter only on TableData when selecting a group of objects for permissions. Therefore, this pull request with hope that the change can be accepted. If not please let me know so that I can inform the consulting department.


JesperSchulz:

Build ID: 10868.

PR #23310, opened 2023-05-10
[Shopify] Shpfy access level

louagej:
### Access Property

In order to promote extensibility of the Shopify connector, removed Access = Internal Property

### DIFF
--- a/Apps/W1/Shopify/README.md
+++ b/Apps/W1/Shopify/README.md
@@ -12,6 +12,45 @@ We're opening for extensibility for specific scenarios, based on feedback from o
 
 Find overview of objects expected to become public in future releases:
 
+### Access Property
+
+In order to promote extensibility of the Shopify connector, changed Access Property for the Following objects:
+
+- ShpfyCommunicationEvents.Codeunit.al
+- ShpfyGraphQLQueries.Codeunit.al
+- ShpfyCue.Table.al
+- ShpfyInitialImportLine.Table.al
+- ShpfyMetafield.Table.al
+- ShpfyShop.Table.al
+- ShpfySynchronizationInfo.Table.al
+- ShpfyTemplatesWarnings.Table.al
+- ShpfyCustomerAddress.Table.al
+- ShpfyCustomerTemplate.Table.al
+- ShpfyProvince.Table.al
+- ShpfyTaxArea.Table.al
+- ShpfyGiftCard.Table.al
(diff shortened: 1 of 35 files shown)


pri-kise:
@louagej  you removed the Access Property from serveral tables.
In the issue you only mention **Table 30107 "Shpfy Customer Template"**.

Could you please elaborate these changes or revert the other changes.

The shopify module is made on purpose as internal to provide access only for specified processes.

If you need access to specific table or objects then please provide an exmaple or a business need for the relevant objects.

louagej:
@pri-kise ,

If tables are set to Access = Internal, you can't even use the as a variable in your code, or subscribe to any insert, modify or delete.
That's why I think this access level should be removed.
Any other isv trying to develop custom functionality for their customers will take advantage of this.

pri-kise:
@louagej   Please check the following yammer post of @AndreiPanko .
https://www.yammer.com/dynamicsnavdev/threads/1762140027789312

> As source code of Shopify Connector is on GitHub we would like to reuse pattern we have for System App:
> 
> Prioritize submitting improvement over extensibility.
-  For example, if your customer needs to export multiple translations of a product to Shopify, instead of asking for making some objects extensible – submit PR that implements this feature.
- If extensibility is really necessary – for example to orchestrate multiple extensions dealing with various sales channels – please create PR directly. Remember to write the proper justification of scenario.
>
> We will not process extensibility requests for connector registered as Issues on GitHub.

I would also like it if they open up this app, but I'm not microsoft and I only try to help you to get this PR merged.

onbuyuka:
Shopify Connector – Extensibility
As you know the Shopify Connector is made non-extensible. We plan to change it in the future, and we want to try a different model.

Motivation: we want to avoid misuse of extensibility to address bugs, minor improvements, or gaps – all of that end-user expects to be part of product. Millions of events spread across the code reduce readability and scalability, increase maintenance and upgrade costs for all parties.

As source code of Shopify Connector is on GitHub we would like to reuse pattern we have for System App:

Prioritize submitting improvement over extensibility. For example, if your customer needs to export multiple translations of a product to Shopify, instead of asking for making some objects extensible – submit PR that implements this feature.
If extensibility is really necessary – for example to orchestrate multiple extensions dealing with various sales channels – please create PR directly. Remember to write the proper justification of scenario.
We will not process extensibility requests for connector registered as Issues on GitHub.

PR #23472, opened 2023-05-18
actions for processing multiple permissions

miljance:
The goal is to introduce the actions that existed on the page that is marked as obsoleted and named "Permissions (legacy)" on a page listing all permission sets. Processing multiple objects in a batch was highly appreciated feature so far.

How it looked on a paged marked for obsoletion:
image

How it is desired to look a new permission set card page:
image

### DIFF
--- a/Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
+++ b/Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
@@ -256,6 +256,275 @@ page 9859 "Tenant Permission Subform"
                         until TenantPermission.Next() = 0;
                 end;
             }
+            group("Allow Read")
+            {
+                Caption = 'Allow Read';
+                Enabled = "Object Type" = "Object Type"::"Table Data";
+                Image = Confirm;
+                action(AllowReadYes)
+                {
+                    AccessByPermission = TableData "Tenant Permission" = M;
+                    ApplicationArea = Basic, Suite;
+                    Caption = 'Yes';
+                    Image = Approve;
+                    ToolTip = 'Allow access to read data in the object.';
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelected('R', "Read Permission"::Yes);
+                    end;
+                }
+                action(AllowReadNo)
+                {
+                    AccessByPermission = TableData "Tenant Permission" = M;
+                    ApplicationArea = Basic, Suite;
+                    Caption = 'No';
+                    Image = Reject;
+                    ToolTip = 'Disallow access to read data in the object.';
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelected('R', "Read Permission"::" ");
+                    end;
+                }
+                action(AllowReadIndirect)
+                {
+                    AccessByPermission = TableData "Tenant Permission" = M;
+                    ApplicationArea = Basic, Suite;
+                    Caption = 'Indirect';
+                    Image = Indent;
+                    ToolTip = 'Allow access to read data in the object if there is read access to a related object.';
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelected('R', "Read Permission"::Indirect);
+                    end;
+                }
+            }
+            group("Allow Insertion")
+            {
+                Caption = 'Allow Insertion';
+                Enabled = "Object Type" = "Object Type"::"Table Data";
+                Image = Confirm;
+                action(AllowInsertYes)
+                {
+                    AccessByPermission = TableData "Tenant Permission" = M;
+                    ApplicationArea = Basic, Suite;
+                    Caption = 'Yes';
+                    Image = Approve;
+                    ToolTip = 'Allow access to insert data in the object.';
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelected('I', "Insert Permission"::Yes);
+                    end;
+                }
+                action(AllowInsertNo)
+                {
+                    AccessByPermission = TableData "Tenant Permission" = M;
+                    ApplicationArea = Basic, Suite;
+                    Caption = 'No';
+                    Image = Reject;
+                    ToolTip = 'Disallow access to insert data in the object.';
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelected('I', "Insert Permission"::" ");
+                    end;
+                }
+                action(AllowInsertIndirect)
+                {
+                    AccessByPermission = TableData "Tenant Permission" = M;
+                    ApplicationArea = Basic, Suite;
+                    Caption = 'Indirect';
+                    Image = Indent;
+                    ToolTip = 'Allow access to insert data in the object if there is insert access to a related object.';
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelected('I', "Insert Permission"::Indirect);
+                    end;
+                }
+            }
+            group("Allow Modification")
+            {
+                Caption = 'Allow Modification';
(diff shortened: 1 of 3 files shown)


Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
```diff
@@ -375,6 +644,46 @@ page 9859 "Tenant Permission Subform"
         CurrPage.Update(false);
     end;
 
+    local procedure UpdateSelected(RIMDX: Text[1]; PermissionOption: Option)
```
pri-kise, line 647: wouldn't it be better to move this code into a permission codeunit?
Or at least to code inside the loop?

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
```diff
@@ -375,6 +644,46 @@ page 9859 "Tenant Permission Subform"
...
+        OriginalTenantPermission: Record "Tenant Permission";
+    begin
+        CurrPage.SetSelectionFilter(TenantPermission);
+
+        if TenantPermission.FindSet() then
+            repeat
+                case RIMDX of
+                    'R':
+                        if TenantPermission."Object Type" = "Object Type"::"Table Data" then
+                            TenantPermission."Read Permission" := PermissionOption;
```
pri-kise, line 659: in our code we often use a pattern to set a Update boolean and only run the modify if anything has changed.

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
```diff
@@ -375,6 +644,46 @@ page 9859 "Tenant Permission Subform"
...
+                            if TenantPermission."Object Type" = "Object Type"::"Table Data" then begin
+                                TenantPermission."Read Permission" := PermissionOption;
+                                TenantPermission."Insert Permission" := PermissionOption;
+                                TenantPermission."Modify Permission" := PermissionOption;
+                                TenantPermission."Delete Permission" := PermissionOption;
+                            end else
+                                TenantPermission."Execute Permission" := PermissionOption;
+                        end;
+                end;
+                TenantPermission.Modify();
```
pri-kise, line 683: I personally don't understand why we should use a Modify(false) on this place.

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
miljance: Makes sense. Will do. This function is a copy of the old function that Microsoft created.

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
miljance: Makes sense. I do not see that it makes big impact here but I can certainly change it.

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
miljance: Perhaps this is a coding style topic. In general, if we already want to avoid modify when not necessary in order to save SQL time we might want to optimize further and do a simple modify without running. I found it very useful in posting routines. In this particular case there is not code on Modify trigger therefore the impact is minor anyway. My suggestion is to do as you suggest if there is a consensus that this is the best practice.

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
```diff
@@ -375,6 +644,46 @@ page 9859 "Tenant Permission Subform"
...
+            until TenantPermission.Next() = 0;
+    end;
```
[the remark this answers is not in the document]
miljance: done

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
miljance: done

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
miljance: done

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
```diff
@@ -256,6 +256,275 @@ page 9859 "Tenant Permission Subform"
...
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelectedPermissionLines('R', "Read Permission"::Indirect);
+                    end;
+                }
+            }
+            group("Allow Insertion")
+            {
+                Caption = 'Allow Insertion';
```
stkillen, line 306:
```suggestion
                Caption = 'Allow Insert';
```

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
```diff
@@ -256,6 +256,275 @@ page 9859 "Tenant Permission Subform"
...
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelectedPermissionLines('I', "Insert Permission"::Indirect);
+                    end;
+                }
+            }
+            group("Allow Modification")
+            {
+                Caption = 'Allow Modification';
```
stkillen, line 351:
```suggestion
                Caption = 'Allow Modify';
```

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
```diff
@@ -256,6 +256,275 @@ page 9859 "Tenant Permission Subform"
...
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelectedPermissionLines('M', "Modify Permission"::Indirect);
+                    end;
+                }
+            }
+            group("Allow Deletion")
+            {
+                Caption = 'Allow Deletion';
```
stkillen, line 396:
```suggestion
                Caption = 'Allow Delete';
```

Modules/System/Permission Sets/src/TenantPermissionSubForm.page.al
```diff
@@ -256,6 +256,275 @@ page 9859 "Tenant Permission Subform"
...
+
+                    trigger OnAction()
+                    begin
+                        UpdateSelectedPermissionLines('D', "Delete Permission"::Indirect);
+                    end;
+                }
+            }
+            group("Allow Execution")
+            {
+                Caption = 'Allow Execution';
```
stkillen, line 441:
```suggestion
                Caption = 'Allow Execute';
```

stkillen (changes requested):
Hi, 

This looks good to me.

Small comment; I think we should align the captions of the actions with the field captions, e.g. "Allow Read", "Allow Insert", "Allow Modify" etc.

miljance:
> Hi,
> 
> This looks good to me.
> 
> Small comment; I think we should align the captions of the actions with the field captions, e.g. "Allow Read", "Allow Insert", "Allow Modify" etc.
> 
> Furthermore, we need some _**tests**_ for these actions.

Tests :) The magic word. I'll create some. Rename - roger.

pri-kise:
@miljance the AL-GO Pipeline checks for rule "AA0005 Only use BEGIN..END to enclose compound statements."
You could fix this by the CodeAction of AL Outline Extension

miljance:
corrected

stkillen (approved):

JesperSchulz:

Build ID: 11520.

PR #23473, opened 2023-05-18
GP migration bug fixes, May 2023

jaymckinney-enavate:
This update fixes four issues found within the GP migration.

**Item on hand quantity incorrect**
- Item on hand quantity no longer includes damaged, returned, in use, or in service counts.

**Vendor/Customer posting group assignment issue**
- Fixed an issue where Vendor and Customer posting groups were not getting assigned every time.

**Master data only should not migrate beginning balances**
- Don't generate beginning balances when migrating on master data.

**Vendor Address replication issue**
- Removed the [ReplicateData = false] property from the GP Vendor Address staging table. This was causing Vendor Addresses not to be replicated in a PTE, but would also cause the same issue for the main extension if the table hardcodings are removed in the service.


nikolakukrika (approved):

JesperSchulz:

Build ID to track: 57047.

JesperSchulz:

Build ID: 10926.

PR #23585, opened 2023-05-24
Add more storage api versions

Krajzys:
Add new storage api version.


pri-kise:
@Krajzys I know you probably need only this specific version, but I might be useful to add the versions in between too.
https://learn.microsoft.com/en-us/rest/api/storageservices/previous-azure-storage-service-versions
Otherwise one had to add a older version after a newer version inside the enum.
This would be important from a technical perspective since only the name is used, but I would be easier to see if a version is missing.

Krajzys:
@pri-kise Okay, I will add the versions in between.

pri-kise (approved):

Krajzys:
I will close this PR, and create a new one from my company repo. Sorry for the mess.

PR #23629, opened 2023-05-26
Add ready SAS authorization

Krajzys:
- Add new authorization option to use pre-generated SAS Token.
This allows to use a token that was generated directly in azure portal, the benefit is that we don't have to store storage account key in BC.
- Fix unnecessary '&' being appended to query text if there weren't any parameters in the query in SAS authorization.

### DIFF
--- a/Modules/System/Azure Storage Services Authorization/src/ReadySAS/StorServAuthReadySAS.Codeunit.al
+++ b/Modules/System/Azure Storage Services Authorization/src/ReadySAS/StorServAuthReadySAS.Codeunit.al
@@ -16,7 +16,6 @@ codeunit 9065 "Stor. Serv. Auth. Ready SAS" implements "Storage Service Authoriz
         UriBuilder: Codeunit "Uri Builder";
         UriText, QueryText : Text;
     begin
-        StorageAccountName := StorageAccount;
         UriText := HttpRequestMessage.GetRequestUri();
 
         UriBuilder.Init(UriText);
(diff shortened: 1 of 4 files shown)


Modules/System/Azure Storage Services Authorization/src/StorServAuthImpl.Codeunit.al
```diff
@@ -50,6 +50,16 @@ codeunit 9063 "Stor. Serv. Auth. Impl."
         exit(StorServAuthSharedKey);
     end;
 
+    [NonDebuggable]
+    procedure ReadySAS(SASToken: Text)
```
pri-kise, line 54: You're missing the return paramter ` Interface "Storage Service Authorization"`.

Krajzys: Right, thanks, fixed.

Modules/System/Azure Storage Services Authorization/src/ReadySAS/StorServAuthReadySAS.Codeunit.al
```diff
@@ -0,0 +1,52 @@
...
+    procedure GetSharedAccessSignature(): Text
+    begin
+        exit(SharedAccessSignature);
+    end;
+
+    [NonDebuggable]
+    procedure SetSharedAccessSignature(NewSharedAccessSignature: Text)
+    begin
+        SharedAccessSignature := NewSharedAccessSignature;
+        if SharedAccessSignature.StartsWith('?') then
```
darjoo, line 46: Does this happen? I'd assume the signature given wouldn't have `?`

Krajzys: I wanted to make sure it's always in one standard format, because when you generate the SASToken from azure it does have `?` at the start, like here:
image

JesperSchulz:

Build ID to track: 57719.

PR #23651, opened 2023-05-31
Additional GP migration updates for May 2023

jaymckinney-enavate:
This submission contains the following enhancements to the GP migration.

**Migrate email addresses with Vendor and Customer addresses**
Include the email address with Vendor and Customer addresses if they are valid email addresses.

**Retain Company migration configuration settings record after migration**
The Company config settings was being deleted after the Company is migrated, but it has since been learned that it's better to keep it for troubleshooting purposes. The record will remain but will be hidden from the configuration page if the Company has been migrated.

### DIFF
--- a/Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerAddress.Table.al
+++ b/Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerAddress.Table.al
@@ -72,6 +72,9 @@ table 4048 "GP Customer Address"
     var
         ShipToAddress: Record "Ship-to Address";
         Customer: Record Customer;
+        GPSY01200: Record "GP SY01200";
+        MailManagement: Codeunit "Mail Management";
+        EmailAddress: Text[80];
         Exists: Boolean;
     begin
         if Customer.Get(CUSTNMBR) then begin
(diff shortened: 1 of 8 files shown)


Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerAddress.Table.al
```diff
@@ -97,6 +100,14 @@ table 4048 "GP Customer Address"
             if (CopyStr(ShipToAddress."Fax No.", 1, 14) = '00000000000000') then
                 ShipToAddress."Fax No." := '';
 
+            if GPSY01200.Get('CUS', CUSTNMBR, ADRSCODE) then
```
nikolakukrika, line 103: Can we move the CUS to a label or a local method, so it is not hardcoded inline

Apps/W1/HybridGP/app/src/Migration/Vendors/GPVendorMigrator.codeunit.al
```diff
@@ -387,13 +390,25 @@ codeunit 4022 "GP Vendor Migrator"
         OrderAddress."Fax No." := HelperFunctions.CleanGPPhoneOrFaxNumber(GPVendorAddress.FAXNUMBR);
         OrderAddress."Post Code" := GPVendorAddress.ZIPCODE;
         OrderAddress.County := GPVendorAddress.STATE;
+
+        if GPSY01200.Get('VEN', Vendor."No.", AddressCode) then
```
nikolakukrika, line 394: Same comment here about hardcoding inline

Apps/W1/HybridGP/app/src/pages/HybridCloudWizardExtension.PageExt.al
```diff
@@ -73,28 +73,9 @@ pageextension 4014 "Hybrid Cloud Wizard Extension" extends "Hybrid Cloud Setup W
...
-                    CleanupSettingsForCompaniesPreviouslyMigrated();
+                if Rec."Product ID" = 'DynamicsGP' then
                     Page.Run(Page::"GP Migration Configuration");
-                end;
             end;
-
         }
     }
-
-    local procedure CleanupSettingsForCompaniesPreviouslyMigrated()
```
nikolakukrika, line 85: We need to test this well. It should include enabling and disabling the cloud migration. We should also test connecting a different DB to bring in a company

nikolakukrika (dismissed):

JesperSchulz:

Build ID: 11137.

PR #23781, opened 2023-06-10
Extend Azure Blob Services API with Content Type

t-tarasiuk:
The intent of this change is to extend PutBlobBlockBlobStream and PutBlobBlockBlobText procedures with ContentType parameter.

### DIFF
--- a/Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
+++ b/Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
@@ -55,7 +55,7 @@ codeunit 9053 "ABS Blob Client"
     /// see: https://go.microsoft.com/fwlink/?linkid=2210588
     /// </summary>    
     /// <param name="ABSContainerContent">Collection of the result (temporary record).</param>
-    /// <returns>An operation reponse object</returns>
+    /// <returns>An operation response object</returns>
     procedure ListBlobs(var ABSContainerContent: Record "ABS Container Content"): Codeunit "ABS Operation Response"
     var
         ABSOptionalParameters: Codeunit "ABS Optional Parameters";
@@ -69,7 +69,7 @@ codeunit 9053 "ABS Blob Client"
     /// </summary>    
     /// <param name="ABSContainerContent">Collection of the result (temporary record).</param>
     /// <param name="ABSOptionalParameters">Optional parameters to pass.</param>
-    /// <returns>An operation reponse object</returns>
+    /// <returns>An operation response object</returns>
     procedure ListBlobs(var ABSContainerContent: Record "ABS Container Content"; ABSOptionalParameters: Codeunit "ABS Optional Parameters"): Codeunit "ABS Operation Response"
     begin
         exit(ABSClientImpl.ListBlobs(ABSContainerContent, ABSOptionalParameters));
@@ -81,7 +81,7 @@ codeunit 9053 "ABS Blob Client"
     /// </summary>    
     /// <param name="BlobList">Collection of the result (BlobList of [Text, XmlNode]).
     /// The key in the dictionary is the full blob path and the value is the complete blob xmlnode from the response</param>
-    /// <returns>An operation reponse object</returns>
+    /// <returns>An operation response object</returns>
     procedure ListBlobs(var BlobList: Dictionary of [Text, XmlNode]): Codeunit "ABS Operation Response"
     var
         ABSOptionalParameters: Codeunit "ABS Optional Parameters";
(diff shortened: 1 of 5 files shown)


t-tarasiuk:

Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
```diff
@@ -204,32 +204,32 @@ codeunit 9051 "ABS Client Impl."
...
     begin
         if UploadIntoStream('', '', '', FileName, SourceInStream) then
-            ABSOperationResponse := PutBlobBlockBlobStream(Filename, SourceInStream, ABSOptionalParameters);
+            ABSOperationResponse := PutBlobBlockBlobStream(Filename, SourceInStream, '', ABSOptionalParameters);
 
         exit(ABSOperationResponse);
     end;
 
-    procedure PutBlobBlockBlobStream(BlobName: Text; var SourceInStream: InStream; ABSOptionalParameters: Codeunit "ABS Optional Parameters"): Codeunit "ABS Operation Response"
+    procedure PutBlobBlockBlobStream(BlobName: Text; var SourceInStream: InStream; ContentType: Text; ABSOptionalParameters: Codeunit "ABS Optional Parameters"): Codeunit "ABS Operation Response"
```
pri-kise, line 212: @t-tarasiuk you must not add a parameter to an existing procedure. (This would be breaking change).
You would have to create a new one with this additional parameter.

My suggestion would be to add this ContentType to the codeunit `ABS Optional Parameters` by adding a setter and a getter method tho this codeunit.
Then you could simple use this codeunit that is already contained in all procedures.

mazhelez: In terms of breaking: in this case the codeunit is _internal_ (so is the procedure), so not really breaking.
There's also the new overload for function in the public client codeunit, so all good.

In terms of design: as far as I remember, the mandatory parameters are to be actual procedure's parameter. Everything non-mandatory is to be placed in `ABS Optional Parameters`.

pri-kise: @mazhelez you're right I'm still having some problems with the new file browser on github, since it does not open the file automatically - therefore I was looking a the wrong file. 

Well since the current approach by using one default content-type is working I would suggest to place this property in the `ABS Optional Parameters` since it isn't mandatory.

t-tarasiuk: Hey @mazhelez, @pri-kise - thank you for a quick turnaround.

The idea behind overloading the existing functions was to maintain consistency with function `AppendBlockText` in codeunit `ABS Blob Client` that have different overloads. But if it does not fit the design we can add a new function `PutBlobBlockBlob `with two overloads: for `InStream `and `Text`, but with mandatory `ContentType `parameter.

Placing ContentType into `ABS Optional Parameters` might be a bit confusing as we have a bunch of existing functions (`PutBlobPageBlob`, `PutBlobAppendBlob`, `AppendBlockText`) that have both `ContentType `and  `ABS Optional Parameters`. In this case it may be not so obvious which where to define the parameter and what priorities they have.

Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
```diff
@@ -146,10 +161,24 @@ codeunit 9053 "ABS Blob Client"
...
+    /// see: https://go.microsoft.com/fwlink/?linkid=2210387
+    /// </summary>
+    /// <param name="BlobName">The name of the blob.</param>
+    /// <param name="SourceInStream">The Content of the Blob as InStream.</param>
+    /// <param name="ContentType">Value for Content-Type HttpHeader (e.g. 'text/plain; charset=UTF-8')</param>
+    /// <param name="ABSOptionalParameters">Optional parameters to pass.</param>
+    /// <returns>An operation response object</returns>
+    procedure PutBlobBlockBlobStream(BlobName: Text; var SourceInStream: InStream; ContentType: Text; ABSOptionalParameters: Codeunit "ABS Optional Parameters"): Codeunit "ABS Operation Response"
+    begin
+        exit(ABSClientImpl.PutBlobBlockBlobStream(BlobName, SourceInStream, '', ABSOptionalParameters));
```
pri-kise, line 181: I think the ContentType parameter should be used here?

t-tarasiuk: @pri-kise - yes, sure. Sorry, missed that one. Updated.

JesperSchulz:
Kind of sounds as if we're ready to go with this one? We'd need some approvals, and off we go! 🚀

pri-kise (approved):

t-tarasiuk:
Hey @JesperSchulz - just want to check if there is anything left from my side to merge this pull request.

JesperSchulz (approved):

JesperSchulz:
Sorry, I've been a little slow here on GitHub the past weeks. We're good to go! Let's do this!

JesperSchulz:

Build ID: 11413.

JesperSchulz:

Build ID to track: 58610.

PR #23954, opened 2023-06-28
Statistical accounts

gleonsummar:
- Correction for apply totaling dimensions filters:
The cell value calculation does not apply the totaling dimensions filters, however the drilldown shows the entries with the totaling dimensions filters.
- Allow to reverse entries with HideDialog option


gleonsummar:
Redo pull request to exclude TBR commits

PR #24029, opened 2023-07-07
Update ConnectivityAppDefinitions.Codeunit.al

RSchoesterIdyn:
We (Idyn) got permission from Søren Alexandersen to add our Direct Banking app for the Belgium market.

See this pull request to add a listing for Direct Banking for the Belgium market.

### DIFF
--- a/Apps/W1/ConnectivityApps/app/src/Codeunits/ConnectivityAppDefinitions.Codeunit.al
+++ b/Apps/W1/ConnectivityApps/app/src/Codeunits/ConnectivityAppDefinitions.Codeunit.al
@@ -25,6 +25,7 @@ codeunit 20352 "Connectivity App Definitions"
         RegisterIQBanking();
         RegisterWiseBanking();
         RegisterDirectBankingNL();
+        RegisterDirectBankingBE();
         RegisterSofteraBankfeed();
     end;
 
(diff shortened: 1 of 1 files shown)


haoranpb:
Looks good to me, will start porting internally


I notice you are having a different App for NL and BE, may I ask why did you decide to make two apps for the two localization? Maybe some localized code prevent you from doing that? (it's an interesting topic for us to figure out how to help partners to build **one** app for all localizations)

JesperSchulz:

Build ID to track: 58877.

RSchoesterIdyn:
In response to [haoranpb]:
(I also wrote this anwer to Søren in a separate email).

The differentiation between Direct Banking NL, Direct Banking BE and the Direct Banking app is in fact a dilemma.

Both the Direct Banking NL and the Direct Banking BE apps are using (dependent on) the Direct Banking (W1) app. And in fact, at the place where our Direct Banking app is visible, installing the Direct Banking (W1) app is sufficient for a working solution.

However, for the Dutch market we decided at the time that it's better to automatically install the Direct Banking NL app as well, since the majority of the Dutch BC customers are using the Telebanking solution.
(So Direct Banking NL supports the Dutch Telebanking solution and in the Dutch telebanking solution there is not such a thing as Connectivity apps, where we have the benefit of visibility.)

For the Belgium market, there is also specific local functionality, but it is less obvious that the existing BC customers are using it.
And besides that, the standard BC functionality has been improved and we believe that we are at a turning point, where the standard reconciliation should be preferred over local functionality.
So, for that reason, we have decided to not automatically install the Direct Banking BE app, because a customer might not require it. And when a customer is at the place where our app can be found it implies that such a customer wants to use the standard reconciliation.

For all countries that we are allowed to add in the future, we will only install the Direct Banking (W1) app and we will not even create local specific functionality anymore.

I hope this explains our considerations.

PR #24171, opened 2023-07-20
Enhance UI of extension settings

pri-kise:
Remove Group Caption since the Caption wasn't translated and it didn't provide any value.

### DIFF
--- a/Modules/System/Extension Management/src/ExtensionSettings.Page.al
+++ b/Modules/System/Extension Management/src/ExtensionSettings.Page.al
@@ -22,7 +22,7 @@ page 2511 "Extension Settings"
         {
             group(Group)
             {
-                ShowCaption = false;
+
                 field(AppId; AppIdValue)
                 {
                     ApplicationArea = All;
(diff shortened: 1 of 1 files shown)


Modules/System/Extension Management/src/ExtensionSettings.Page.al
```diff
@@ -22,6 +22,7 @@ page 2511 "Extension Settings"
         {
             group(Group)
             {
+                ShowCaption = false;
```
darjoo, line 25: Seems like a caption was added recently. 
`Caption = 'General';`

I'll go ahead and process the rest of the PR. If you could just remove this and I'll merge the PR in.

darjoo (changes requested):

Modules/System/Extension Management/src/ExtensionSettings.Page.al
pri-kise: I wasn't sure what you expect me to do.

Remove `ShowCaption = false;` or what do you mean and we keep the "Caption=General", that isn't released yet, but part of the internal master?

darjoo (approved):

JesperSchulz:

Build ID: <None>.

Issue #24061, opened 2023-07-12
[protected function] Page "Email Editor" - "Send" Action


MattiLE:
Hi AL-Team,
can you please swap the code from the send action trigger into a protected function?
We want to use our own editor and call the send function from our page extension.

```
----OLD----
action(Send)
            {
                Promoted = true;
                PromotedOnly = true;
                PromotedCategory = Process;
                Caption = 'Send email';
                ToolTip = 'Send the email.';
                ApplicationArea = All;
                Enabled = not EmailScheduled;
                Image = SendMail;

                trigger OnAction()
                var
                    IsEmailDataValid: Boolean;
                begin
                    IsEmailDataValid := EmailEditor.ValidateEmailData(TempEmailAccount."Email Address", EmailMessageImpl);

                    if IsEmailDataValid then begin
                        IsNewOutbox := false;
                        EmailEditor.SendOutbox(Rec);
                        EmailAction := Enum::"Email Action"::Sent;

                        CurrPage.Close();
                    end;
                end;
            }
----OLD----


>>>NEW

action(Send)
            {
                Promoted = true;
                PromotedOnly = true;
                PromotedCategory = Process;
                Caption = 'Send email';
                ToolTip = 'Send the email.';
                ApplicationArea = All;
                Enabled = not EmailScheduled;
                Image = SendMail;

                trigger OnAction()
                begin
                    SendAction();
                end;
            }

protected procedure SendAction()
    var
        IsEmailDataValid: Boolean;
    begin
        IsEmailDataValid := EmailEditor.ValidateEmailData(TempEmailAccount."Email Address", EmailMessageImpl);

        if IsEmailDataValid then begin
            IsNewOutbox := false;
            EmailEditor.SendOutbox(Rec);
            EmailAction := Enum::"Email Action"::Sent;

            CurrPage.Close();
        end;
    end;
>>> NEW
```

JesperSchulz:
Thanks for reporting this. We agree, and we’ll publish a fix asap, either in an update for the current version or in the next major release. Please do not reply to this, as we do not monitor closed issues. If you have follow-up questions or requests, please create a new issue where you reference this one.

Build ID: 11478.

JesperSchulz:
Availability update: We will publish a fix for this issue in the next update for release 22.

Build ID to track: 59036.

PR #24201, opened 2023-07-24
"Email Editor" - "Send" Action protected

MattiLE:
Hi AL-Team,
can you please swap the code from the send action trigger into a protected function?
We want to use our own editor and call the send function from our page extension.

Ticket: 
https://github.com/microsoft/ALAppExtensions/issues/24061


pri-kise:
@MattiLE the Issue you mention was already closed.
Therefore this fix will be added by someone of microsoft.
Since we summer right now, it could be that this will be only availabe in BC 23 upwards.

You should check the latest versions  and the preview versions for this requested change.

darjoo:
I have made sure there's an internal PR to get it into 22.x. The naming is slightly different `ValidateAndSendEmailData`
I'll close this PR.

Issue #22142, opened 2023-02-14
Add new overload ContentType for PutBlobBlockBlobStream


Arthurvdv:
When storing, for example PDF documents, on Azure Blob Storage with the [PutBlobBlockBlobStream](https://github.com/microsoft/ALAppExtensions/tree/main/Modules/System/Azure%20Blob%20Services%20API#putblobblockblobstream-method-) function the content type is set to 'text/plain; charset=utf-8'.

![image](https://user-images.githubusercontent.com/44637996/218686178-8ba8e1e6-d7e8-4794-a31c-4023742cfe5d.png)

I would like to set/change the Content-Type to 'application/pdf'.

The [PutBlobPageBlob](https://github.com/microsoft/ALAppExtensions/tree/main/Modules/System/Azure%20Blob%20Services%20API#putblobpageblob-method-) has a overload `ContentType: Text`. Can this also be implemented for the PutBlobBlockBlobStream function?

The new functions to add should be;

```AL
procedure PutBlobBlockBlobStream(BlobName: Text; ContentType: Text; var SourceStream: InStream): Codeunit "ABS Operation Response"
```
and 

```AL
procedure PutBlobBlockBlobStream(BlobName: Text; ContentType: Text; var SourceStream: InStream; OptionalParameters: Codeunit "ABS Optional Parameters"): Codeunit "ABS Operation Response"
```

DennyHW:
We neet to store images that are referenced by an external web site.
If the ContentType is not set to "image" the picture is not loaded.

Actually the value is set by default as a generic 'application/octet-stream'
![immagine](https://github.com/microsoft/ALAppExtensions/assets/53569676/a0f86ec5-19e7-42e8-9ac1-f6f1edaf3f95)

Please give us an overload an more control, like you did with other blob type like PageBlob.
Following the universal code initiative, this is becoming a tricky situation to solve without rewriting a lot o code.

Thanks.

pri-kise:
@Arthurvdv  could you check out the recently merged PR https://github.com/microsoft/ALAppExtensions/pull/23781

Can this issue be closed now?

Arthurvdv:
@pri-kise, thank you for the PR, unfortunately it still seems to be an issue.

I've created an next minor container (sandbox/22.4.59101.0/base), where the new overload is present.

For the test I've hardcoded the value ```application/pdf```, where the Content-Type of the file in Azure Blob still shows ```text/plain; charset=utf-8```.

```AL
[...]

PR #24232, opened 2023-07-27
Added x-ms-blob-content-type to "ABS Optional Parameters"

Arthurvdv:
Make it possible to set the blob’s content type, releated to issue https://github.com/microsoft/ALAppExtensions/issues/22142.

image
https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=azure-ad#request-headers-all-blob-types

### DIFF
--- a/Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
+++ b/Modules/System/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
@@ -229,6 +229,15 @@ codeunit 9047 "ABS Optional Parameters"
     end;
 
     /// <summary>
+    /// Sets the value for 'x-ms-blob-content-type' HttpHeader for a request.
+    /// </summary>
+    /// <param name="Value">Text value specifying the HttpHeader value</param>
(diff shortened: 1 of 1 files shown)


JesperSchulz:
@pri-kise, you seem to have been involved in the discussions in #22142. Could you do a swift review of this PR?

pri-kise:
I don't think that this is quite the right approach.
If I read the documenation correctly, then my rollback in PR #24233 should fix the header `Content-Type` and then this change here isn't needed.

>If the client specifies a standard HTTP header, but not the custom property header, the value is stored in the corresponding custom property that's associated with the blob, and it's returned by a call to Get Blob Properties. For example, if the client sets the Content-Type header on the request, that value is stored in the blob's x-ms-blob-content-type property.

see: _https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=azure-ad#remarks_

JesperSchulz:
@Arthurvdv, could you verify? And if you agree, please close this PR.

JesperSchulz:
I will close this PR, assuming it isn't needed. @Arthurvdv, if you need this after all, please don't hesitate to reopen this PR and we will look at it again!

PR #24233, opened 2023-07-27
Azure Blob Storage - Fix Content Headers

pri-kise:
I fix the HttpHeaders of the Content.

This was a bug introduced in BC20 by fixing the vairable names.

CommitID: 6f7bd47e19261df4200689741d0a2c6248e89991

Additional i refactor to new Length Method of Stream.


pri-kise:
@mazhelez could you maybe take a look at this PR?

This is only fix for "Search and Replace" bug that was introduced by Replacing any "Content" string with "HttpContent"

mazhelez:
> @mazhelez could you maybe take a look at this PR?
> 
> This is only fix for "Search and Replace" bug that was introduced by Replacing any "Content" string with "HttpContent"

Looks alright.

JesperSchulz:

Build ID: 11720.

darjoo (approved):

PR #24354, opened 2023-08-03
GP migration updates for July 2023

jaymckinney-enavate:
This PR includes July 2023 updates to the GP migration.

**General Performance**
- Don't perform the staging combine step if the module is not configured to migrate. An initial step for a migration is the combining of multiple staging tables into a single table for later processing (replacement of the SQL queries). This area of the code is now optimized to only perform this operation when configured to and skip otherwise. If configured to migrate less data, this would reduce both the time a migration would take, and the amount of space used by the staging tables.

**Posting enhancements**
- Allow skip posting. Added 'skip posting' config settings and incorporated them in the posting code to give the ability to skip auto posting of Account, Customer, Bank, and/or Vendor batches.
- Posting Group account number now defaults to Posting Setup account when the Posting Group account number is empty.

**AR/AP**
- Prevent $0 transactions that cause posting to fail.
- Enhancements to Vendor Bank Account migrations. If the Vendor has more than 1 bank account, then the Code field will be appended with a number after the Vendor ID. For example: ACETRAVEL-1, ACETRAVEL-2. All Bank Account records migrated will set the Use for Electronic Payments field = true.
- Remove appended BC doc number to External Document No. field in AP and AR documents. This area of the code was appending the Document No. that is assigned to the migrated AP and AR transactions to the end of the External Doc. No. field. Feedback has been provided by Partners and Customers that this causes confusion, is unnecessary and can break integrations.
- Correct how Countries/Regions are created during migration. The migration now has the ability to perform a search on known countries, obtained by GP's list of known countries.

**Item**
- Support GP Rounding Precision For Item Unit Cost. This change updates the General Ledger Setup record "Unit-Amount Rounding Precision" value to match the greatest number of decimal places supported by a GP Item.

**Historical Snapshot improvements**
- Restart GP Snapshot if it times out (Increased the Max Attempts property).
- Snapshot performance enhancements.

**Other**
- OnResetAllCloudData() and CompanyOnAfterDelete() now cleans up necessary tables.
- Ablity to update Company migration settings and rerun the Snapshot with the changed settings.
- Added a page to add the ability to remove a Hybrid Company Status record

### DIFF
--- a/Apps/W1/HybridGP/app/src/pages/GPCompanyMigrationSettings.Page.al
+++ b/Apps/W1/HybridGP/app/src/pages/GPCompanyMigrationSettings.Page.al
@@ -1,0 +1,155 @@
+page 40056 "GP Company Migration Settings"
+{
+    ApplicationArea = All;
+    Caption = 'GP Company Migration Settings';
+    PageType = List;
+    SourceTable = "GP Company Additional Settings";
+    SourceTableView = sorting(Name) where("Name" = filter(<> ''));
+    UsageCategory = Lists;
+    Editable = true;
+    DeleteAllowed = false;
+    InsertAllowed = false;
+    ModifyAllowed = true;
+
+    layout
+    {
+        area(content)
+        {
+            repeater(General)
+            {
+                field(Name; Rec.Name)
+                {
+                    ToolTip = 'Specify the name of the Company.';
+                }
+                field("Global Dimension 1"; Rec."Global Dimension 1")
+                {
+                    ToolTip = 'Specify the segment from Dynamics GP you would like as the first global dimension in Business Central.';
+                }
+                field("Global Dimension 2"; Rec."Global Dimension 2")
+                {
+                    ToolTip = 'Specify the segment from Dynamics GP you would like as the second global dimension in Business Central.';
+                }
+                field("Migrate Bank Module"; Rec."Migrate Bank Module")
+                {
+                    ToolTip = 'Specify whether to migrate the Bank module.';
+                }
+                field("Migrate Customer Classes"; Rec."Migrate Customer Classes")
+                {
+                    ToolTip = 'Specify whether to migrate customer classes.';
+                }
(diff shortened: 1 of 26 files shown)


Apps/W1/HybridGP/app/src/pages/GPCompanyMigrationSettings.Page.al
```diff
@@ -0,0 +1,155 @@
+page 40056 "GP Company Migration Settings"
```
nikolakukrika, line 1: I believe this object should be added to the permission sets, together with the new query

Apps/W1/HybridGP/app/src/pages/GPCompanyMigrationSettings.Page.al
```diff
@@ -0,0 +1,155 @@
...
+{
+    ApplicationArea = All;
+    Caption = 'GP Company Migration Settings';
+    PageType = List;
+    SourceTable = "GP Company Additional Settings";
+    SourceTableView = sorting(Name) where("Name" = filter(<> ''));
+    UsageCategory = Lists;
```
nikolakukrika, line 8: Minor - Worksheet is usually good for this kind of tables that should always be editable.

nikolakukrika (dismissed):

Apps/W1/HybridGP/app/src/pages/GPCompanyMigrationSettings.Page.al
jaymckinney-enavate: Updated, thanks

Apps/W1/HybridGP/app/src/pages/GPCompanyMigrationSettings.Page.al
jaymckinney-enavate: Updated, thanks

nikolakukrika (approved):

JesperSchulz:

Build ID: 11735.

JesperSchulz:

Build ID to track: 59590.

PR #24426, opened 2023-08-14
[Shopify] Shpfy Order Transaction: removed Access Internal

edenisss:
Hi,

For Paypal transactions (indeed anything other than the standard Shopify Gateway) we need to get access to the Authorization field which is currently sat on “Shpfy Order Transaction", however this table is flagged as Internal.
Basically we drop this on to the External Document No. field in a Sales Order to enable it to flow through to the Bank Ledger Entries.
The change will allow to access the data directly from the table.

Since the table contains sensitive information a list of fields marked as Internal:
        field(15; "Source Name"; Code[20])
        field(16; "Credit Card Bin"; Code[10])
        field(17; "AVS Result Code"; Code[1])
        field(18; "CVV Result Code"; Code[1])
        field(19; "Credit Card Number"; Text[30])


AndreiPanko (approved):

JesperSchulz:
Will release with 2023 Wave 2.

PR #24444, opened 2023-08-16
Update Repo Version for AL-GO Settings

pri-kise:
I try to fix problems with LocalDevEnv since 22.3 is no preview version anymore.


pri-kise:
I'm gonna close this PR for the moment.
Maybe this whole settings file has to be modified.
I think we might want to use the Normal BC22.4 artifact, but I'm no expert with AL-GO yet.

This really has to be fixed, since currently the LocalDevEnv.ps1 isn't working.

JesperSchulz:
Yes, we'll look into that.

PR #24477, opened 2023-08-21
Add Procedure to Delete E-Mail Accounts

pri-kise:
This PR adds a new method to delete Accounts without UI.
Currently it isn't possible to delete E-Mail Accounts.
With this PR I want to provide the ability to delete e-Mail Accounts via code that we can clean this table On Test/Dev Environments when working with customer data to avoid accidently sending an e-Mail.

Espcecially in OnPrem where we aren't able to trigger the Cleanup Environment triggers.

### DIFF
--- a/Modules/System/Email/src/Account/EmailAccountImpl.Codeunit.al
+++ b/Modules/System/Email/src/Account/EmailAccountImpl.Codeunit.al
@@ -45,8 +45,13 @@ codeunit 8889 "Email Account Impl."
         TempEmailAccount.SetCurrentKey(Name);
     end;
 
-    [InherentPermissions(PermissionObjectType::TableData, Database::"Email Rate Limit", 'rd')]
     procedure DeleteAccounts(var EmailAccountsToDelete: Record "Email Account")
+    begin
+        DeleteAccounts(EmailAccountsToDelete, false);
+    end;
+
+    [InherentPermissions(PermissionObjectType::TableData, Database::"Email Rate Limit", 'rd')]
+    procedure DeleteAccounts(var EmailAccountsToDelete: Record "Email Account"; HideDialog: Boolean)
     var
         CurrentDefaultEmailAccount: Record "Email Account";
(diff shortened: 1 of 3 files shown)


mjmatthiesen (approved):

Modules/System/Email/src/Account/EmailAccountImpl.Codeunit.al
```diff
@@ -101,8 +108,12 @@ codeunit 8889 "Email Account Impl."
...
-        if PromptNewDefaultAccountChoice(NewDefaultEmailAccount) then
+        NewDefaultEmailAccountSelected := false;
+        if not HideDialog then begin
+            Commit();  // Commit the accounts deletion in order to prompt for new default account
+            NewDefaultEmailAccountSelected := PromptNewDefaultAccountChoice(NewDefaultEmailAccount);
+        end;
+        if NewDefaultEmailAccountSelected then
             MakeDefault(NewDefaultEmailAccount)
         else
             EmailScenario.UnassignScenario(Enum::"Email Scenario"::Default); // remove the default scenario as it is pointing to a non-existent account
```
JesperSchulz, line 119: If HideDialog is true, we would always remove the default scenario, even if the default account was not deleted, wouldn't we?

pri-kise: @JesperSchulz no, we wouldn't:

Check out the line 102:
```al
        if AllEmailAccounts.Get(CurrentDefaultAccountId, Connector) then
            exit; // The default account was not deleted or it never existed
```
The procedure is exited if the default account still exists. (This is already tested for the UI tests.

JesperSchulz: Perfect. All good then :-)

JesperSchulz (approved):

JesperSchulz:
Releases with 23.0.

JesperSchulz:

Build ID: 12044.

PR #24481, opened 2023-08-21
Azure Blob Storage improvements

ajkauffmann:
Some improvements to the Azure Blob Storage module:

New functions:
* GetBlobProperties
* BlobExists

Remark: BlobExists uses GetBlobProperties, similar to the C# code of Azure SDK.

Improvement on CopyBlob to support blobs from the same storage account.

### DIFF
--- a/Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
+++ b/Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
@@ -566,8 +566,8 @@ codeunit 132920 "ABS Blob Client Test"
         ABSOperationResponse := ABSBlobClient.GetBlobProperties(BlobName);
 
         // [THEN] The properties are as expected
-        Assert.IsTrue(ABSOperationResponse.GetHeaderValueFromResponseHeaders('x-ms-creation-time') <> '');
-        Assert.AreEqual(ABSOperationResponse.GetHeaderValueFromResponseHeaders('x-ms-blob-type'), 'BlockBlob');
+        Assert.IsTrue(ABSOperationResponse.GetHeaderValueFromResponseHeaders('x-ms-creation-time') <> '', 'Property x-ms-creation-time is missing');
+        Assert.AreEqual(ABSOperationResponse.GetHeaderValueFromResponseHeaders('x-ms-blob-type'), 'BlockBlob', 'Property x-ms-blob-type is wrong');
 
         // Clean-up
         ABSContainerClient.DeleteContainer(ContainerName);
@@ -605,7 +605,7 @@ codeunit 132920 "ABS Blob Client Test"
         BlobExists := ABSBlobClient.BlobExists(BlobName);
 
         // [THEN] The blob exists
-        Assert.IsTrue(BlobExists);
+        Assert.IsTrue(BlobExists, 'The return value of procedure BlobExists should be true.');
 
         // Clean-up
         ABSContainerClient.DeleteContainer(ContainerName);
--- a/Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
+++ b/Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
@@ -535,6 +535,82 @@ codeunit 132920 "ABS Blob Client Test"
         ABSContainerClient.DeleteContainer(ContainerName);
     end;
 
+    [Test]
+    procedure GetBlobPropertiesTest()
+    var
+        ABSOperationResponse: Codeunit "ABS Operation Response";
+        ContainerName, BlobName, BlobContent : Text;
+    begin
+        // [SCENARIO] Given a storage account and a container, PutBlobBlockBlob operation succeeds and GetBlobAsText returns the content
+        // [GIVEN] Shared Key Authorization
+        SharedKeyAuthorization := StorageServiceAuthorization.CreateSharedKey(AzuriteTestLibrary.GetAccessKey());
+
+        // [GIVEN] ABS Container 
+        ContainerName := ABSTestLibrary.GetContainerName();
+        ABSContainerClient.Initialize(AzuriteTestLibrary.GetStorageAccountName(), SharedKeyAuthorization);
+        ABSContainerClient.SetBaseUrl(AzuriteTestLibrary.GetBlobStorageBaseUrl());
+
+        ABSOperationResponse := ABSContainerClient.CreateContainer(ContainerName);
+        Assert.IsTrue(ABSOperationResponse.IsSuccessful(), 'Operation CreateContainer failed');
+
+        // [GIVEN] Block Blob
+        BlobName := ABSTestLibrary.GetBlobName();
+        BlobContent := ABSTestLibrary.GetSampleTextBlobContent();
+        ABSBlobClient.Initialize(AzuriteTestLibrary.GetStorageAccountName(), ContainerName, SharedKeyAuthorization);
+        ABSBlobClient.SetBaseUrl(AzuriteTestLibrary.GetBlobStorageBaseUrl());
+
+        ABSOperationResponse := ABSBlobClient.PutBlobBlockBlobText(BlobName, BlobContent);
+        Assert.IsTrue(ABSOperationResponse.IsSuccessful(), 'Operation PutBlobBlockBlob failed');
+
+        // [WHEN] Properties are read
+        ABSOperationResponse := ABSBlobClient.GetBlobProperties(BlobName);
+
+        // [THEN] The properties are as expected
+        Assert.IsTrue(ABSOperationResponse.GetHeaderValueFromResponseHeaders('x-ms-creation-time') <> '');
+        Assert.AreEqual(ABSOperationResponse.GetHeaderValueFromResponseHeaders('x-ms-blob-type'), 'BlockBlob');
+
+        // Clean-up
+        ABSContainerClient.DeleteContainer(ContainerName);
+    end;
+
+    [Test]
+    procedure BlobExistsTest()
+    var
+        ABSOperationResponse: Codeunit "ABS Operation Response";
+        ContainerName, BlobName, BlobContent : Text;
+        BlobExists: Boolean;
+    begin
+        // [SCENARIO] Given a storage account and a container, PutBlobBlockBlob operation succeeds and GetBlobAsText returns the content
+        // [GIVEN] Shared Key Authorization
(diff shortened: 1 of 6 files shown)


Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
```diff
@@ -590,10 +608,14 @@ codeunit 9051 "ABS Client Impl."
     var
         ABSOperationResponse: Codeunit "ABS Operation Response";
         Operation: Enum "ABS Operation";
-
     begin
         ABSOperationPayload.SetOperation(Operation::CopyBlob);
         ABSOperationPayload.SetOptionalParameters(ABSOptionalParameters);
+        if not SourceName.StartsWith('https://') then begin // SourceName is in the same storage account, use ABSOperationPayload to construct URI.
```
pri-kise, line 614: If someone uses azurite (http://127.0.0.1) for testing, is this check here still valid?
Should we maybe update the documentation of the CopyBlob to highlight that we can use a fullname (URL to the source name only the blob name?

Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
```diff
@@ -621,6 +621,54 @@ codeunit 9053 "ABS Blob Client"
         exit(ABSClientImpl.GetBlobTags(BlobName, Tags, OptionalParameters))
     end;
 
+    /// <summary>
+    /// The Get Blob Properties operation returns all user-defined metadata, standard HTTP properties, and system properties for the blob. It doesn't return the content of the blob.
+    /// see: https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=azure-ad
+    /// </summary>
+    /// <param name="BlobName">The name of the blob.</param>
+    /// <returns>An operation response object. The Blob properties are in the response headers.</returns>
+    procedure GetBlobProperties(BlobName: Text): Codeunit "ABS Operation Response"
```
pri-kise, line 630: I really like this addition.
could you maybe add the other optional request header parameter from the documentation you mention in the summary? I think then it would be more complete.

_x-ms-upn_: Optional. Version 2020-06-12 and later. Valid for accounts with hierarchical namespace enabled. If the user identity values that are returned in the x-ms-owner are true, they're transformed from Azure Active Directory object IDs to user principal names. If the values are false, they're returned as Azure Active Directory object IDs. The default value is false. Note that group and application object IDs are not translated, because they don't have unique friendly names.
https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=azure-ad#request-headers

Modules/System/Azure Blob Services API/src/ABSClientImpl.Codeunit.al
ajkauffmann: Good point. A check for http:// is added.

Modules/System/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
ajkauffmann: Added the request header x-ms-upn with value true.

pri-kise: @ajkauffmann I'm sorry, I think my comment wasn't very clear, my bad.

I thought of something like the following for the `codeunit 9047 "ABS Optional Parameters"`:

```al
    /// <summary>
    /// Sets the value for 'x-ms-upn' HttpHeader for a request.
    /// Version 2020-06-12 and later.
    /// Valid for accounts with hierarchical namespace enabled.
    /// If the user identity values that are returned in the x-ms-owner are true, they're transformed from Azure Active Directory object IDs to user principal names.
    /// If the values are false, they're returned as Azure Active Directory object IDs. The default value is false.
    /// Note that group and application object IDs are not translated, because they don't have unique friendly names.
    /// </summary>
    /// <param name="Value">Boolean value specifying the HttpHeader value</param>
    procedure ReturnUserPrincipalName("Value": Boolean)
    var
        ValueText: Text;
    begin
        // Set as text, because otherwise it might give different formatted values based on language locale
        if "Value" then
            ValueText := 'true'
        else
            ValueText := 'false';

        SetRequestHeader('x-ms-upn', ValueText);
    end;
```

pri-kise:
@ajkauffmann I'm sorry, I think my comment wasn't very clear, my bad.

I thought of something like the following for the `codeunit 9047 "ABS Optional Parameters"`:

```al
    /// <summary>
    /// Sets the value for 'x-ms-upn' HttpHeader for a request.
    /// Version 2020-06-12 and later.
    /// Valid for accounts with hierarchical namespace enabled.
    /// If the user identity values that are returned in the x-ms-owner are true, they're transformed from Azure Active Directory object IDs to user principal names.
    /// If the values are false, they're returned as Azure Active Directory object IDs. The default value is false.
    /// Note that group and application object IDs are not translated, because they don't have unique friendly names.
    /// </summary>
    /// <param name="Value">Boolean value specifying the HttpHeader value</param>
    procedure ReturnUserPrincipalName("Value": Boolean)
    var
        ValueText: Text;
    begin
        // Set as text, because otherwise it might give different formatted values based on language locale
        if "Value" then
            ValueText := 'true'
        else
            ValueText := 'false';

        SetRequestHeader('x-ms-upn', ValueText);
    end;
```

ajkauffmann:
@pri-kise, my bad, I didn't understand correctly. The x-ms-upn request header has now been added as an optional parameter.

pri-kise (approved):

JesperSchulz:
@ajkauffmann, will you provide a few tests as well?

ajkauffmann:
@JesperSchulz so far I didn't have enough time to write some extra tests. Is it okay to do so next week?

JesperSchulz:
Of course. We have until EOW next week, if you want this to be part of 2023 Wave 2!

ajkauffmann:
@JesperSchulz Tests have been added!

Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
```diff
@@ -535,6 +535,82 @@ codeunit 132920 "ABS Blob Client Test"
...
+        ABSBlobClient.SetBaseUrl(AzuriteTestLibrary.GetBlobStorageBaseUrl());
+
+        ABSOperationResponse := ABSBlobClient.PutBlobBlockBlobText(BlobName, BlobContent);
+        Assert.IsTrue(ABSOperationResponse.IsSuccessful(), 'Operation PutBlobBlockBlob failed');
+
+        // [WHEN] Blob Exists is called
+        BlobExists := ABSBlobClient.BlobExists(BlobName);
+
+        // [THEN] The blob exists
+        Assert.IsTrue(BlobExists);
```
pri-kise, line 608: @ajkauffmann did you see the errors?
I think you're test are missing the  Message/Details in the Assert Statements.
e.g. :
`Assert.IsTrue(BlobExists);` should be `Assert.IsTrue(BlobExists, 'The return value of procedure BlobExists should be true.');`

Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
```diff
@@ -535,6 +535,82 @@ codeunit 132920 "ABS Blob Client Test"
...
+
+        ABSOperationResponse := ABSBlobClient.PutBlobBlockBlobText(BlobName, BlobContent);
+        Assert.IsTrue(ABSOperationResponse.IsSuccessful(), 'Operation PutBlobBlockBlob failed');
+
+        // [WHEN] Properties are read
+        ABSOperationResponse := ABSBlobClient.GetBlobProperties(BlobName);
+
+        // [THEN] The properties are as expected
+        Assert.IsTrue(ABSOperationResponse.GetHeaderValueFromResponseHeaders('x-ms-creation-time') <> '');
+        Assert.AreEqual(ABSOperationResponse.GetHeaderValueFromResponseHeaders('x-ms-blob-type'), 'BlockBlob');
```
pri-kise, line 570:
```suggestion
        Assert.AreEqual(ABSOperationResponse.GetHeaderValueFromResponseHeaders('x-ms-blob-type'), 'BlockBlob', 'The value for property `x-ms-blob-type` is wrong.');
```

@ajkauffmann you're missing the  message parameter for method `AreEqual`

Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
ajkauffmann: Of course... I was in a rush and the compiler fooled me... It has been corrected!

Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
ajkauffmann: Of course... I was in a rush and the compiler fooled me... It has been corrected!

JesperSchulz:
Could you not have used "internalsVisibleTo"? Anyway, I better get this one processed asap! Last day if we want to make 23.0.

JesperSchulz:
FYI: I'm going with the "internalsVisibleTo" solution, as I don't want to expose functions for testability only.

ajkauffmann:
@JesperSchulz with internalsVisibleTo the response headers would only be available to the test app. But then the whole feature of reading properties would become useless because the properties are returned in the response headers.
Adding another function to convert the response headers into a dictionary would not really have any added value. Only if certain headers should remain invisible. I don't see a reason for that.
Yes please, merge that thing and make us happy! 😊

JesperSchulz:
Alright, I'll make it public... Case closed then 😎 Will release with 2023 Wave 2.

JesperSchulz (approved):

JesperSchulz:
@ajkauffmann, now I'm getting push back, that the internal function is Non Debuggable too. Will be be exposing secrets if we make that function public? I think that change will have to go through a security review - hence exposing that function won't make it for 23.0. Unless there are no security concerns?

ajkauffmann:
@JesperSchulz Yes, only response headers are exposed.
Agree with GetOperation, I noticed the same.

So no changes required anymore, we are good to go?

JesperSchulz:
We're good to go. It's going to make it. One of the final changes for 23.0 ;-)

JesperSchulz:

Build ID: 12093.

JesperSchulz:

Build ID to track: 12131.

PR #24490, opened 2023-08-23
Add Interface for "Azure Function" Implementation

Spycoclown:
Increase the testability of codeunits that call Azure Functions

### DIFF
--- a/Modules/System/Azure Function/AzureFunctions.Codeunit.al
+++ b/Modules/System/Azure Function/AzureFunctions.Codeunit.al
@@ -22,6 +22,7 @@ codeunit 7804 "Azure Functions"
     /// <param name="QueryDict">Dictionary of query parameters for the request.</param>
     /// <returns>Instance of Azure function response object.</returns>
     [NonDebuggable]
+    [Obsolete('Replaced by SendGetRequest function with AzureFunctionsInterface parameter.', '22.0')]
     procedure SendGetRequest(AzureFunctionsAuthentication: Interface "Azure Functions Authentication"; QueryDict: Dictionary of [Text, Text]): Codeunit "Azure Functions Response"
     begin
         exit(AzureFunctionsImpl.SendGetRequest(AzureFunctionsAuthentication, QueryDict));
@@ -35,6 +36,7 @@ codeunit 7804 "Azure Functions"
     /// <param name="ContentTypeHeader">Content type of the body to add to the request header.</param>
     /// <returns>Instance of Azure function response object.</returns>
     [NonDebuggable]
+    [Obsolete('Replaced by SendPostRequest function with AzureFunctionsInterface parameter.', '22.0')]
     procedure SendPostRequest(AzureFunctionsAuthentication: Interface "Azure Functions Authentication"; Body: Text; ContentTypeHeader: text): Codeunit "Azure Functions Response"
     begin
         exit(AzureFunctionsImpl.SendPostRequest(AzureFunctionsAuthentication, Body, ContentTypeHeader));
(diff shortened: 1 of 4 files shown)


pri-kise:
@Spycoclown are you planning to add tests as well? 
currently this change does no provide this much value.

Spycoclown:
No, tests for the interface are not planned at this point. The purpose is to ensure that when one uses this codeunit, one can write tests for it without actually firing an HTTP send. When using these Azure Function codeunit, one can provide an implementation for the tests that does not include a send. Alternatively, it would of course be possible to capture the send using an event handle pattern. Either one would be very good.

pri-kise:
Ah okay. I understand.
You'd like to test your process for process that uses azure functions for specific scenarios in a custom app.

I still would prefer a Setter Method for the Azure Function Codeunit: 
and add a local procedure the get the Interface (fallback to the default implementation codeunit)

Additionally the current implementation codeunit has Access = Internal (BY Design) therefore we wouldn't be able to use this Codeunit in Code to call the new methods with this codeunit.

Spycoclown:
The issue with the internal setter is that only the System App can access it, and externals can't write tests for it. Therefore, it's important not to make it internal.

Spycoclown:
@pri-kise ja, die Setter Methode finde ich auch gut, aber du hast recht, Internal ist natürlich auch ein Problem, kann man das ändern?

pri-kise:
I would suggest to add a 2nd global variable  `AzureFunctionsInterfaceSet: Boolean`;

Add a local  method somehow called 
```al
GetAzureFunctionImplementation(): Interface "IAzureFunctions"
begin
// if not Initiazlized -> Return Default
// else -> return global Interface
end;
```
similiar to the following methods inside the factory codeunit of @vjekob
https://github.com/vjekob/bctechdays2023/blob/ed7ee27f5fe854ee74bbbf3e121551eb8cbdd66b/app/src/Midjourney.Interfaces/Factory/MidjourneyFactory.Codeunit.al#L33

pri-kise (approved):

Modules/System/Azure Function/AzureFunctions.Codeunit.al
```diff
@@ -37,7 +48,7 @@ codeunit 7804 "Azure Functions"
     [NonDebuggable]
     procedure SendPostRequest(AzureFunctionsAuthentication: Interface "Azure Functions Authentication"; Body: Text; ContentTypeHeader: text): Codeunit "Azure Functions Response"
     begin
-        exit(AzureFunctionsImpl.SendPostRequest(AzureFunctionsAuthentication, Body, ContentTypeHeader));
+        exit(GetAzureFunctionInterface.SendPostRequest(AzureFunctionsAuthentication, Body, ContentTypeHeader));
```
pri-kise, line 51: You must add parantheses to any procedure
>   Error: AA0008 You must specify open and close parenthesis after 'GetAzureFunctionInterface'.

Spycoclown: yeah, I saw it too, but I didn't have time. I fixed it now ;) Thx

Spycoclown:
@JesperSchulz is something missing or do i just have to wait?

Spycoclown:
great, thx for the information.

pri-kise:
@Spycoclown you could check out the new RestClient Module.

Maybe it's an option to refactor this module to make use of this RestClient Module.
Then you would only have to provide the a way to set the HttpClientHandler on the public facade codeunit.
For testing you then could simple replace the HttpClientHandler Interface with one for testing.

Spycoclown:
@pri-kise yes, I can have a look a it. Should I use the new SecuredText?

PR #24571, opened 2023-08-25
GP migration changes for 08/2023

jaymckinney-enavate:
#### Remove pre-migration data cleanup code
- A Company should only be migrated once. If it needs to be done again, it should first be deleted. Because of this, the pre-migration data cleanup code was unnecessary.

#### Ability to skip blank segments
- Blank segments were causing migrations to fail.
 
#### Use the "Oldest GL Year to Migrate" setting to also limit the Accounting Periods created in BC.
- This prevents unnecessary Accounting Periods from being created.

#### Additional changes to PO line logic.
- Much more accurate counts and unit costs.
- No longer includes the currency with the PO header. The amounts migrated are already in the functional currency.

#### Enhancements to migrating Vendor Bank Accounts
- Always rename the bank code to Vendor No. and append a number if there are multiple accounts.
- If only one account, don't append number and assign it as the preferred Vendor bank account.
- If the Vendor No. is 20 characters in length (max size), the resulting bank code will use the first 18 characters of the Vendor No. 
- Update the US extension to update the 'Use for electronic payments' for all preferred Vendor bank accounts.

#### Enhancement to migrated email addresses

Currently the migration tool migrates email addresses contained in the INET1 field from the GP SY01200.

The migration tool will be expanded to also migrate the following fields in addition to the INET1 field.   These will be placed in the E-Mail field on the Customer (18) and Vendor (23) tables respectively.  

- EmailToAddress
- EmailCcAddress
- EmailBCCAddress

The Customer and Vendor record E-mail field allows for multiple email addresses separated by a semicolon.
The Customer and Vendor address record's E-mail field only allows a single email address.

Email address selection order: 

1. To
2. Cc
3. Bcc
4. INET1

If the email address is too large for the field, it is skipped in favor of a smaller one.
If the email address is syntactically incorrect, it is also skipped in favor of a correct one.


Apps/W1/HybridGP/app/src/Migration/Customers/GPCustomerTransactions.page.al
```diff
@@ -1,8 +1,13 @@
 page 4094 "GP Customer Transactions"
 {
+    Caption = 'GP Customer Transactions';
+    AdditionalSearchTerms = 'staging';
     PageType = List;
     SourceTable = "GP Customer Transactions";
-    Caption = 'Customer Transactions';
+    DeleteAllowed = false;
```
nikolakukrika, line 7: Why are we blocking insert and delete? 
Maybe we should raise a warning?

Apps/W1/HybridGP/app/src/Migration/History/GPPOPReceiptApply.table.al
```diff
@@ -276,6 +276,7 @@ table 4060 "GPPOPReceiptApply"
     {
     }
 
+    [Obsolete('This procedure is no longer used, use GetSumQtyShippedByUnitCost instead.', '23.0')]
```
nikolakukrika, line 279: We need #if not CLEAN23, #endif tags

Apps/W1/HybridGP/app/src/Migration/History/GPPOPReceiptApply.table.al
```diff
@@ -294,6 +295,27 @@ table 4060 "GPPOPReceiptApply"
         exit(TotalShipped);
     end;
 
+    procedure GetSumQtyShippedByUnitCost(PO_Number: Text[18]; PO_LineNo: Integer; Location: Text[12]; UnitCost: Decimal): Decimal
+    var
+        TotalShipped: Decimal;
+    begin
+        TotalShipped := 0;
+        Rec.Reset();
```
nikolakukrika, line 303: It is dangerous to use Rec, because you will delete all filters. It is better to use a different variable and to copy filters or use a rec to filter on the values.

Apps/W1/HybridGP/app/src/Migration/History/GPPOPReceiptApply.table.al
```diff
@@ -311,4 +333,24 @@ table 4060 "GPPOPReceiptApply"
 
         exit(TotalInvoiced);
     end;
+
+    procedure GetSumQtyInvoicedByUnitCost(PO_Number: Text[18]; PO_LineNo: Integer; Location: Text[12]; UnitCost: Decimal): Decimal
+    var
+        TotalInvoiced: Decimal;
+    begin
+        TotalInvoiced := 0;
+        Rec.Reset();
```
nikolakukrika, line 342: Same comment here regarding the reset of the filters.

Apps/W1/HybridGP/app/src/Migration/Support/GPConfiguration.table.al
```diff
@@ -33,6 +33,9 @@ table 4024 "GP Configuration"
         {
             DataClassification = SystemMetadata;
             InitValue = false;
+            ObsoleteState = Pending;
```
nikolakukrika, line 36: Same comment regarding if not CLEAN tags

Apps/W1/HybridGP/app/src/Migration/Support/HelperFunctions.codeunit.al
```diff
@@ -303,54 +303,14 @@ Codeunit 4037 "Helper Functions"
         CustomerDataMigrationFacade.CreateCountryIfNeeded(CountryCode, CountryName, AddressFormatToSet::"City+County+Post Code", ContactAddressFormatToSet::"After Company Name");
     end;
 
+    [Obsolete('Data cleanup is no longer performed before migration.', '23.0')]
```
nikolakukrika, line 306: Same comment regarding IF Not cleantags

Apps/W1/HybridGP/app/src/codeunits/HybridGPManagement.codeunit.al
```diff
@@ -16,14 +16,25 @@ codeunit 4016 "Hybrid GP Management"
...
 
         UpdateStatusOnHybridReplicationCompleted(RunId, NotificationText);
         HandleInitializationofGPSynchronization(RunId, SubscriptionId, NotificationText);
+
+        GPConfiguration.GetSingleInstance();
+        if not GPConfiguration."Use Two Step Process" then begin
+            HybridReplicationSummary.SetCurrentKey("Start Time");
+            if HybridReplicationSummary.FindLast() then
+                if HybridReplicationSummary.Status = HybridReplicationSummary.Status::UpgradePending then
+                    Session.StartSession(SessionId, Codeunit::"Start Data Upgrade", CompanyName, HybridReplicationSummary, GetDefaultJobTimeout());
```
nikolakukrika, line 36: We do not trigger the upgrade like this. There is already a procedure that starts the data upgrade. The new method is not needed.

Apps/W1/HybridGP/app/src/pages/GPMigrationGeneralSettings.Page.al
```diff
@@ -0,0 +1,57 @@
+page 4102 "GP Migration General Settings"
```
nikolakukrika, line 1: If you agree, I think we should leave the 2 step out of this PR and keep the code running as it is. Lets align the code once the troubleshooting changes are in. I believe we will need to churn more code, so it would be good to align the changes.

nikolakukrika (approved):

Apps/W1/HybridGP/app/src/Migration/Support/GPConfiguration.table.al
jaymckinney-enavate: Could this cause future upgrades to fail since it would be actually removing the field?

Apps/W1/HybridGP/app/src/pages/GPMigrationGeneralSettings.Page.al
jaymckinney-enavate: Will do

nikolakukrika (approved):
Left comments regarding obsoletions

Apps/W1/HybridGP/app/src/Migration/Support/GPConfiguration.table.al
jaymckinney-enavate: Thanks

nikolakukrika (approved):

JesperSchulz:

Build ID: 12510.

JesperSchulz:

Build ID to track: 12567.

Issue #24604, opened 2023-08-28
Codeunit Data Compression utility methods


hemisphera:
Duplicate of issue #419 from the `BusinessCentralApps` repo for implementation, has been approved over there by @JesperSchulz.

---- original issue ----

**🕮 Describe the issue**

The codeunit `Data Compression` is missing two handy utility methods:

1. `IsZip` that indicates whether a given stream actually is a ZIP stream. This comes in handy when a user has the ability to either provide a ZIP or a non-ZIP stream, and the code needs to react accordingly. This codeunit already has a method that does so for GZip streams, but not for ZIP.
2. An overload for `ExtractEntry` that creates TempBlob Codeunit instead of an outstream. Without this, I often find myself writing exactly this, because I need to pass along the decompressed data and receiving only an OutStream, I need to first manually write it it to a TempBlob to allow it being read further down my code.

pri-kise:
I'm really looking forward to your PR to address this issue. :)

PR #24625, opened 2023-08-29
#24604: Add new "IsZip" method and "ExtractEntry" overload

hemisphera:
Add new "IsZip" method and new overload for "ExtractEntry". 

@pri-kise haven't got the chance to really test this in the context of the System Application (copied somewhat from another app of mine where I implemented this). Not exactly sure, but I'm having a lot of issues with getting `System.Drawing` (among others) to load properly in this repo, following the getting started guide. Any idea?

### DIFF
--- a/Modules/System/Data Compression/src/DataCompression.Codeunit.al
+++ b/Modules/System/Data Compression/src/DataCompression.Codeunit.al
@@ -113,6 +113,22 @@ codeunit 425 "Data Compression"
     end;
 
     /// <summary>
+    /// Extracts an entry from the ZipArchive to a TempBlob instance.
+    /// </summary>
+    /// <param name="EntryName">The name of the ZipArchive entry to be extracted.</param>
+    /// <param name="TempBlob">The TempBlob the uncompressed binary data will be written to.</param>
+    procedure ExtractEntry(EntryName: Text; var TempBlob: Codeunit "Temp Blob"): Integer
+    var
+        os: OutStream;
+        EntryLength: Integer;
+    begin
+        Clear(TempBlob);
+        TempBlob.CreateOutStream(os);
+        DataCompressionImpl.ExtractEntry(EntryName, os, EntryLength);
+        exit(EntryLength);
+    end;
+
+    /// <summary>
     /// Adds an entry to the ZipArchive.
     /// </summary>
     /// <param name="InStreamToAdd">The InStream that contains the binary content that should be added as an entry in the ZipArchive.</param>
@@ -140,6 +156,16 @@ codeunit 425 "Data Compression"
     procedure IsGZip(InStream: InStream): Boolean
     begin
         EXIT(DataCompressionImpl.IsGZip(InStream));
+    end;
+
+    /// <summary>
+    /// Determines whether the given InStream is compressed with Zip.
+    /// </summary>
+    /// <param name="InStream">An InStream that contains binary content.</param>
+    /// <returns>Returns true if and only if the given InStream is compressed with Zip</returns>
+    procedure IsZip(InStream: InStream): Boolean
+    begin
+        exit(DataCompressionImpl.IsZip(InStream));
     end;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


Modules/System/Data Compression/src/DataCompressionImpl.Codeunit.al
```diff
@@ -108,6 +108,18 @@ codeunit 421 "Data Compression Impl."
         end;
     end;
 
+    [TryFunction]
+    procedure IsZip(InputInStream: InStream)
```
pri-kise, line 112: Shouldn't it be possible to use a similiar pattern as for the GZIP Compressed Stream.
See Header: https://en.wikipedia.org/wiki/ZIP_(file_format)#File_headers

hemisphera: Yeah. I thought about that too. Delegating the check to the actual routines that are already in place seemed better to me rather than reinvent the wheel.
If the header check is preferred, I can do that.

pri-kise: We can wait for Feedback from microsoft on this topic.

pri-kise:
And that someone of microsoft who is responsible for the system application accepts this PR there should be some tests added (copied) for checking your new methods.

JesperSchulz:
A test, we would need. Other than that, this looks fine, I'd say!

Modules/System/Data Compression/src/DataCompression.Codeunit.al
```diff
@@ -112,6 +112,22 @@ codeunit 425 "Data Compression"
...
+    /// Extracts an entry from the ZipArchive to a TempBlob instance.
+    /// </summary>
+    /// <param name="EntryName">The name of the ZipArchive entry to be extracted.</param>
+    /// <param name="TempBlob">The TempBlob the uncompressed binary data will be written to.</param>
+    procedure ExtractEntry(EntryName: Text; var TempBlob: Codeunit "Temp Blob"): Integer
+    var
+        os: OutStream;
+        EntryLength: Integer;
+    begin
+        Clear(TempBlob);
```
JesperSchulz, line 125: No logic in the facade, please.

Modules/System/Data Compression/src/DataCompressionImpl.Codeunit.al
```diff
@@ -108,6 +108,27 @@ codeunit 421 "Data Compression Impl."
         end;
     end;
 
+    procedure IsZip(InputInStream: InStream): Boolean
+    var
+        impl2: Codeunit "Data Compression Impl.";
```
JesperSchulz, line 113: CodeCop will reject this variable name. Must be DataCompressionImpl2, I fear.

hemisphera: On a second thought, I agree with @pri-kise on this. I personally try to avoid a `TryFunction` for business logic, especially if "low level" like this. I have rewritten the check and now it's testing the ZIP header - just as `IsGZip` does.

Modules/System/Data Compression/src/DataCompression.Codeunit.al
hemisphera: Fixed.

Modules/System/Data Compression/src/DataCompressionImpl.Codeunit.al
hemisphera: Done.

pri-kise:
@hemisphera to get your test-App to work with, you only have to add Any to the dependency in the app.json of your TestApp.
I had the same issue today: 
https://github.com/microsoft/ALAppExtensions/pull/24667/files#diff-ba6aadc04b9281606c6d381e675e81649a214df954df883f3f00d303bb9cad15

pri-kise (approved):

JesperSchulz:
Releases with 23.0 🥳

JesperSchulz:

Build ID: 12071.

PR #24637, opened 2023-08-29
GP - Hotfix to correct previous merge issue

jaymckinney-enavate:
A previous merge placed code in the wrong procedure, which prevented accounts from being created when the migration was configured to only migrate GL master data.


nikolakukrika (approved):

JesperSchulz:

Build ID: 12048.

Issue #12381, opened 2021-05-17
[Extensibility Enhancement] Add Check for Max Field Length - Translation Module


pri-kise:
I'm using the translation module and I would like to know what do you think about a check for the Translation table that returns an error if the translation text exceeds the Max. Length of the table field?
https://github.com/microsoft/ALAppExtensions/blob/5c490ab80d8f5c4ac26641604eba79dbe1b1eb47/Modules/System/Translation/src/Translation.Table.al#L33

I could create a PR, which throws an error if you agree that this is the expected behaviour.

mazhelez:
@pri-kise interesting suggestion. Can you elaborate on that?

What are the scenarios you want to cover? What could be a useful reaction to such an error?

pri-kise:
In my opinion the common use case for the translation module is to enable the user to store translations for specific fields that are used as an alternative for Language Specific Documents / Records.
If i look into the baseapp that current has its own translation tables (Item Translation, Payment Method Translation), they are commonly used to transfer the translations as replacement of the default value (Default Language).
Therefore I would like to throw the error early in the process - while storing the translations - instead of resulting in a shortened translated text during runtime.

JesperSchulz:
I see your point! However, I'm not sure if an error maybe isn't too harsh? What if you'd like to store the proper translation, even if it doesn't fit into the field? Would that never be a scenario? Also, what about the system, where users already have stored translations which exceed the field length? How would we handle those once your rule has been implemented? Would we just truncate them?

Maybe this validation could be an optional thing you configure? Maybe it could just be a warning, which informs you that you've added a string which exceeds the field length?

I do see a valid use case here and we would love to see your PR, but you should rethink if an hard stop error is the right thing to implement :-)

Closing this issue and waiting eagerly to see your solution as a PR. Have a lovely weekend!

PR #24667, opened 2023-08-30
Check for Field Length on setting via UI

### LINKED ISSUE #1239: [Event Change] Table 81 - OnBeforeLookUpAppliesToDocCust
The publisher should contain a "Handled"-Flag and the record should be passed as VAR

image

image

Same for publisher
- OnBeforeLookUpAppliesToDocEmpl
- OnBeforeLookUpAppliesToDocVend

pri-kise:
Fixes  [#12381](https://github.com/microsoft/ALAppExtensions/issues/12381)
Currently the length check is optional an only performed via UI

### DIFF
--- a/Modules/System Test Libraries/Translation/src/TranslationTestTable.Table.al
+++ b/Modules/System Test Libraries/Translation/src/TranslationTestTable.Table.al
@@ -21,6 +21,10 @@ table 137121 "Translation Test Table"
         {
             Editable = false;
         }
+        field(4; TextFieldWithLimitedLength; Text[100])
+        {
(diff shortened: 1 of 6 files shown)


pri-kise:
Maybe @JesperSchulz you can take a look at this PR, since we had the communication on this old issue.

JesperSchulz:
I'll take a look!

JesperSchulz (approved):

JesperSchulz:
Releases with 23.0 🥳

JesperSchulz:

Build ID: 12071.

PR #24803, opened 2023-09-06
AL Rest Client

ajkauffmann:
Provides functionality to easily call REST web services

The module contains methods to support
- Calling web services with just one line of code
- Creating request content from Text, JSON, XML or binary data
- Read the response as Text, JSON, XML or binary data
- Authenticate using basic authentication

Some examples:
```
// Getting text
ResponseText := ALRestClient.Get(Url).Content().AsText();

// Getting binary data
TempBlob := ALRestClient.Get(Url).Content().AsBlob();

// Posting binary data and returning binary data
ALHttpContent.Create(TempBlob)
TempBlob := ALRestClient.Post(Url, ALHttpContent).Content().AsBlob();

// Getting a JsonObject
JsonObject := ALRestClient.GetAsJson(Url).AsObject();

// Posting a JsonObject and returning the result as JsonObject
JsonObject := ALRestClient.PostAsJson(Url, JsonObject).AsObject();

// Using Basic Authentication
HttpAuthenticationBasic.Initialize('user01', 'Password123');
ALRestClient.Initialize(HttpAuthenticationBasic);
ALHttpResponseMessage := ALRestClient.Get('https://httpbin.org/basic-auth/user01/Password123');
```

Remarks:
- Support for OAuth can be added later. A module is already available, but it requires some internal discussion.
- Telemetry signal RT0019 (Outgoing Web Service Request) will be lost for the ISV because the platform only sends it to the app that uses HttpClient directly. This needs to be addressed somehow.
- Documentation (like a readme.md file) can be generated from the XML comments, but maybe it's better to write a more extensive document how to use this new module.


Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
```diff
@@ -0,0 +1,86 @@
...
+    begin
+        AssertInitialized();
+        HttpRequestMessage.SetRequestUri(Uri);
+    end;
+
+    /// <summary>Adds a request header to the HttpRequestMessage object.</summary>
+    /// <param name="HeaderName">The name of the header to add.</param>
+    /// <param name="HeaderValue">The value of the header to add.</param>
+    /// <remarks>If the header already exists, it will be overwritten.</remarks>
+    procedure AddRequestHeader(HeaderName: Text; HeaderValue: Text)
```
PeterConijn, line 46: Since you're using this to add authorization headers in the AL Rest Client Impl. codeunit, should this procedure not be tagged as NonDebuggable? Alternatively, there can be a separate procedure, e.g. AddAuthorizationHeader(HeaderName: Text; HeaderValue: Text) that is tagged as NonDebuggable.

Modules/System/AL Rest Client/src/Authentication/HttpAuthenticationBasic.Codeunit.al
```diff
@@ -0,0 +1,47 @@
+/// <summary>Implementation of the "Http Authentication" interface for a request that requires basic authentication</summary>
+codeunit 50304 "Http Authentication Basic" implements "Http Authentication"
+{
+    var
+        GlobalUsername, GlobalPassword : Text;
```
PeterConijn, line 5: I think these globals should also be tagged as NonDebuggable

PeterConijn (commented):
This looks really awesome, AJ. I did find some security considerations, but I can't wait until this is in BC

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
```diff
@@ -0,0 +1,264 @@
...
+        HttpContent.ReadAs(InStr);
+    end;
+
+    /// <summary>Gets the content of the HTTP response message as an XmlDocument.</summary>
+    /// <returns>The content of the HTTP response message as an XmlDocument.</returns>
+    /// <remarks>Fails in case the content is not a valid XML document.</remarks>
+    procedure AsXmlDocument() ReturnValue: XmlDocument
+    var
+        XmlReadOptions: XmlReadOptions;
+        xml: Text;
```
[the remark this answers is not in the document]
ajkauffmann: Good catch, removed it.

Modules/System/AL Rest Client/src/ALHttpResponseMessage.Codeunit.al
```diff
@@ -0,0 +1,99 @@
...
+    end;
+
+    /// <summary>Sets the HTTP response message.</summary>
+    /// <param name="ResponseMessage">The HTTP response message.</param>
+    procedure SetResponseMessage(var ResponseMessage: HttpResponseMessage)
+    begin
+        AssertInitialized();
+        HttpResponseMessage := ResponseMessage;
+        ALHttpContent := ALHttpContent.Create(ResponseMessage.Content);
+        HasResponse := true;
```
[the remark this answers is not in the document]
ajkauffmann: It's used as an internal flag to avoid reading values from an uninitialized HttpResponseMessage object.
There is a response if the status code has a non-zero value.
In theory, SetResponseMessage could be an internal procedure because it is normally only used after making the call. But for testing purposes if could be useful to have a public procedure.
And I don't think SetResponseMessage should return HasResponse. That does not have added value I'd say.

Modules/System/AL Rest Client/src/ALHttpResponseMessage.Codeunit.al
```diff
@@ -0,0 +1,99 @@
...
+    begin
+        Clear(HttpResponseMessage);
+        HasResponse := false;
+        Initialized := true;
+    end;
+
+    /// <summary>Indicates whether the HTTP response message has a success status code.</summary>
+    /// <returns>True if the HTTP response message has a success status code; otherwise, false.</returns>
+    /// <remarks>Any value in the HTTP status code range 2xx is considered to be successful.</remarks>
+    procedure IsSuccessStatusCode() Result: Boolean
```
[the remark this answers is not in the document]
ajkauffmann: I'm a big fan of using named return values. The reason for that is that I've read somewhere that giving return values a name turns them into assigned variables which theoretically is faster than using unnamed variables. I just can't find the source anymore.
The performance gain is probably very small for a single function but could have an impact when used as much as possible.

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
```diff
@@ -0,0 +1,153 @@
...
+
+    local procedure AssertInitialized()
+    begin
+        if not Initialized then
+            Error(NotInitializedErr);
+    end;
+
+    procedure SendRequest(var ALHttpRequestMessage: Codeunit "AL Http Request Message") ALHttpResponseMessage: Codeunit "AL Http Response Message"
+    begin
+        if not TrySendRequest(ALHttpRequestMessage, ALHttpResponseMessage) then;
```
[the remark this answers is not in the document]
ajkauffmann: This is up for discussion. I chose to catch the errors here and handle them later. The procedure Send will throw the error when there is a connection problem. But I believe that http status codes should not be handled in any way, that's up to the caller.
The reason for encapsulating it in a TryFunction is to throw the correct error message here, so calling functions can grab GetLastErrorText. At the same time, the error message is also stored in the "AL Http Response Message" object, so that is a duplicate.

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
ajkauffmann: Good point. I'll go with NonDebuggable on the AddRequestHeader because it can be used for other confidential information as well.

Modules/System/AL Rest Client/src/Authentication/HttpAuthenticationBasic.Codeunit.al
ajkauffmann: I totally agree.

Modules/System/AL Rest Client/src/ALHttpResponseMessage.Codeunit.al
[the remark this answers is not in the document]
PeterConijn: If you can find that source, I would love to see it, since I was always taught the opposite and if named return values do give a performance gain, I would like to institute that in our organization.

PeterConijn (approved):

Modules/System Tests/AL Rest Client/app.json
```diff
@@ -0,0 +1,56 @@
...
+    {
+      "id":  "63ca2fa4-4f03-4f2b-a480-172fef340d3f",
+      "name":  "System Application",
+      "publisher":  "Microsoft",
+      "version":  "22.0.0.0"
+    }
+  ],
+  "screenshots": [],
+  "platform": "22.0.0.0",
+  "idRanges": [
```
JesperSchulz, line 41: Let's use range 2350.. for this module. I'll reserve.

JesperSchulz:
Looks like a logo is missing. I'll add it. Would love to make the compilation succeed to start with :-)

ajkauffmann:
Replaced the usage of enum "Http Request Type" from the OAuth module with a new enum "Http Method". I figured it was more appropriate for the module to have its own enum instead of taking a dependency on the OAuth module.
Also added a dependency on Base64 Convert to solve a compiler error.

Modules/System/AL Rest Client/app.json
```diff
@@ -0,0 +1,40 @@
+{
+  "id": "812b339d-a9db-4a6e-84e4-fe35cbef0c44",
+  "name": "AL Rest Client",
```
IhorHandziuk, line 3: No need to add `AL`, just `Rest Client` is good.

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
```diff
@@ -0,0 +1,261 @@
+/// <summary>Holder object for the Http Content data.</summary>
+codeunit 2354 "AL Http Content"
```
IhorHandziuk, line 2: No need to prefix any of the objects with `AL`. The consumers could easily differentiate between the built-in HttpClient and this one by type.

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
```diff
@@ -0,0 +1,261 @@
...
+    /// <param name="Content">The content to send to the server. Valid types are Codeunit (Temp Blob), InStream, JsonObject, JsonArray, JsonToken, XmlDocument, Text</param>
+    /// <param name="ContentType">The content type of the content to send to the server.</param>
+    /// <returns>The AL Http Content object.</returns>
+    /// <remarks>
+    /// Text types will automatically get 'text/plain' if not specified in the ContentType parameter.
+    /// Json types will automatically get 'application/json'. 
+    /// Xml types will automatically get 'text/xml'. 
+    /// Codeunit "Temp Blob" and InStream types will automatically get 'application/octet-stream' if not specified in the ContentType parameter.
+    /// </remarks>
+    procedure Create(Content: Variant; ContentType: Text) ALHttpContent: Codeunit "AL Http Content"
```
IhorHandziuk, line 39: I know it's more tedious, but I think it would be better to have an overload for every supported type instead of accepting a `Variant`. This way developers would see an error at compile time instead of run time if they attempt to pass something unsupported.

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
```diff
@@ -0,0 +1,87 @@
+/// <summary>Holder object for the HTTP request data.</summary>
+codeunit 2352 "AL Http Request Message"
+{
+    var
+        HttpRequestMessage: HttpRequestMessage;
+        Initialized: Boolean;
+        NotInitializedErr: Label 'The Http Request Message has been initialized';
```
IhorHandziuk, line 7: ...has _not_ been...

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
```diff
@@ -0,0 +1,87 @@
...
+
+    /// <summary>Gets the HttpRequestMessage that is represented by the AL HttpRequestMessage object.</summary>
+    /// <returns>The HttpRequestMessage that is represented by the AL HttpRequestMessage object.</returns>
+    procedure GetRequestMessage() ReturnValue: HttpRequestMessage
+    begin
+        AssertInitialized();
+        ReturnValue := HttpRequestMessage;
+    end;
+
+    local procedure AssertInitialized()
```
IhorHandziuk, line 82: I don't understand the intention with using this procedure. Essentially, it makes sure that the consumer will always have to call `Initialize` before using any of the functions, but the actual code of `Initialize` just clears the global variable.

From the procedures on this codeunit it should be quite clear to any developer that it holds state, and that if they want to work with a new request message, they should just call `Clear(ALHttpRequestMessage)`, which would clear the global variables.

IhorHandziuk: The same applies to other codeunits that use the same pattern.

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
```diff
@@ -0,0 +1,261 @@
+/// <summary>Holder object for the Http Content data.</summary>
+codeunit 2354 "AL Http Content"
+{
```
IhorHandziuk, line 3: We have a convention of using the Facade pattern for system application modules. Meaning that public codeunits should not have any logic code present. In some cases it may just seem like extra work for very little readability gain, but it does make it easier to navigate the application when consistently applied.

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
```diff
@@ -0,0 +1,153 @@
...
+        HttpResponseMessage: HttpResponseMessage;
+    begin
+        AssertInitialized();
+
+        ALHttpResponseMessage.Initialize();
+        ConnectionError := false;
+
+        if HttpAuthentication.IsAuthenticationRequired() then
+            if not Authorize(ALHttpRequestMessage) then begin
+                ErrorMessage := GetLastErrorText();
```
IhorHandziuk, line 114: It would be nice to send some telemetry when we encounter errors. We should of course make sure that the error text could never have customer content (e. g. information from the request).
This applies to the errors below in this procedure as well.

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
```diff
@@ -0,0 +1,153 @@
+codeunit 2351 "AL Rest Client Impl."
+{
+    Access = Internal;
+    InherentEntitlements = X;
```
IhorHandziuk, line 4: Please add `InherentEntitlements` and `InherentPermissions` properties to all the other objects as well. There is no permission requirement to use the built-in HttpClient, this functionality should not be different.

Modules/System/AL Rest Client/src/ALRestClient.Codeunit.al
```diff
@@ -0,0 +1,339 @@
+/// <summary>Provides functionality to easily work with the HttpClient object.</summary>
```
IhorHandziuk, line 1: [Not directly related to this PR]
Normally when we add new functionality like this, we also try to have some uptake in the application.
Not only it would bring the benefit of this module to the features that uptake it, but it would also set the example for future implementations as to how to use this functionality.

The emailing apps in this repo could be a very good place to do this if you choose to do so 🙂.

IhorHandziuk (commented):
Overall a very good module. We could benefit a lot from using this functionality.

Modules/System/AL Rest Client/src/ALRestClient.Codeunit.al
NKarolak: @IhorHandziuk
Probably it's only me, but I don't get your message here. Who do you address? AJ, to change his code (then: what is the actual instruction?) or Microsoft, to plan to use the new module in near future? Or am I totally mistaken?

duichwer: @IhorHandziuk 
Some other PRs are waiting since they should use this module.
E g. The PR for a generic Microsoft Graph API wrapper.
Therefore we should get this working and nobody stops Microsoft or anyone else from refactoring any other Module afterwards.

Modules/System/AL Rest Client/src/Authentication/HttpAuthenticationBasic.Codeunit.al
```diff
@@ -0,0 +1,48 @@
...
+{
+    var
+        [NonDebuggable]
+        GlobalUsername, GlobalPassword : Text;
+
+    [NonDebuggable]
+    /// <summary>Initializes the authentication object with the given username and password</summary>
+    /// <param name="Username">The username to use for authentication</param>
+    /// <param name="Password">The password to use for authentication</param>
+    procedure Initialize(Username: Text; Password: Text)
```
christianbraeunlich, line 12: Just a small note: if we don't merge this before v23, we should make use of the new SecretText data type. Or change it afterwards

JesperSchulz: That is an excellent point. We could release without it initially and uptake later though.

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
kb-dynaway: and a dot in the end.

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
```diff
@@ -0,0 +1,261 @@
...
+    procedure SetContentTypeHeader(ContentType: Text)
+    begin
+        if ContentType = '' then
+            Error(ContentTypeEmptyErr);
+        SetHeader('Content-Type', ContentType);
+    end;
+
+    /// <summary>Sets the Content-Encoding header of the HttpContent object.</summary>
+    /// <param name="ContentEncoding">The value of the header to add.</param>
+    /// <remarks>If the header already exists, it will be overwritten.</remarks>
```
maciejmlocek, line 120: I don't think that the remark statement is true, because it is possible to have multiple values for "Content-Encoding" and the code under this procedure does not overwrite the whole "Content-Encoding", but adds a new one at the end of the list.

Modules/System Tests/AL Rest Client/src/ALHttpContentTests.Codeunit.al
```diff
@@ -0,0 +1,606 @@
...
+        ALHttpContent: Codeunit "AL Http Content";
+        HttpContent: HttpContent;
+        HttpHeaders: HttpHeaders;
+        HeaderValues: List of [Text];
+        TextContent: Text;
+    begin
+        // [SCENARIO] Create AL Http Content object with text
+
+        // [GIVEN] AL Http Content created with text
+        ALHttpContent := ALHttpContent.Create('Hello World');
```
bjarkihall, line 20: Could this constructor pattern (Fluent-Interface or method-chaining pattern) become a footgun for someone not familiar with it?
This example compiles fine, but the TextContent doesn't contain the text:
```al
ALHttpContent.Create('Hello World');
TextContent := ALHttpContent.AsText();
```
Maybe it's good to mention that the return value should overwrite the caller in the summary of procedures like ALHttpContent.Create, or (if possible) if the HttpContent is by-reference you could just make the returned codeunit point to the same HttpContent instance via GetHttpContent/SetHttpContent? The third alternative would be the CurrCodeunit hack.

Modules/System/AL Rest Client/src/ALRestClient.Codeunit.al
```diff
@@ -0,0 +1,339 @@
...
+    begin
+        HttpResponseMessage := Send(Method, RequestUri, ALHttpContent.Create(Content));
+    end;
+
+    /// <summary>Sends a request with the specific Http method and the given content and content type to the specified Uri and returns the response message.</summary>
+    /// <remarks>The function fails with an error message if the request could not be sent or a response was not received.
+    /// If a response was received, then the response message object contains information about the status.</remarks>
+    /// <param name="Method">The HTTP method to use.</param>
+    /// <param name="RequestUri">The Uri the request is sent to.</param>
+    /// <param name="Content">The content to send. Valid types are: Codeunit (Temp Blob), InStream, JsonObject, JsonArray, Jsontoken, XmlDocument, Text</param>
```
bjarkihall, line 293: Here a Variant can be Codeunit "Temp Blob", but the overload suggests Codeunit "AL Http Content", so assume the explicit overload takes priority over the Variant type?
Also, the PostAsJson/PutAsJson/PatchAsJson/etc. all use overloads instead of a Variant, which seems a bit inconsistent.
I would suggest dropping the Variant type in favor of explicit types, so the compiler can help.

Modules/System Tests/AL Rest Client/src/ALRestClientTests.Codeunit.al
```diff
@@ -0,0 +1,395 @@
...
+    var
+        ALRestClient: Codeunit "AL Rest Client";
+        ALHttpResponseMessage: Codeunit "AL Http Response Message";
+        JsonObject: JsonObject;
+    begin
+        // [SCENARIO] Test GET request
+
+        // [GIVEN] An uninitialized AL Rest Client
+        // [WHEN] The Get method is called
+        ALHttpResponseMessage := ALRestClient.Get('https://httpbin.org/get');
```
bjarkihall, line 19: I'm seeing a lot of the hardcoded https://httpbin.org/ endpoint here.
I'm thinking, if httpbin.org is down, or somehow broken or simply blocking high-traffic, the autotests will fail and potentially block the pipeline, is this okay?
Would it make sense to place this baseaddress in a helper-function?
It would make it easier to switch the url, or even have a pragma to switch between a local/remote server.

Modules/System Tests/AL Rest Client/src/ALRestClientTests.Codeunit.al
```diff
@@ -0,0 +1,395 @@
...
+    procedure TestBaseAddress()
+    var
+        ALRestClient: Codeunit "AL Rest Client";
+        ALHttpResponseMessage: Codeunit "AL Http Response Message";
+        JsonObject: JsonObject;
+    begin
+        // [SCENARIO] Test GET request with base address
+
+        // [GIVEN] An initialized AL Rest Client with base address
+        ALRestClient.SetBaseAddress('https://httpbin.org');
```
bjarkihall, line 147: This tests `'https://httpbin.org' + '/get'` (the happy-path), but what about cases like:
`'https://httpbin.org/' + 'get'`
`'https://httpbin.org' + 'get'`
`'https://httpbin.org/' + '/get'`
Should the module help and make sure the url ends up having a single slash in the correct place?
Or at least in the summary comment?
Alternatively, why expose the concept of a BaseAddress and relative urls, instead of just always requiring a full/absolute url?

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
```diff
@@ -0,0 +1,153 @@
...
+    local procedure SetDefaultUserAgentHeader()
+    begin
+        SetUserAgentHeader(GetUserAgentString());
+    end;
+
+    local procedure GetUserAgentString() UserAgentString: Text
+    var
+        ModuleInfo: ModuleInfo;
+    begin
+        if NavApp.GetCurrentModuleInfo(ModuleInfo) then
```
bjarkihall, line 66: Won't this always be Microsoft + System Application, instead of the caller's module?
The HttpClient already sets a default User-Agent header if I remember correctly, so is this really needed?

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
```diff
@@ -0,0 +1,87 @@
...
+        AssertInitialized();
+        HttpRequestMessage.GetHeaders(RequestHeaders);
+        if RequestHeaders.Contains(HeaderName) then
+            RequestHeaders.Remove(HeaderName);
+        RequestHeaders.Add(HeaderName, HeaderValue);
+    end;
+
+    /// <summary>Sets the HttpRequestMessage that is represented by the AL HttpRequestMessage object.</summary>
+    /// <param name="RequestMessage">The HttpRequestMessage to set.</param>
+    procedure SetHttpRequestMessage(var RequestMessage: HttpRequestMessage)
```
bjarkihall, line 60: This is called `SetHttpRequestMessage` but the getter is called `GetRequestMessage`, shouldn't their names match?

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
```diff
@@ -0,0 +1,87 @@
...
+        AssertInitialized();
+        HttpRequestMessage.SetRequestUri(Uri);
+    end;
+
+    /// <summary>Adds a request header to the HttpRequestMessage object.</summary>
+    /// <param name="HeaderName">The name of the header to add.</param>
+    /// <param name="HeaderValue">The value of the header to add.</param>
+    /// <remarks>If the header already exists, it will be overwritten.</remarks>
+    [NonDebuggable]
+    procedure AddRequestHeader(HeaderName: Text; HeaderValue: Text)
```
bjarkihall, line 47: `ALHttpContent` has `SetHeader` while `ALHttpRequest` has `AddRequestHeader`, shouldn't they both be `SetHeader`?
I can understand a distinction in `ALHttpClient` like `SetDefaultRequestHeader` but one could argue it should also just be `SetHeader`, since each belongs to their own codeunit.

I could also see some of the helpers setting a specific header like:
`SetUserAgentHeader(HeaderValue)`
But wouldn't this be more scalable?
`SetHeader(HeaderType: Enum "Http Header Type"; HeaderValue: Text)`
Usage:
`SetHeader("Http Header Type"::UserAgent, HeaderValue)`
This way, there's no need for specific helper functions for each known header.

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
```diff
@@ -0,0 +1,153 @@
...
+        ConnectionErr: Label 'Connection to the remote service ''%1'' could not be established.', Comment = '%1 = url, e.g. https://microsoft.com';
+        RequestFailedErr: Label 'The request failed: %1 %2', Comment = '%1 = HTTP status code, %2 = Reason phrase';
+        UserAgentLbl: Label 'Dynamics 365 Business Central - |%1| %2/%3', Locked = true, Comment = '%1 = App Publisher; %2 = App Name; %3 = App Version';
+
+    procedure Initialize()
+    begin
+        ClearAll();
+
+        HttpClient.Clear();
+        HttpClient.Timeout := 60000;
```
bjarkihall, line 23: Here the Timeout is hardcoded (without explanation for this value) and `SetTimeout` is internal, so there's no way to overwrite it.

Modules/System/AL Rest Client/src/ALRestClient.Codeunit.al
```diff
@@ -0,0 +1,339 @@
...
+    /// <remarks>The base address will be used for every request that is sent with this AL Rest Client instance.
+    /// Calls to the Get, Post, Patch, Put and Delete methods must  use a relative path which will be appended to the base address.
+    /// The AL Rest Client will be initialized if it was not initialized before.</remarks>
+    /// <param name="Url">The base address to use.</param>
+    procedure SetBaseAddress(Url: Text)
+    begin
+        if not IsInitialized then
+            Initialize();
+
+        ALRestClientImpl.SetBaseAddress(Url);
```
bjarkihall, line 49: I recall the HttpClient throwing a runtime error if some of the "default" properties were set after a request had been made, for example BaseAddress, Timeout, the Default Headers, Authorization, etc. but I didn't see any tests validating this.
Should the module clear the client if it's setting these properties after Init+Send? Or do you know if this has been somehow addressed under-the-hood?

ajkauffmann:
Thank you all for the comments so far. Much appreciated and great feedback!
I'm going to work through all of it this week and will provide a number of updates. Including an update to solve the issue with telemetry signal  RT0019 and the popup question in sandboxes for allowing or blocking outgoing web requests (which was now only once for the system app instead of per app that actually started the request).

Modules/System/AL Rest Client/src/ALHttpResponseMessage.Codeunit.al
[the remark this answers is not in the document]
ajkauffmann: I found it: https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-al-complextypes

> It's also possible to use a named return value. Internally, the exit-statement as seen in the example above causes an assignment to an allocated return value. The assignment will have a small performance cost based on the type.

It seems to apply to complex types and not really to simple types. I've just adopted named return variables as a good habit. 🤓

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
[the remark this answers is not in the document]
ajkauffmann: After some consideration, I've decided to remove the TryFunction and not throw an error at all. Instead, the SendRequest function returns true or false and the error message is inside the response codeunit.
An error occurring during setting the authorization header now throws an error immediately. That's considered to be a runtime error, something I can't handle.
If the outgoing call is blocked by the environment or fails because of no connection to the server, then this will also result in an error.
As a result, only when a response was received there will be no error message. This is similar to how the HttpClient works out of the box.

Modules/System/AL Rest Client/app.json
ajkauffmann: Changed to "Rest Client".

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
ajkauffmann: Removed the prefix "AL" everywhere.

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
ajkauffmann: I agree, the Create method has now been replaced by overloads for every supported type.

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
ajkauffmann: Removed the initialization check, so this is not relevant anymore.

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
ajkauffmann: But it also means that the Send method on codeunit "Rest Client" needs to have the same overloads.

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
ajkauffmann: Removed the Initialized pattern on the "Request Message" and "Response Message" codeunits.
But kept it on the "Rest Client" codeunit because it needs to be initialized with default values.

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
ajkauffmann: Applied the Facade pattern to these codeunits:
* Http Content
* Http Request Message
* Http Response Message

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
ajkauffmann: The code has been redesigned to not handle an error from authorization. Instead, the error will just be raised. It is a runtime error anyway, and there is no need to handle it differently than any other error.
The other error conditions are also not raising an error anymore. If the connection fails, either because of outgoing calls being blocked or not receiving a response, then the SendRequest method returns false and an error is raised from the public Rest Client codeunit.
The error message for unsuccessful HTTP status is only used from the Json methods of the public Rest Client codeunit.
After all, this is in line with the default behavior of the HttpClient.

ajkauffmann: The methods of the public Rest Client codeunit could send telemetry. But it would only end up in the customers telemetry and not in the ISV's telemetry (apart from Microsoft).
This is up for discussion, please advise.

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
ajkauffmann: Added to every codeunit.

Modules/System/AL Rest Client/src/ALRestClient.Codeunit.al
ajkauffmann: I'll take a look at other modules that can implement this new module. I guess the Azure Blob Storage and Azure Functions module are also candidates.

Modules/System/AL Rest Client/src/Authentication/HttpAuthenticationBasic.Codeunit.al
ajkauffmann: I've updated the platform version to 23 and runtime version to 12.0 to support SecretText.
Changed Password to SecretText and also an overload function for Http Headers of type SecretText.
Because the password needs to be converted to Base64 for basic authentication, it could potentially leak from the Base64 codeunit. I've removed the dependency on the Base64 module and created a small function to convert SecretText to Base64 while keeping it a SecretText.

Modules/System/AL Rest Client/src/ALHttpContent.Codeunit.al
ajkauffmann: Nice catch! I've changed the comment.

Modules/System/AL Rest Client/src/ALRestClient.Codeunit.al
ajkauffmann: Yes, agree. It was also mentioned by @IhorHandziuk, so I guess this is the way to go. I've dropped the Variant parameter in favor of explicit types.

Modules/System Tests/AL Rest Client/src/ALRestClientTests.Codeunit.al
ajkauffmann: First of all, this is exposed because the HttpClient features this.
Unfortunately, the HttpRequestMessage in AL does not support relative URLs. While the DotNet object does support it, so this is simply a overlooked feature.
But for this reason, I combine both base address and the request uri when constructing the HttpRequestMessage. Because I don't know how somebody wants to use it, I just combine the two. It's up to the developer to provide correct values, I'm not going to babysit them. 😊

The option of using a base address can be very useful in case the Rest Client is used in a codeunit that supports multiple API endpoints of a server. The Rest Client can be initialized one time with the base address and then the functions can use the relative URL. This is actually the best way of using the Http Client anyway.

Modules/System Tests/AL Rest Client/src/ALRestClientTests.Codeunit.al
ajkauffmann: I expect Microsoft to disable this test eventually, also when using a local server. The same is done in the Azure Blob Storage module with tests that use Azurite.

So this is only to be used for manual testing, which is what I do anyway. The httpbin.org service can run as a docker container, but I didn't bother to go that route.

After all, I now have a Client Handler in place, that supports fake responses. However, using it for these tests would mean I'd just test my own fake responses. It is very useful for testing real-life scenarios though.

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
ajkauffmann: Yes, this would be the System Application if the caller doesn't overwrite it.
The HttpClient does not set a default User-Agent header, that's why I chose to do it. Just like other modules in the System Application that use the HttpClient.

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
ajkauffmann: I agree, the Get procedure has been renamed.

Modules/System/AL Rest Client/src/ALHttpRequestMessage.Codeunit.al
ajkauffmann: The name has now been aligned on Http Content and Http Request Message to be SetHeader. With an explanation that it will set a new value for an existing header or add the header if it does not exists.
The method for the Http Client keep the name SetDefaultRequestHeader to distinguish it from the others. They work different because they are added to every request done with the same Http Client instance.

Not sure about using Enum for known header types. That would involve a huge case statement or an interface and an implementation codeunit for every single enum value to return the correct header name. I think we should keep the current code and see if there is need to come up with an Enum in the future.

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
ajkauffmann: I agree. The default is 100.000 (100 seconds), there is actually no reason to set it to another value.
I've added functions to set and get the timeout.

ajkauffmann:
Just create a fairly large commit to resolve almost all comments.

But most importantly, I've introduced an interface "Http Client Handler". A codeunit "Default Client Handler" is also provided, that can also serve as boilerplate code for apps. The main purpose of this handler codeunit is to let the original app send the actual http request. As a result, telemetry signal RT0019 will be sent to the ISV's AppInsight and the feature to block outgoing web requests is applied to the original app.

The Http Client Handler must be specified when initializing the Rest Client. In case no handler codeunit is specified, it will take the default one provided in the module.

This pattern can also be used for tests to create fake responses or a mocking codeunit to handle the request differently.

Modules/System/Rest Client/src/RestClient.Codeunit.al
```diff
@@ -0,0 +1,403 @@
...
+    /// <summary>Sends a GET request to the specified Uri and returns the response content as JsonToken.</summary>
+    /// <remarks>The function fails with an error message if the request could not be sent or a response was not received.
+    /// The function also fails in case the response does not contain a success status code or a valid JSON content.</remarks>
+    /// <param name="RequestUri">The Uri the request is sent to.</param>
+    /// <returns>The response content as JsonToken</returns>
+    procedure GetAsJson(RequestUri: Text) JsonToken: JsonToken
+    var
+        HttpResponseMessage: Codeunit "Http Response Message";
+    begin
+        HttpResponseMessage := Send(Enum::"Http Method"::GET, RequestUri);
```
grobyns, line 224: seems like this should move into Impl.

Modules/System/Rest Client/src/RestClient.Codeunit.al
```diff
@@ -0,0 +1,403 @@
...
+    /// The function also fails in case the response does not contain a success status code or a valid JSON content.</remarks>
+    /// <param name="RequestUri">The Uri the request is sent to.</param>
+    /// <param name="Content">The content to send as a JsonToken.</param>
+    /// <returns>The response content as JsonToken</returns>
+    procedure PostAsJson(RequestUri: Text; Content: JsonToken) Response: JsonToken
+    var
+        HttpResponseMessage: Codeunit "Http Response Message";
+        HttpContent: Codeunit "Http Content";
+    begin
+        HttpResponseMessage := Send(Enum::"Http Method"::POST, RequestUri, HttpContent.Create(Content));
```
grobyns, line 264: seems like this should move to the Impl.

grobyns: same for the other instances below...

Modules/System/Rest Client/src/RestClient.Codeunit.al
```diff
@@ -0,0 +1,403 @@
...
+codeunit 2350 "AL Rest Client"
+{
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
+    var
+        RestClientImpl: Codeunit "Rest Client Impl.";
+        DefaultHttpClientHandler: Codeunit "Http Client Handler";
+        HttpAuthenticationAnonymous: Codeunit "Http Authentication Anonymous";
+        IsInitialized: Boolean;
```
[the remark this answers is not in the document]
ajkauffmann: Agreed, that would be better. Has been changed.

Modules/System/Rest Client/src/RestClient.Codeunit.al
ajkauffmann: Can you please elaborate? Should a Facade codeunit not use any of its own functions or overloads?

ajkauffmann:
@JesperSchulz It seems the pipeline is not ready for platform 12.0 at the moment.

Modules/System/Rest Client/src/RestClient.Codeunit.al
grobyns: A façade can use its own methods but shouldn't contain logic so really, when it calls its own methods its generally only calling another overload with some default value.

Modules/System/Rest Client/src/RestClient.Codeunit.al
ajkauffmann: Moved to implementation codeunit.

Modules/System/Rest Client/src/RestClient.Codeunit.al
ajkauffmann: Moved all logic to implementation codeunit.

Modules/System Tests/AL Rest Client/src/ALHttpContentTests.Codeunit.al
bjarkihall: Yes, I'm in favor of supporting both, the example you showed was exactly what I had in mind.

Modules/System Tests/Rest Client/app.json
```diff
@@ -0,0 +1,54 @@
+{
+  "id": "ae153cbb-ad55-447c-9226-5af3ef57280f",
+  "name": "AL Rest Client Tests",
```
bjarkihall, line 3: You might want to call this "Rest Client Tests", since the System module is called "Rest Client"

bjarkihall: The PR had some good improvements added and it's a nice module in general, I gave it another quick round of review and didn't see anything outstanding, so now I'm just looking forward to get a standard module for Http.
Well done and thanks for the contribution!

ajkauffmann: Renamed the test app.

ajkauffmann:
@JesperSchulz Just wanted to inform you that I've increased the object range to 2360 in order to include the extra codeunits that were needed for implementing the facade pattern.

IhorHandziuk (approved):
Great job on addressing all the PR comments! 🙂

ajkauffmann:
@JesperSchulz I was thinking, would it make sense to uptake namespaces? The base app seems to do that already, but I've not seen it in the system app. Would it make sense to put this module in a namespace like System.Http or something like that?

JesperSchulz:
The system application is fully namespaced in 23.0, and as a matter of fact, this module will have to be namespaced in order to be able to get checked in. And yes, it needs to start with "system." 😉 I can add the namespaces, when we start processing this one. Right now I've requested a few more code reviews from our platform teams, to ensure that they're fine with this wrapper. Once they've given the "go ahead", it gets interesting. I am cautiously hoping to add the "processing-PR" tag next week!

JesperSchulz:
> @JesperSchulz It seems the pipeline is not ready for platform 12.0 at the moment.

It is not, no. But don't sweat it. As long as it compiles on 2023 Wave 2 / 12.0, it's all good. Hopefully I'll soon be able to give this a spin in our internal build system!

ajkauffmann:
> > @JesperSchulz It seems the pipeline is not ready for platform 12.0 at the moment.
> 
> It is not, no. But don't sweat it. As long as it compiles on 2023 Wave 2 / 12.0, it's all good. Hopefully I'll soon be able to give this a spin in our internal build system!

I've tested with a 23.0 preview build, and all tests succeed. It's great to have the SecretText implemented from the beginning, that's why I needed v23.

Modules/System/Rest Client/src/HttpMethod.Enum.al
```diff
@@ -0,0 +1,68 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+/// <summary>
+/// This enum contains the REST Http Methods.
+/// </summary>
+enum 2350 "Http Method"
+{
+    Extensible = true;
```
duichwer, line 11: Is this extensibility of this ENUM really intended?

ajkauffmann:
One small issue is missing, which was reported to @KennieNP a while ago: the HttpRequest object in the AL platform doesn't support relative URLs, while the underlying .Net object does. I've worked around it, but I'd be more happy if it would be supported in the platform object.

Modules/System/Rest Client/src/RestClient.Codeunit.al
duichwer, line 1: It would be useful if the examples contained in the PR description are added to this Codeunit documentation as example.
Then it'll be easier to develop with this new Codeunit.
There should be a remark on the new HTTP Handler for telemetry.

Modules/System/Rest Client/src/HttpMethod.Enum.al
ajkauffmann: Not really. The enum was copied from the OAuth module, and I kept everything the same.
But thinking about it, for REST you use 5 HTTP methods: GET, POST, PATCH, PUT, DELETE
That also means that the HEAD and OPTIONS methods can be removed. Would you agree?

When the module gets its own namespace, then another module could define a Http Method enum in its own namespace in case other HTTP methods are needed.

Modules/System/Rest Client/src/RestClient.Codeunit.al
ajkauffmann: Yes, some documentation would be useful. Even better if other modules use this module, then you have real-world examples.

JesperSchulz:
@KennieNP also suggested we add some retry-logic to this wrapper, while at it (I could not spot it when rushing through the code today). It would greatly improve the HttpClient call stability in the app if we did. He even provided some sample code to illustrate his intend ;-)

`

    // RetryStrategy can be any of these strings (should be an enumeration)
    // RegularIntervalRetry
    // ExponentialBackoff
    // Randomization


    local procedure SendRequest(
        Client: HttpClient;
        RequestMessage: HttpRequestMessage;
        UserErrorMessage: Text[200];
        IsCollectable: boolean;
        MaxNumberOfRetries: Integer;
        RetryStrategy: Text;
        RetrySleeptimeInMilliseconds: Integer
    ) ResponseMessage: HttpResponseMessage
    var
        IsSuccessful: Boolean;
        ErrorInfoObject: ErrorInfo;
        ErrorCustomDimensions: Dictionary of [Text, Text];
        vResponseMessage: HttpResponseMessage;
        HttpStatusCode: integer;
    begin
        IsSuccessful := Client.Send(RequestMessage, vResponseMessage);

        if not IsSuccessful then begin
            ErrorInfoObject.DetailedMessage := 'The HttpClient.Send could not be sent, this is likely an error in the AL code.';
            ErrorInfoObject.Title := 'Could not call external service';
            ErrorInfoObject.Message := UserErrorMessage;
            ErrorInfoObject.Collectible := IsCollectable;
            ErrorCustomDimensions.Add('HTTPMethod', RequestMessage.Method);
            ErrorInfoObject.CustomDimensions := ErrorCustomDimensions;
            Error(ErrorInfoObject);
        end;

        if not vResponseMessage.IsSuccessStatusCode() then begin
            HttpStatusCode := vResponseMessage.HttpStatusCode();

            // Handling status codes 429 and 503 requires the client to adopt a retry logic while providing a cool off period. 
            case HttpStatusCode of
                419:
                    vResponseMessage := SendRequestRetry(Client, RequestMessage, MaxNumberOfRetries, RetryStrategy, RetrySleeptimeInMilliseconds);
                503:
                    vResponseMessage := SendRequestRetry(Client, RequestMessage, MaxNumberOfRetries, RetryStrategy, RetrySleeptimeInMilliseconds);
            end;

            if not vResponseMessage.IsSuccessStatusCode() then begin
                HttpStatusCode := vResponseMessage.HttpStatusCode();
                // final error handling after possible retries
                if ((HttpStatusCode = 429) or (HttpStatusCode = 503)) then begin
                    ErrorInfoObject.DetailedMessage := 'The HttpClient.Send call was successful, but the external service responded with a ' + Format(HttpStatusCode) + ' HTTP status code even after retrying.';
                    ErrorCustomDimensions.Add('RetryAttempts', Format(MaxNumberOfRetries));
                end;
                if (HttpStatusCode = 504) then begin
                    ErrorInfoObject.DetailedMessage := 'The HttpClient.Send call was successful, but the external service timed out (responded with a ' + Format(HttpStatusCode) + ' HTTP status code). You probably need to refactor your AL code.';
                end;
                if (HttpStatusCode > 500) then begin
                    ErrorInfoObject.DetailedMessage := 'The HttpClient.Send call was successful, but the external service responded with a 5xx error. This call will not be retried.';
                end;

                ErrorInfoObject.Title := 'Could not call external service';
                ErrorInfoObject.Message := UserErrorMessage;
                ErrorInfoObject.Collectible := IsCollectable;
                ErrorCustomDimensions.Add('HttpStatusCode', Format(HttpStatusCode));
                ErrorCustomDimensions.Add('HTTPMethod', RequestMessage.Method);
                ErrorInfoObject.CustomDimensions := ErrorCustomDimensions;
                Error(ErrorInfoObject);
            end
            else // retry worked! :)
                ;

        end;

        ResponseMessage := vResponseMessage
    end;



    // Asserts that Client.Send(RequestMessage, vResponseMessage) is successfull (that the AL code is sound)
    local procedure SendRequestRetry(
        Client: HttpClient;
        RequestMessage: HttpRequestMessage;
        MaxAttempts: Integer;
        RetryStrategy: Text;
        SleeptimeInMilliseconds: Integer
    ) ResponseMessage: HttpResponseMessage
    var
        vResponseMessage: HttpResponseMessage;
        Attempt: Integer;
        IsSuccessful: Boolean;
    begin
        Attempt := 1;
        while Attempt <= MaxAttempts do begin
            IsSuccessful := Client.Send(RequestMessage, vResponseMessage);
            if (IsSuccessful) then
                break
            else begin
                case RetryStrategy of
                    'RegularIntervalRetry':
                        RegularIntervalRetry(SleeptimeInMilliseconds);
                    'ExponentialBackoff':
                        ExponentialBackOff(SleeptimeInMilliseconds, Attempt);
                    'Randomization':
                        Random(SleeptimeInMilliseconds);
                    else
                        RegularIntervalRetry(SleeptimeInMilliseconds);
                end;
                // if RetryStrategy = 'RegularIntervalRetry' then 
                // else if ;
            end;
            ;
            Attempt := Attempt + 1;
        end;
        ResponseMessage := vResponseMessage
    end;

    local procedure RegularIntervalRetry(SleeptimeInMilliseconds: Integer)
    begin
        Sleep(SleeptimeInMilliseconds);
    end;

    local procedure ExponentialBackOff(SleeptimeInMilliseconds: Integer; Attempt: Integer)
    begin
        Sleep(SleeptimeInMilliseconds * Power(2, Attempt));
    end;

    local procedure Random(SleeptimeInMilliseconds: Integer)
    begin
        Sleep(System.Random(SleeptimeInMilliseconds));
    end;
}
`

This doesn't have to be part of the initial PR, but we'll probably add it eventually, as it's a great suggestion.

ajkauffmann:
That's a great suggestion! I agree it can be added later. If time permits I may add it to the current PR.
For the retry strategy: if that's an enum, we could consider adding an interface to it. Then it can easily be extended by others with alternative strategies.
You may want to already increase the number range so we have enough objects reserved! 😄

JesperSchulz:
Feel free to extend the number range 😊 There should be plenty of available numbers...
Love the idea of making the retry strategy extensible. I guess retying can be done in many different ways.
Adding a retry strategy to the module will further boost the value proposition of this module. We are indeed seeing a lot of failures in telemetry due to the lack of this capability.
This module is coming along nicely 🥳 I'd love to get it on the road with 23.1 (November release) - let's see if we can make that happen!

Modules/System/Rest Client/src/HttpContent.Codeunit.al
```diff
@@ -0,0 +1,279 @@
...
+    begin
+        HttpContent := Create(Content, '');
+    end;
+
+    /// <summary>Initializes a new instance of the Http Content class with the specified "Temp Blob" Codeunit content and content type.</summary>
+    /// <param name="Content">The content to send to the server.</param>
+    /// <param name="ContentType">The content type of the content to send to the server.</param>
+    /// <returns>The Http Content object.</returns>
+    /// <remarks>If the ContentType parameter is not specified, it will be set to 'application/octet-stream'.</remarks>
+    procedure Create(Content: Codeunit "Temp Blob"; ContentType: Text) HttpContent: Codeunit "Http Content"
```
pri-kise, line 105: I think this procedure is missing documentation on the Default Encoding UTF8 that is used to read the tempBlob.
Maybe an overload where we can specify the Encoding might be useful.

pri-kise:
over all this PR looks very promising.
Personally I'd like to have an Generi POST method on the rest client, that accepts an InStream, but as a starter the current Generic Send is okay, too.

JesperSchulz:
Short update: Tomorrow we are internally opening up for PRs through the new BCApps repository. I will make this PR the first official AL PR to be processed in the new repository 🥳 From November 1st, the BCApps repository will open to the public too. By then, I hope this PR has successfully been processed though.

Modules/System Tests/Rest Client/app.json
```diff
@@ -0,0 +1,54 @@
...
+  "url": "https://go.microsoft.com/fwlink/?LinkId=72401",
+  "dependencies": [
+      {
+        "id": "812b339d-a9db-4a6e-84e4-fe35cbef0c44",
+        "name": "Rest Client",
+        "publisher": "Microsoft",
+        "version": "24.0.0.0"
+      },
+      {
+          "id":  "5d86850b-0d76-4eca-bd7b-951ad998e997",
```
JesperSchulz, line 20: This dependency is from the Base App. The System App may not take dependencies on the layers above. I'll see if I can't remove that dependency.

JesperSchulz: Apparently wasn't needed at all.

Modules/System/Rest Client/src/HttpContent.Codeunit.al
JesperSchulz: Good suggestion. I've chosen to get this PR processed and into the main branch, as we've got a few other PRs waiting to uptake this one. @ajkauffmann, you might want to consider this comment for a follow-up PR (once we switch to the main branch, where you'll find your submitted module 🥳).

Modules/System/Rest Client/src/RestClient.Codeunit.al
JesperSchulz: We'll do some uptake before 24.0 is released!

Modules/System/AL Rest Client/src/ALRestClientImpl.Codeunit.al
JesperSchulz: Let's address this in a followup PR. We'll go with this PR as the initial version.

JesperSchulz (approved):
This is in a good enough shape to get submitted. Further improvements can be made in chaser PRs. With this module in, we can start uptake 🥳

JesperSchulz (approved):

JesperSchulz:
The PR went into the main branch (24.0) 🥳 As soon as we release the 23 bits, we can merge this PR. Further improvements can then be done to the added module.

mazhelez (approved):

PR #24875, opened 2023-09-20
Add metadata to Sharepoint connector

lorenzfleerackers:
I intend to enhance the existing SharePoint connector by incorporating metadata functionality. 

I have implemented this enhancement by introducing new events and enabling extensibility in the tables. As a result, when establishing a connection, I can now seamlessly integrate my metadata fields into the appropriate tables and populate them with data through events in the "SharePoint Client" codeunit.

### DIFF
--- a/Modules/System/SharePoint/src/SharePointClientImpl.Codeunit.al
+++ b/Modules/System/SharePoint/src/SharePointClientImpl.Codeunit.al
@@ -358,13 +358,13 @@ codeunit 9101 "SharePoint Client Impl."
         SharePointUriBuilder.SetObject('lists');
 
         Metadata.Add('type', 'SP.List');
-        SharepointEvents.OnAfterAddListMetaData(Metadata);
         Request.Add('__metadata', Metadata);
         Request.Add('AllowContentTypes', true);
         Request.Add('BaseTemplate', 100);
         Request.Add('ContentTypesEnabled', true);
         Request.Add('Description', ListDescription);
         Request.Add('Title', ListTitle);
+        SharepointEvents.OnAfterAddListMetaData(Request);
 
         SharePointHttpContent.FromJson(Request);
         SharePointHttpContent.SetRequestDigest(GetRequestDigest(SharePointUriBuilder.GetHost()));
@@ -392,9 +392,9 @@ codeunit 9101 "SharePoint Client Impl."
         SharePointUriBuilder.SetObject('items');
 
         Metadata.Add('type', ListItemEntityTypeFullName);
-        SharepointEvents.OnAfterAddListItemMetaData(Metadata);
         Request.Add('__metadata', Metadata);
         Request.Add('Title', ListItemTitle);
+        SharepointEvents.OnAfterAddListItemMetaData(Request);
 
         SharePointHttpContent.FromJson(Request);
         SharePointHttpContent.SetRequestDigest(GetRequestDigest(SharePointUriBuilder.GetHost()));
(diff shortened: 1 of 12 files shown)


pri-kise (commented):
What are you trying to achive by these events?
Why can't this be added to the module itself?

And you added events to an internal object - therefore those events won't be subscribable outside of the system application.
If you need events then you have to add them on a public codeunit and make them internal to call them from an an internal object.

lorenzfleerackers:
> What are you trying to achive by these events? Why can't this be added to the module itself?
> 
> And you added events to an internal object - therefore those events won't be subscribable outside of the system application. If you need events then you have to add them on a public codeunit and make them internal to call them from an an internal object.


What do you mean by "adding it to the module itself?"

That's correct, I hadn't noticed that they were internal codeunits. I've made a change so that the events are now located in a public codeunit.

Modules/System/SharePoint/src/SharePointClientImpl.Codeunit.al
```diff
@@ -364,6 +364,7 @@ codeunit 9101 "SharePoint Client Impl."
         Request.Add('ContentTypesEnabled', true);
         Request.Add('Description', ListDescription);
         Request.Add('Title', ListTitle);
+        SharepointEvents.OnAfterAddListMetaData(Request);
```
grobyns, line 367: Adding events in the middle of methods is comparable to code customization. It's not a good pattern to follow. Also just consider that events are generally raised and taken out of context. You now need to recreate that context in a subscriber and figure out if you need to act. The more context is passed in the event the better but it will never be very good.

An alternative to this approach (as was suggested you look into) could be to change the methods so they take in a dictionary for example. You can then pass in all your data from the start and there's no need for events. It's a bit more work and you need to take care to not break the existing public methods but I think the end result will be much better.

JesperSchulz:
Rather than adding extension points, you should add the needed metadata to the module itself by creating a PR which adds the required code. We should not have several other apps add to the behavior. All apps depending on this module should understand what it is capable of.
You are very welcome to create a new PR where you are adding the capabilities you desire to this module. We don't want to introduce any extensibility at this point, unless of course your arguments for doing so are very sound.

JesperSchulz (changes requested):
PR in current shape rejected. We do not want to introduce extensibility points at the suggested places. Rather, the module itself should get improved with the desired capabilities.

lorenzfleerackers:
> Rather than adding extension points, you should add the needed metadata to the module itself by creating a PR which adds the required code. We should not have several other apps add to the behavior. All apps depending on this module should understand what it is capable of. You are very welcome to create a new PR where you are adding the capabilities you desire to this module. We don't want to introduce any extensibility at this point, unless of course your arguments for doing so are very sound.

If I adjust it as you suggest, then it's not variable, is it? Because each customer will use different metadata.

pri-kise:
I think it would be helpful if you'd provide some example requests, that you're trying to send.
Currently this module uses the  (legacy) REST API V1 of the sharepoint, which doesn't have this much documentation.

Therefore it's somehow difficult to understand what exactly is missing.

Additionally you might be right, that processing the respone needs an event to populate the response accordingly, but if you provide an real example response then it might be easier to understand.

Furthermore maybe it's possible to use a generic variable like the dictionary to type in AL to improve the procedure `CreateList`, that this methods accepts custom parameters. Then it's easier for others to populate custom sharepoint fields.

I'm only able to find the following documentation
https://learn.microsoft.com/en-us/sharepoint/dev/sp-add-ins/working-with-lists-and-list-items-with-rest#working-with-lists-by-using-rest
If i understand you correctly, then it's possible to provide here more in the request body??
Could you please provide an example and explain this example a more in depth?

JesperSchulz:
As this PR is rejected in its current form, I will close it (we're trying to empty this repo, as we're moving to a new location). Please create a new PR in the new https://github.com/microsoft/BCApps repository, once you're ready. Please reference this PR when you do for history. Thanks!

PR #24899, opened 2023-09-25
Allowed to set more Environment Infos in tests

jonaswre:
This PR allows the developer to set AppId and EnvironmentName.

It also adds the ability to set IsSaaS and IsSandbox to false instead of setting it to true or not setting it.

### DIFF
--- a/Modules/System Test Libraries/Environment Information/src/EnvironmentInfoTestLibrary.Codeunit.al
+++ b/Modules/System Test Libraries/Environment Information/src/EnvironmentInfoTestLibrary.Codeunit.al
@@ -19,11 +19,24 @@ codeunit 135094 "Environment Info Test Library"
     /// This functions should only be used for testing purposes.
     /// </remarks>
     /// <param name="EnableSandboxForTest">The value to be set to the testability sandbox flag.</param>
-    [Scope('OnPrem')]
     procedure SetTestabilitySandbox(EnableSandboxForTest: Boolean)
     begin
+        EnvironmentInformationImpl.SetTestMode(true);
         EnvironmentInformationImpl.SetTestabilitySandbox(EnableSandboxForTest);
     end;
+
+    /// <summary>
+    /// Sets the test mode flag.
+    /// </summary>
+    /// <remarks>
+    /// This functions should only be used for testing purposes.
+    /// </remarks>
+    /// <param name="EnableTests">The value to be set to the test flag.</param>
+    procedure SetTestability(EnableTests: Boolean)
+    begin
+        EnvironmentInformationImpl.SetTestMode(EnableTests);
+    end;
+
 
     /// <summary>
     /// Sets the testability SaaS flag.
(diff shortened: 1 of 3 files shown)


Modules/System Test Libraries/Environment Information/src/EnvironmentInfoTestLibrary.Codeunit.al
```diff
@@ -19,30 +19,63 @@ codeunit 135094 "Environment Info Test Library"
...
+    begin
+        EnvironmentInformationImpl.SetTestMode(true);
+        EnvironmentInformationImpl.SetTestabilityAppId(NewAppId);
+    end;
+
+    /// <summary>
+    /// Sets the Environment Name that of the current Environment.
+    /// </summary>
+    /// <param name="NewEnvironmentName">The desired new Environment Name.</param>
+    procedure SetTestabilityEn(NewEnvironmentName: Text)
```
pri-kise, line 68: I think the name of this procedure should be `SetTestabilityEnvironmentName`?

jonaswre: Yes correct, i've fixed that...

JesperSchulz:
@nikolakukrika, could you review this PR?

Modules/System/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
```diff
@@ -62,12 +68,27 @@ codeunit 3702 "Environment Information Impl."
...
+    begin
+        TestabilityAppId := AppId;
+    end;
+
+    procedure SetTestabilityEnvironmentName(Name: Text)
+    begin
+        TestabilityEnvironmentName := Name;
+    end;
+
+    procedure SetTestMode(Mode: Boolean)
```
nikolakukrika, line 81: It would be better to use an internal event so we can bind to it from test. It is cleaner because we can bind/unbind the manual subscriber from the test. With the global the setting will last for the entire session and it can affect the other tests. We cannot cleanup in case of the failure, thus a failing test could have set the value and it would stay for all other tests that will run within that session.

Modules/System/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
```diff
@@ -14,6 +14,9 @@ codeunit 3702 "Environment Information Impl."
         NavTenantSettingsHelper: DotNet NavTenantSettingsHelper;
         TestabilitySoftwareAsAService: Boolean;
         TestabilitySandbox: Boolean;
+        TestabilityAppId: Text;
```
nikolakukrika, line 17: Same comments here, we should move these to internal events.

Modules/System Test Libraries/Environment Information/src/EnvironmentInfoTestLibrary.Codeunit.al
```diff
@@ -19,30 +19,63 @@ codeunit 135094 "Environment Info Test Library"
     /// This functions should only be used for testing purposes.
     /// </remarks>
     /// <param name="EnableSandboxForTest">The value to be set to the testability sandbox flag.</param>
-    [Scope('OnPrem')]
```
nikolakukrika, line 22: I believe it is fine to move these away from OnPrem scope as we do not publish the libraries in SaaS. Will need to include the security people to review.

Modules/System/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
nikolakukrika: With this approach we would need a new codeunit in which you can set the parameters and bind it.

nikolakukrika (changes requested):

Modules/System/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
grobyns: FYI: remember that internals should not be exposed directly to test code. It should be exposed to a test library which then has a public method for the tests to use.

Modules/System Tests/Environment Information/src/EnvironmentInformationTest.Codeunit.al
```diff
@@ -52,8 +49,20 @@ codeunit 135091 "Environment Information Test"
...
+    procedure TestIsSandboxIsFalseWhenTestabilitySandboxIsNotSet()
+    begin
+        // [Scenario] Set the testability to false. IsSandBox returns correct values.
+
+        // [Given] Set the testability sandbox to false
+        EnvironmentInfoTestLibrary.SetTestability(false);
+
+        // [When] Poll for IsSandbox
+        // [Then] Should return false
+        Assert.IsFalse(EnvironmentInformation.IsSandbox(), 'Testability should have dictacted a non-sandbox environment');
```
grobyns, line 61: typo

Modules/System Tests/Environment Information/src/EnvironmentInformationTest.Codeunit.al
```diff
@@ -78,5 +86,18 @@ codeunit 135091 "Environment Information Test"
         // [Then] Should return false
         Assert.IsFalse(EnvironmentInformation.IsSaaS(), 'Testability should have dictacted a non- SaaS environment');
```
grobyns, line 87: This typo was made everywhere... I'm guessing 'dictated' is what was meant.

jonaswre: I will fix it when I implement the rest.

JesperSchulz:
@jonaswre, it looks like this PRs isn't too far away from getting approved! I looks like a few small adjustments are needed, and then it's ready to get processed 🥳
In the meantime, our System Application has been moved from our internal repository / ALAppExtensions, to the fully open sourced repo https://github.com/microsoft/BCApps/pulls. In order for this to get processed, it must be moved to BCApps.
Could you do that, as I would prefer that you remain the author of this change. I could do it too, but then my name is on this PR - which wouldn't be right.

jonaswre:
@JesperSchulz I'll make the fixes requested in this PR and then open another PR in the new BCApps repo.

Thanks for the heads up

JesperSchulz:
Awesome. When done, please close this PR. Thanks a ton!

jonaswre:
I've implemented the requested changes and migrated to the BCApps repo.

https://github.com/microsoft/BCApps/pull/342

PR #24903, opened 2023-09-25
Parse blob names with hierarchical namespace

### LINKED ISSUE #24891: ABSBlobClient.ListBlobs(ABSContainerContent) duplicates directory names in storage accounts with hierarchical namespace
**Steps to reproduce**

1. Create an Azure storage account with Azure Data Lake Gen2 capabilities, make sure that the hierarchical namespace is enabled
image

2. Create a new container under this account
3. Open the container in the container browser on the Azure portal and create a directory ("Add Directory" action).
4. Upload a file into the directory.
5. Invoke the procedure ListBlobs(ABSContainerContent) from the codeunit "ABS Blob Client". Sample code is below.

```
    procedure ListBlobsInContainer(ContainerName: Text; var ABSContainerContent: Record "ABS Container Content")
    var
        ABSBlobClient: Codeunit "ABS Blob Client";
        ABSOperationResponse: Codeunit "ABS Operation Response";
        Authorization: Interface "Storage Service Authorization";
    begin
        Authorization := StorageServicesAuthorization.CreateSharedKey(GetSharedKey());
        ABSBlobClient.Initialize(GetStorageAccount(), ContainerName, Authorization);
        ABSOperationResponse := ABSBlobClient.ListBlobs(ABSContainerContent);
    end;
```
7. The returned list of BLOBs in the ABSContainerContent table contains two entries for the directory.

image

**Expected result**
The list in the screenshot must contain two entries for block blobs "foo" "foo/bar.txt".

**Cause**
"ABS Container Content Helper" splits the blob name interpreting each slash symbol as a delimiter for a  directory name, and returns a separate entry for each slash-delimited substring. This approach successfully imitates the directory structure for the flat container namespace. But if the hierarchical namespace is enabled, the API response already contains folder names, so the splitting of the BLOB name duplicates entries.
Other overloads of ListBlobs do not have this issue, because they do not call the splitting method and return the node list from the API response as is.

Here is the API response from Postman for the reference
```
<?xml version="1.0" encoding="utf-8"?>
<EnumerationResults ServiceEndpoint="https://***.blob.core.windows.net/" ContainerName="samplecontainer">
    <Blobs>
        <Blob>
            <Name>foo</Name>
            <Properties>
                <Creation-Time>Sat, 23 Sep 2023 21:04:18 GMT</Creation-Time>
                <Last-Modified>Sat, 23 Sep 2023 21:04:18 GMT</Last-Modified>
                <Etag>0x8DBBC78A6E95AF7</Etag>
                <ResourceType>directory</ResourceType>
                <Content-Length>0</Content-Length>
                <Content-Type>application/octet-stream</Content-Type>
                <Content-Encoding />
                <Content-Language />
                <Content-CRC64>AAAAAAAAAAA=</Content-CRC64>
                <Content-MD5 />
                <Cache-Control />
                <Content-Disposition />
                <BlobType>BlockBlob</BlobType>
                <AccessTier>Hot</AccessTier>
                <AccessTierInferred>true</AccessTierInferred>
                <LeaseStatus>unlocked</LeaseStatus>
                <LeaseState>available</LeaseState>
                <ServerEncrypted>true</ServerEncrypted>
            </Properties>
            <OrMetadata />
        </Blob>
        <Blob>
            <Name>foo/bar.txt</Name>
            <Properties>
                <Creation-Time>Wed, 27 Sep 2023 20:55:36 GMT</Creation-Time>
                <Last-Modified>Wed, 27 Sep 2023 20:55:36 GMT</Last-Modified>
                <Etag>0x8DBBF9C194887C4</Etag>
                <ResourceType>file</ResourceType>
                <Content-Length>1</Content-Length>
                <Content-Type>text/plain</Content-Type>
                <Content-Encoding />
                <Content-Language />
                <Content-CRC64 />
                <Content-MD5>dpT0pmMW5TyM3Z2ZVL1hHQ==</Content-MD5>
                <Cache-Control />
                <Content-Disposition />
                <BlobType>BlockBlob</BlobType>
                <AccessTier>Hot</AccessTier>
                <AccessTierInferred>true</AccessTierInferred>
                <LeaseStatus>unlocked</LeaseStatus>
                <LeaseState>available</LeaseState>
                <ServerEncrypted>true</ServerEncrypted>
            </Properties>
            <OrMetadata />
        </Blob>
    </Blobs>
    <NextMarker />
</EnumerationResults>
```

adrogin:
Resolves #24891 

Proposed solution. more tests to come.


Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
```diff
@@ -611,6 +611,68 @@ codeunit 132920 "ABS Blob Client Test"
...
+
+        // [THEN] "ABS Container Content" contains two records for folders and one record "blob"
+        Assert.RecordCount(ABSContainerContent, 3);
+
+        VerifyContinerContentType(ABSContainerContent, ABSTestLibrary.GetSampleResponseRootDirName(), Enum::"ABS Blob Resource Type"::Directory);
+        VerifyContinerContentType(ABSContainerContent, ABSTestLibrary.GetSampleResponseSubdirName(), Enum::"ABS Blob Resource Type"::Directory);
+        VerifyContinerContentType(ABSContainerContent, ABSTestLibrary.GetSampleResponseFileName(), Enum::"ABS Blob Resource Type"::File);
+    end;
+
+    local procedure VerifyContinerContentType(var ABSContainerContent: Record "ABS Container Content"; BlobName: Text[2048]; ExpectedResourceType: Enum "ABS Blob Resource Type")
```
pri-kise, line 667: There's one small typo. Should be `VerifyContainerContentType`

Modules/System Tests/Azure Blob Services API/src/ABSTestLibrary.Codeunit.al
```diff
@@ -152,6 +152,129 @@ codeunit 132921 "ABS Test Library"
         exit(Document);
     end;
 
+    procedure GetServiceResponseBlobWithHierarchicalName() ResponseText: Text;
+    begin
+        ResponseText := '<?xml version="1.0" encoding="utf-8"?>';
```
pri-kise, line 157: We should use StringBuilder instead of string concatenation for Creating Test Responses for performance reasons

Modules/System Tests/Azure Blob Services API/src/ABSBlobClientTest.Codeunit.al
adrogin: Thanks for catching this! Updated.

Modules/System Tests/Azure Blob Services API/src/ABSTestLibrary.Codeunit.al
adrogin: Changed concatenation to TextBuilder. Although, to be honest, I would not bother about performance too much in this case. Performance difference between the two versions will be in the order of magnitude of nanoseconds (considering that no fancy compiler optimisation happens behind the scene, which can eliminate any difference). Not a performance bottleneck, I'd say.

pri-kise (approved):

pri-kise:
lgtm. @JesperSchulz could you maybe take a look and process this internally?

JesperSchulz:
I surely can! Let's do this!

JesperSchulz (approved):
Looks good to me.

JesperSchulz (dismissed):

JesperSchulz:
Processed for 24.0. Will backport to 23.1! Will merge here as soon as build passes.

JesperSchulz (approved):

PredragMaricic (approved):

JesperSchulz:

Build ID: 12955.

JesperSchulz:

Build ID to track: 13938.

StefanMaron:
Could it be that this leads to some unexpected errors?
image
Version: US Business Central 23.3 (Platform 23.0.15562.0 + Application 23.3.14876.15103)

I dont think there really is a way for me to debug this myself with the system app still being locked down.

adrogin:
@StefanMaron There is a test covering the case of the service returning a response without the ResourceType node. But it looks like this test does not help much. Let me double-check.

PR #24967, opened 2023-10-03
Shopify allow use of customer creation and update

BernhardKloibmueller:
We would like to make a custom mapping with different fields to identify existing customers and decide based on them to create a new customer from a shopify customer.

Since the enum "Shpfy Customer Mapping" is extensible we thought this should be possible,
Unfortunately the codeunits for creating and updating customers are only accessible internally and there is no other interface to change a customer based on shopify data.

We would like to avoid writing our own Create Customer and Update Customer codeunits just because they are only internal allowed. From our point of view there is no particular reason to not allow these codeunits to be used by an extension.

If we understood the usage of a codeunit wrong I would accept any clarification.
Thank you in advance.

### DIFF
--- a/Apps/W1/Shopify/app/src/Customer/Codeunits/ShpfyCustomerImport.Codeunit.al
+++ b/Apps/W1/Shopify/app/src/Customer/Codeunits/ShpfyCustomerImport.Codeunit.al
@@ -3,8 +3,6 @@
 /// </summary>
 codeunit 30117 "Shpfy Customer Import"
 {
-    Access = Internal;
-
     trigger OnRun()
     var
         CustomerAddress: Record "Shpfy Customer Address";
@@ -49,12 +47,12 @@ codeunit 30117 "Shpfy Customer Import"
     /// Get Customer.
     /// </summary>
     /// <param name="ShopifyCustomerResult">Parameter of type Record "Shopify Customer".</param>
-    internal procedure GetCustomer(var ShopifyCustomerResult: Record "Shpfy Customer")
+    procedure GetCustomer(var ShopifyCustomerResult: Record "Shpfy Customer")
     begin
         ShopifyCustomerResult := ShopifyCustomer;
     end;
 
-    internal procedure SetAllowCreate(Value: Boolean)
+    procedure SetAllowCreate(Value: Boolean)
     begin
         AllowCreate := Value;
     end;
@@ -63,7 +61,7 @@ codeunit 30117 "Shpfy Customer Import"
     /// Set Customer.
     /// </summary>
     /// <param name="Id">Parameter of type BigInteger.</param>
-    internal procedure SetCustomer(Id: BigInteger)
+    procedure SetCustomer(Id: BigInteger)
     begin
         if Id <> 0 then begin
             Clear(ShopifyCustomer);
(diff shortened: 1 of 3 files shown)


BernhardKloibmueller:
@microsoft-github-policy-service agree company="NAVAX Consulting GmbH"

Apps/W1/Shopify/app/src/Customer/Codeunits/ShpfyCustomerImport.Codeunit.al
```diff
@@ -49,12 +47,12 @@ codeunit 30117 "Shpfy Customer Import"
...
     /// </summary>
     /// <param name="ShopifyCustomerResult">Parameter of type Record "Shopify Customer".</param>
-    internal procedure GetCustomer(var ShopifyCustomerResult: Record "Shpfy Customer")
+    procedure GetCustomer(var ShopifyCustomerResult: Record "Shpfy Customer")
     begin
         ShopifyCustomerResult := ShopifyCustomer;
     end;
 
-    internal procedure SetAllowCreate(Value: Boolean)
+    procedure SetAllowCreate(Value: Boolean)
```
pri-kise, line 55: we should have a good documentation for all procedures that aren't marked as internal.

pri-kise:
@BernhardKloibmueller could you maybe add a small example (code draft) for the extensibility?

I try to understand how we can improve the following documentation:
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-extending-shopify

with this new extensiblity.

pri-kise:
Additionally I'm not sure if you should simple remove the Internal property.

Maybe we should make use of the facade pattern and add new codeunits for this reason.

@JesperSchulz  and @AndreiPanko  for help on this.

AndreiPanko:
This is a difficult one.
I agree that it is unfortunate that while "Shpfy Customer Mapping"  is extensible, this is not used in Create/Map loggic. So new options would not make any different.

Would be great to see PR that replaces the current implementation with extensible one. That's of course more work.

Usually functions of extension are not available from outside and I would prefer to keep it this way to reduce surface. However I don't see why not to expose some of the methods via facade pattern we have (see modules in system app). 
For example I'm reluctant to open SetShop method as it is not really needed as shopify customer record now have reference to shop.

BernhardKloibmueller:
@AndreiPanko @pri-kise 
Thank you for the responses.

Then I will not make any more updates until you decide how to proceed with the topic.
When you confirm it, then I will do the changes in this PR

What would we like to achieve in detail:
We want not to reuse customers existing in business central for shopify mapping. we only want to create new customers in business central. This is because our customer has already a custom webshop with many different benefit systems which should not interfere. This behavior we now could achieve with one of the existing events which is used in the E-Mail mapping option.

But we needed to refuse a feature because we could not implement it yet. This was Mapping based on if the customer was created by shopify itself. This we wanted to make possible with the interface.

@pri-kise :
Regarding the documentation:
At them moment this is missing the customer events completely. Maybe you can add an example to fill additional customer data.

for example something like this:
```
    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Shpfy Customer Events", OnAfterCreateCustomer, '', false, false)]
    local procedure OnAfterCreateCustomer(Shop: Record "Shpfy Shop"; var ShopifyCustomerAddress: Record "Shpfy Customer Address"; var Customer: Record Customer);
    begin
        Customer."House Number" := GetHouseNumberFormAddress(ShopifyCustomerAddress."Address 1");
        Customer."Street" := GetStreetNameFormAddress(ShopifyCustomerAddress."Address 1");
        Customer.CalculateAddressFromStreetAndHouseNumber();
        Customer.Modify(false);

        ShopifyCustomerAddress."Address 1" := Customer.Address;
        ShopifyCustomerAddress.Modify(false);
    end;
```

Is this helpful or do you need more?
Thank you all in advance.

onbuyuka:
Not planned

PR #25012, opened 2023-10-09
Barud - tests added to the FA Locations and Fixed Assets APIs

barud-dyna:


barud-dyna:
Doubled with https://github.com/microsoft/ALAppExtensions/pull/24169

PR #25057, opened 2023-10-11
Removed access internal

PThiemann-AINO:
Hi,

This change allows developers to reference to the "Word Templates Related Table" table object within their own code. 

Without the internal access we are able to transfer word templates with all the related data we need between multiple environments and up to 300 companies per environment. 

Please also note the Yammer conversation between Lorenz Roth and @JesperSchulz : https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=1520243730882560

### DIFF
--- a/Modules/System/Word Templates/src/WordTemplatesRelatedTable.Table.al
+++ b/Modules/System/Word Templates/src/WordTemplatesRelatedTable.Table.al
@@ -12,7 +12,6 @@ using System.Reflection;
 /// </summary>
 table 9990 "Word Templates Related Table"
 {
-    Access = Internal;
     Extensible = false;
     InherentEntitlements = X;
     InherentPermissions = X;


pri-kise:
@PThiemann-AINO I read this thread on yammer, but I don't get what the agreement was.

I thought that @JesperSchulz agreed to add the copy functionality to the system application, but maybe that's a misunderstanding on my side.

I'm not quite sure if Jesper simple wanted you to remove the Access property from this table.

PThiemann-AINO:
@pri-kise The copy function used to be part of our app before moving it from C/AL to AL. First a lot was missing but since then you brought back a lot of functionality we were depending on. One last thing missing on our side is the copy function and we can re-integrate it as soon the access property has been removed. 

We might provide you with the functionality in future but currently this is still a decision to be made by us.

pri-kise:
> @pri-kise The copy function used to be part of our app before moving it from C/AL to AL. First a lot was missing but since then you brought back a lot of functionality we were depending on. One last thing missing on our side is the copy function and we can re-integrate it as soon the access property has been removed.
> 
> We might provide you with the functionality in future but currently this is still a decision to be made by us.

I'm not the one to decide if this access property is removed, since I'm not working for microsoft.

I only can tell, that in the past the Access properties weren't always removed, when someone created a PR here.
Therefore I tried to understand this yammer thread and guide you into the right direction.

JesperSchulz:
Sorry that it took me so long to get involved in the discussion of this pull request. Better late than never I guess, so here we go:

Unfortunately there seems to have been a small misunderstanding in the Viva Engage conversation. While I would applaud a pull request here on GitHub to expand on the API you need to fulfill your business needs, what is suggested in this pull request is not what I meant. The "Word Template Related Table" is an implementation detail and should as such not get exposed. Opening up the table for extensions to tamper with, could lead to unwanted side-effects. It would also rob us of the opportunity to change and improve the implementation in the way we see fit going forward, which is one of the main goals with our modules. Rather than requesting access to the internal implementation details, you should look at the [module's facade](https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/Word%20Templates/src/WordTemplate.Codeunit.al) and expand it with public procedure(s) you require.

As this repository is getting retired for System Application pull requests and the provided pull request in its current shape cannot be accepted, I am taking the liberty of closing this PR. If you decide to expand the Word Template's API, please create an issue on the new [BCApps ](https://github.com/microsoft/BCApps/issues) repository, reference this PR, and let's discuss the issue and possible solutions, before a new PR is created.

More details on the process for making contributions can be found here: https://github.com/microsoft/BCApps/blob/main/CONTRIBUTING.md

My apologies again for not being able to advise on this matter earlier. I hope we can find a solution which satisfies all parties on our new BCApps GitHub repository 😊

PR #25184, opened 2023-11-06
GP - Historical Snapshot Payment Apply records

jaymckinney-enavate:
This change enhances the GP Historical Snapshot to include Payment Apply records to both the Payables and Receivables historical transactions.

### DIFF
--- a/Apps/W1/DynamicsGPHistoricalData/app/src/Pages/HistPayablesApply.Page.al
+++ b/Apps/W1/DynamicsGPHistoricalData/app/src/Pages/HistPayablesApply.Page.al
@@ -1,0 +1,182 @@
+namespace Microsoft.DataMigration.GP.HistoricalData;
+
+page 41023 "Hist. Payables Apply"
+{
+    ApplicationArea = All;
+    UsageCategory = None;
+    Caption = 'Historical Payables Apply';
+    PageType = Card;
+    SourceTable = "Hist. Payables Apply";
+    DeleteAllowed = false;
+    InsertAllowed = false;
+    ModifyAllowed = false;
+    DataCaptionExpression = DataCaptionExpressionTxt;
+
+    layout
+    {
+        area(content)
+        {
+            group(General)
+            {
+                Caption = 'General';
+
+                field("Vendor No."; Rec."Vendor No.")
+                {
+                    ToolTip = 'Specifies the value of the Vendor No. field.';
+                }
+                field("Vendor Name"; Rec."Vendor Name")
+                {
+                    ToolTip = 'Specifies the value of the Vendor Name field.';
+                }
+                field("Document Type"; Rec."Document Type")
+                {
+                    ToolTip = 'Specifies the value of the Document Type field.';
+                }
+                field("Apply To Voucher No."; Rec."Apply To Voucher No.")
+                {
+                    ToolTip = 'Specifies the value of the Apply To Voucher No. field.';
+                }
+                field("Apply To Document No."; Rec."Apply To Document No.")
+                {
+                    ToolTip = 'Specifies the value of the Apply To Document No. field.';
+                }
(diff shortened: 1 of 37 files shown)


Apps/W1/DynamicsGPHistoricalData/app/src/Pages/HistPayablesApplyList.Page.al
```diff
@@ -0,0 +1,93 @@
...
+    layout
+    {
+        area(content)
+        {
+            repeater(General)
+            {
+                field("Apply To Document No."; Rec."Apply To Document No.")
+                {
+                    ToolTip = 'Specifies the value of the Apply To Document No. field.';
+                    Visible = not IsInvoice;
```
nikolakukrika, line 26: Application area is missing on the fields, see if caption is needed (if it can take the caption from table without the warning)

Apps/W1/DynamicsGPHistoricalData/app/src/Pages/HistPayablesApply.Page.al
```diff
@@ -0,0 +1,182 @@
...
+    {
+        area(content)
+        {
+            group(General)
+            {
+                Caption = 'General';
+
+                field("Vendor No."; Rec."Vendor No.")
+                {
+                    ToolTip = 'Specifies the value of the Vendor No. field.';
```
nikolakukrika, line 25: Applicaiton area is missing, check if the caption is missing on the fields

Apps/W1/DynamicsGPHistoricalData/app/src/Pages/HistPayablesApplyList.Page.al
nikolakukrika: Same comment on all pages

Apps/W1/DynamicsGPHistoricalData/app/src/Pages/HistReceivablesApplyList.Page.al
```diff
@@ -0,0 +1,87 @@
...
+                }
+                field("Write Off Amount"; Rec."Write Off Amount")
+                {
+                    ToolTip = 'Specifies the value of the Write Off Amount field.';
+                }
+            }
+        }
+    }
+
+    procedure FilterByDocumentNo(DocType: Enum "Hist. Receivables Doc. Type"; DocumentNo: Code[35])
```
nikolakukrika, line 70: Who is calling this method? If the method is called before the page is opened, then there is no need for CurrPage.Update at the end. I would update IsInvoice from the OnAfterGetRecord and OnAfterGetCurrentRecord triggers

nikolakukrika (approved):

Apps/W1/DynamicsGPHistoricalData/app/src/Pages/HistPayablesApplyList.Page.al
jaymckinney-enavate: As of BC22, the field level ApplicationArea property is inherited from the page if it isn't explicitly set on the field. I prefer this because it reduces redundancy.

https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-applicationarea-property

I don't have any warnings and the captions show up fine on the pages.

Apps/W1/DynamicsGPHistoricalData/app/src/Pages/HistReceivablesApplyList.Page.al
jaymckinney-enavate: Changes made, thanks

JesperSchulz:

Build ID to track: 14002.

JesperSchulz:

Build ID: 14000.

PR #25223, opened 2023-11-15
Refactor Service Declaration Mgt for temporary SalesHeader/PurchHeader

### LINKED ISSUE #25222: [Bug]: codeunit 5012 "Service Declaration Mgt." - Hard Get On Header leads to problems with Temporary Record
### Describe the issue

Currently there are a few Event Subscribers in the `codeunit 5012 "Service Declaration Mgt."`, that make a hard get on the SalesHeader, PurchHeader instead of using the GetHeader Functions.

This leads to problems with temporary SalesHeader for price calculations

### Expected behavior

Those event subscribers should use the GetHeader methods

### Steps to reproduce

Create a temporary Record for a SalesHeader an and then create a tempory sales line for a resource.

### Additional context

Temporary SalesHeader and Purch Header are often forgotten.

pri-kise:
Fixes #25222

### DIFF
--- a/Apps/W1/ServiceDeclaration/app/src/Processing/ServiceDeclarationMgt.Codeunit.al
+++ b/Apps/W1/ServiceDeclaration/app/src/Processing/ServiceDeclarationMgt.Codeunit.al
@@ -318,14 +318,16 @@ codeunit 5012 "Service Declaration Mgt."
     var
         SalesHeader: Record "Sales Header";
     begin
+        if SalesLine.IsTemporary() then
+            exit;
         SalesLine."Applicable For Serv. Decl." := false;
         if not IsFeatureEnabled() then
             exit;
         if Item.Type = Item.Type::Service then begin
-            SalesHeader := SalesLine.GetSalesHeader();
             SalesLine."Service Transaction Type Code" := Item."Service Transaction Type Code";
-            SalesLine."Applicable For Serv. Decl." :=
-              SalesHeader."Applicable For Serv. Decl." and (not Item."Exclude From Service Decl.");
+            if SalesHeader.Get(SalesLine."Document Type", SalesLine."Document No.") then
+                SalesLine."Applicable For Serv. Decl." :=
+                  SalesHeader."Applicable For Serv. Decl." and (not Item."Exclude From Service Decl.");
         end;
     end;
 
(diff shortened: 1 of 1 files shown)


pri-rapu:
LGTM - resolves errors for price calculation with temp. sales headers.

NKarolak (approved):

dannoe (approved):

BertDeTemmerman:
Can this be checked and resolved please?
This is a blocking issue for us, and the only valid solution we found is deinstalling the app.
So we would like to have this in the next CU.

JesperSchulz:
I will see what we can do! Processing PR!

Apps/W1/ServiceDeclaration/app/src/Processing/ServiceDeclarationMgt.Codeunit.al
```diff
@@ -322,7 +322,7 @@ codeunit 5012 "Service Declaration Mgt."
         if not IsFeatureEnabled() then
             exit;
         if Item.Type = Item.Type::Service then begin
```
JesperSchulz, line 324: Why are we not just exiting here, if the SalesHeader is temporary? Would that not do the trick? Calling GetSalesHeader comes at a cost (does more than getting and invokes events).

JesperSchulz (commented):
Just making sure we're doing the right thing. If you can confirm that "if IsTemporary() then exit" doesn't suffice, I'll rush this in - hopefully in time for 23.2 (today is last day for regular check-ins).

duichwer:
Someone reported an issue with Standard Sales Code.
When they are set to automatic, then the lines are inserted before the SalesHeader is inserted.

Take a look on the SalesHeader Insert Trigger.

Therefore the GetSakesHeader is needed at least there.

Apps/W1/ServiceDeclaration/app/src/Processing/ServiceDeclarationMgt.Codeunit.al
duichwer: See my comment below.

This PR does fix issues with Standard Sales Lines, too.

Additional we maybe can improve this in the future if we get a new method to simple calculate the Sales or Purch Prices for a Customer) Vendor.
Then we maybe won't need the TemSalesHeader anymore.

JesperSchulz:
@duichwer, how about we do both. We exit when IsTemporary, but we use the GetSalesHeader instead of get. Any objections?

JesperSchulz:
I have pushed the changes we are going to ship with 23.2. I hope those changes mostly do the trick. If not, could you please reopen the issue and create a new follow-up PR? I just really wanted to deliver something for 23.2, but the original PR got pushback, because it was calling events and functions not needed in this context. I'm sorry for overruling, but I thought it's better to deliver a solution to some of the problems, than none at all for 23.2. I hope that was ok.

JesperSchulz:

Build ID to track: 14141.

JesperSchulz:

Build ID: 14143.

AndreasMoth (approved):

PR #25990, opened 2024-02-21
[Bug]: Is not posible to create a G/L Account Mapping using Client

### LINKED ISSUE #25251: [Bug]: Is not posible to create a G/L Account Mapping using Client
### Describe the issue

Is not possible to create a new Standard account from the list.
Is not possible to create a new Standard account category using web client.

### Expected behavior

The Standard account can be created using web client.

### Steps to reproduce

1. Extend the enum "Standard Account Type" with any new enum.
2. Create a GL Account Mapping Using new Standard Account Type
3. Create a Standard Account Group. Here is a new problem. You can do this only with the Configuration package(
4. Fill out Standard Account Group No. in line and lookup in the field standard account no.
5. Try to create a new record on the standard accounts page.

### Additional context

image

vlkov:
#### Summary 
Fixes table relation property, because the first pk field is Type(Enum)
Added not blank property to primary key fields
Change control order on page like as primary key order
#### Work Item(s) 
Fixes #25251

Fixes [AB#502390](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/502390)

### DIFF
--- a/Apps/W1/AuditFileExport/app/src/GLAccountMapping/StandardAccount.Table.al
+++ b/Apps/W1/AuditFileExport/app/src/GLAccountMapping/StandardAccount.Table.al
@@ -20,11 +20,12 @@ table 5263 "Standard Account"
         {
             DataClassification = CustomerContent;
             Caption = 'Category No.';
-            TableRelation = "Standard Account Category" where("Standard Account Type" = field(Type));
+            TableRelation = "Standard Account Category"."No." where("Standard Account Type" = field(Type));
         }
         field(3; "No."; Code[20])
         {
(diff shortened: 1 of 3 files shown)


JesperSchulz:
We'll look into it! Thanks for creating the PR 😊

JesperSchulz (approved):

JesperSchulz:
@vlkov, unfortunately this didn't make it in time for 2024 wave 1. If needed, we could backport it for release with a minor update, even though the changes are somewhat semantically breaking. What do you think?

PR #26133, opened 2024-03-19
Prevent duplicate record in the Master Data Management - Synchronization Tables list

### LINKED ISSUE #26105: [Bug]: Duplicate record in the Master Data Management - Synchronization Tables list
### Describe the issue

Duplicate record is inserted in the Synchronization Tables list when the selected table has a self-reference or the recursive search returns to the initial table.
image

### Expected behavior

Duplicate record entries should not be allowed in the synchronization list.

### Steps to reproduce

1. Open the Master Data Management - Synchronization Tables
2. Add a new entry and select the Dimension table
3. The dialog appears requesting the confirmation to add related tables:
_The chosen table has a relation to the following tables that are currently not included in the synchronization: IC Dimension, Dimension.
Do you want to synchronize these tables too?_
4. Press *Yes* to confirm
5. Dimension table is added to the list twice (see the screenshot above).

### Additional context

_No response_

adrogin:
Resolves #26105 

The loop adding related tables to the synchronization list is triggered before the initial table requested by the user is added, therefore it will be added to the list of related tables if any of the relations points back to the table itself. An additional check is added in the loop precondition to handle this case.

Fixes [AB#534785](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/534785)


Apps/W1/MasterDataManagement/app/src/pages/MasterDataSynchTables.Page.al
dcenic, line 1: Don't add an extra parameter exclude the top level table in this method. The table relates to itself and it is OK that this method adds it to the list. Instead - when adding tables in Synchronization Tables, just don't add it twice

Apps/W1/MasterDataManagement/app/app.json
dcenic, line 1: Don't expose internals to another app. Instead, make a method in the test library and call it from the test.

Apps/W1/MasterDataManagement/app/src/pages/MasterDataSynchTables.Page.al
dcenic: In the AssistEdit, when you are taking tables out of the dictionary that this method found, and about to call AddTable, add a check if (RelatedTableNo <> AllObjWithCaption."Object Id"). That will make this change a one-liner. No need to change signatures of methods.

Apps/W1/MasterDataManagement/test library/src/MDMTestTableB.Table.al
dcenic, line 1: Instead of introducing two new test tables, you can just have one test table that references a well-known Base App table that will always be there in decades to come (e.g. Customer), and also references itself.

Apps/W1/MasterDataManagement/app/app.json
dcenic: If you make the test go through the TestPage, like test method SynchronizeAdditionalTable, then you don't need to add anything to the test library either. You already have everything that is needed to write one more test.

dcenic (changes requested):

Apps/W1/MasterDataManagement/app/app.json
adrogin: > Don't expose internals to another app. Instead, make a method in the test library and call it from the test.

Majority of the apps in this repository expose internals to their tests, with those limiting the visibility to libraries being rather an exception. Is this the practice we should follow then going forward?

>If you make the test go through the TestPage, like test method SynchronizeAdditionalTable, then you don't need to add anything to the test library either. You already have everything that is needed to write one more test.

And if I run the method directly or via the library I don't have to involve UI to test internal application logic. :-)
TestPages are slow and binding the test to UI controls it not necessary here. I'd rather have this method in a codeunit, outside of the page, to be honest, to make it more testable.

Apps/W1/MasterDataManagement/app/src/pages/MasterDataSynchTables.Page.al
adrogin: > In the AssistEdit, when you are taking tables out of the dictionary that this method found, and about to call AddTable, add a check if (RelatedTableNo <> AllObjWithCaption."Object Id") (add it even before you get its table metadata - you can skip it right away). That will make this change a one-liner. No need to change signatures of methods.

This is exactly what I would like to avoid - binding the test to UI controls and having the caller care about preparing the data or postprocessing the received list. Ideally, the graph traversing should return the initial node as part of the list. The issue here is that we explicitly separate this initial table from rest because of the confirmation request which applies only to the related tables (not including the top-level one itself).

Problem number two with this suggestion is that `FindRelatedTables` returns the the list of table names as a text variable to show to the user in the confirmation request ("There are related tables X, Y, Z. Do you want to include them as well?"). Something along these lines. This list "X, Y, Z" is also part of the recursive traverse and must be consistent with the dictionary.
I know, Ideally it must be separated and table names must be retrieved after the dictionary is collected. But this means reading table metadata once again after exiting the recursive search. The author of the code had to choose the lesser evil and picked this solution.

Apps/W1/MasterDataManagement/test library/src/MDMTestTableB.Table.al
adrogin: > Instead of introducing two new test tables, you can just have one test table that references a well-known Base App table that will always be there in decades to come (e.g. Customer), and also references itself.

Why is it a better solution? I don't see much difference between adding one test table or two test tables, frankly. Referencing a well-known base app table means introducing a dependency which is not relevant for the test, while it can be quite self-sufficient. So I followed a simple logic - the less the dependency on something not directly related to the test, the better.

Apps/W1/MasterDataManagement/app/app.json
dcenic: Don't expose internals to another app. Write the test so that it uses test page.

Apps/W1/MasterDataManagement/app/src/pages/MasterDataSynchTables.Page.al
dcenic: I agree. Even in the hypothetical situation where the table has only one relation: to itself - we want the RelatedTablesToAdd to be empty, so that we don't ask any question.

Apps/W1/MasterDataManagement/test library/src/MDMTestTableB.Table.al
dcenic: This way you waste one table id. On the other hand, there will be no Business Central without Customer table, so you can count on it being there. It's not demo data, it's something that will not go away. But OK, let's take one extra table id just for this.

dcenic (changes requested):
Don't expose internals to one more app

Apps/W1/MasterDataManagement/app/app.json
JesperSchulz: @adrogin / @dcenic, sounds like you're in disagreement around the way to test this change.
@adrogin, do you see a path forward, or does this require further discussions?

adrogin: @JesperSchulz @dcenic  Not so much of a disagreement, really. Not a problem to add a procedure to the test library of course. I just wanted to understand the criteria, when an app exposes internals to a test and which cases require a library, if there is a consistent requirement.

dcenic: The reason is security. To fix this bug, you don't need to expose all the internal methods to yet another app, no matter who the publisher of that app is. Test library app can see the internals and that's enough to make the test.

dcenic (approved):

JesperSchulz:
And we're good to go! I will process this one soon 🥳

JesperSchulz:
@adrogin, I did merge your changes internally, so they are ready to be released!

But could you maybe pull from origin main to also make this PR pass? Then we can merge it, rather than close it 😊 Build fails in current state of the PR.

JesperSchulz (approved):

adrogin:
@JesperSchulz Sure, done.

PR #26216, opened 2024-03-29
[Shopify] Add Translations Export to Products and Variants

tinestaric:

here‘s a quick summary:

Shopify Shop has a new action that takes the user to the Languages page. Languages cannot be manually inserted or deleted, they can only be refreshed with an action that pulls them from Shopify.
Before a language can be enabled for syncing, a language code needs to be mapped to it, as Shopify is using locales, and I can‘t directly map them. 

A new Scope is required to get shop locales – read_locales

On export of products and variants, an update of translations is triggered. The translation update first collects all TranslatableContentDigests for a resource, as these are needed when requesting an update of a translation.
The update next collects translations for the product Title, Product Body, and Variant „option 1“ in a temporary table for each enabled language. It compares the translations with the translations that were last synced. If the translation is the same, it removes it from the buffer. Next, the buffer is used to build the query and sent to Shopify to update translations.

Continuing with a couple of questions/considerations regarding the issue

•	The way I designed Shopify Languages is that the user can‘t insert/delete them, but has to pull them from Shopify with an action. This uses ShopLocales query which needs a read_locales scope that‘s not currently part of the scopes. I wasn‘t sure if that‘s an issue or not.
•	I made the import of locales skip the primary language as I wanted to keep the logic for that one on the Shop.“Langauge Code“ field.
•	I‘ve refactored the CreateProductBody procedure in the Product Export codeunit so it takes the language code as a parameter. This means I can reuse it for translations, however, there are events in this procedure at the start/end, which will now trigger more often if more languages are synced. I added the language code as a parameter in the publishers, and technically nothing is broken until additional languages are synced, but I wanted to mention this. 
•	I wasn‘t sure how I should approach translations when they‘re missing. Should I take the existing approach, take the translated text or default text, or should I only take the translated text, and if it doesn‘t exist, skip that property? 
This is mainly an issue with the product body. If Ext. Text is translated, but Attributes are not, should I skip on attributes or send attributes in the default language?
Currently, I skip the Title if it‘s not translated, but I send the body in every case. Let me know if that should be changed.
•	For the update of a translation, I need to send the TranslatableContentDigest as part of the request. Currently, I‘ve done this in a way, that I collect all digests for a resource during an update, but that means that every update of a product/variant now has an additional „get“ call, before it maybe sends an update for translations. I think this could be done better, and would appreciate your input here.
•	Resource Type is almost the same enum as used in Metafields, I think these two should be made the same.
•	I‘m not a fan of what I did with Translations Mgt. Is there a better place where I could put this procedure for retrieving the item translation?

Fixes #26819
Fixes [AB#460580](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/460580)


Apps/W1/Shopify/app/src/Base/Tables/ShpfyLanguage.Table.al
```diff
@@ -0,0 +1,72 @@
...
+    {
+        field(1; "Shop Code"; Code[20])
+        {
+            Caption = 'Shop Code';
+            DataClassification = CustomerContent;
+            Editable = false;
+            TableRelation = "Shpfy Shop";
+        }
+
+        field(2; Locale; text[2])
```
pri-kise, line 20: I don't know how it's in the cloud, but Text fields in primary key are evaluated as Code in primary keys depending on the SQL setting.
Is it important that the values are a Text and not a Code?

There has been a thread on this topic on yammer some months ago, but the search isn't working today and I cannot find this thread right now.

Apps/W1/Shopify/app/src/Base/Tables/ShpfyLanguage.Table.al
```diff
@@ -0,0 +1,72 @@
...
+            end;
+        }
+        field(4; "Language Code"; Code[10])
+        {
+            Caption = 'Language Code';
+            TableRelation = Language;
+            DataClassification = CustomerContent;
+
+            trigger OnValidate()
+            begin
```
pri-kise, line 44: @tinestaric should we verify that the `"Language Code"` assignment is unique?

Apps/W1/Shopify/app/src/Translations/Tables/ShpfyTranslation.Table.al
```diff
@@ -0,0 +1,127 @@
...
+        }
+        field(3; "Locale"; text[2])
+        {
+            Caption = 'Locale';
+            DataClassification = SystemMetadata;
+        }
+
+        field(4; Name; Text[100])
+        {
+            Caption = 'Sync translations';
```
pri-kise, line 32: This captions seems to be wrong.

Apps/W1/Shopify/app/src/Base/Tables/ShpfyLanguage.Table.al
tinestaric: Huh, I did not know about that behavior...
In this case, I need the locale to be in lowercase, as it's case-sensitive in Shopify, and it rejects a translation update otherwise. 
Technically, I can work around it, and transform it to lowercase before I send the update, but on a local Sandbox container, the Text key was maintained as text.

Apps/W1/Shopify/app/src/Translations/Tables/ShpfyTranslation.Table.al
tinestaric: Good catch.

Apps/W1/Shopify/app/src/Base/Tables/ShpfyLanguage.Table.al
tinestaric: Hmm, valid question. PO's reviewing this item this week, I'll let them decide.

Apps/W1/Shopify/app/src/Base/Tables/ShpfyLanguage.Table.al
pri-kise: Here some links on this topic:
https://github.com/StefanMaron/BusinessCentral.LinterCop/discussions/459
https://www.yammer.com/dynamicsnavdev/threads/2583811689037824

Apps/W1/Shopify/app/src/Base/Enums/ShpfyResourceType.Enum.al
```diff
@@ -0,0 +1,20 @@
+namespace Microsoft.Integration.Shopify;
```
onbuyuka, line 1: I think this file needs to be in Translations folder

Apps/W1/Shopify/app/src/Base/Pages/ShpfyLanguages.Page.al
```diff
@@ -0,0 +1,58 @@
+namespace Microsoft.Integration.Shopify;
```
onbuyuka, line 1: I think this file needs to be in Translations folder

Apps/W1/Shopify/app/src/Base/Tables/ShpfyLanguage.Table.al
```diff
@@ -0,0 +1,72 @@
+namespace Microsoft.Integration.Shopify;
```
onbuyuka, line 1: I think this file needs to be in Translations folder

onbuyuka (changes requested):

Apps/W1/Shopify/app/src/Base/Tables/ShpfyLanguage.Table.al
onbuyuka: @tinestaric please take a look

tinestaric:
@onbuyuka Moved the files under the Translations folder.
I'd argue that Resource Type is better placed in Base, as Metafields will requires the same enum, and potentially more "generic" objects as well.

onbuyuka (dismissed):

JesperSchulz (approved):

onbuyuka (approved):

AndreiPanko:
@tinestaric Hi, you wrote "on export".

Does it also include "Add item to Shopify"? As I can not see that it is triggered from 

report 30106 "Shpfy Add Item to Shopify"
Codeunit "Shpfy Create Product",
which call ProductExport.FillInProductFields(Item, TempShopifyProduct); but it doesn't call translation logic.

Am I missing something?

tinestaric:
@AndreiPanko
I must have missed the "Add item to Shopify". As you've noticed, it will only push additional translations on sync items.

Should I make another PR to add the calls to translation logic in ProductAPI.CreateProduct?

AndreiPanko:
@Tinestaric. Thank you Tine. We fixed it internally. No action needed.

PR #26707, opened 2024-06-20
[Shopify] Integration to use Sales Unit of Measure for inventory synchronization

### LINKED ISSUE #26719: [Bug]: [Shopify] Integration to use Sales Unit of Measure for inventory synchronization
### Describe the issue

We saw that the Sales Unit of Measure should be used for inventory synchronization but there is a bug that prevents it from doing it. In line 61, Item.???Sales Unit Of Measure??? will always be empty, because in the function in line 51, it is cleared. It means that if the statement in line 63 never is true, the Base Unit of Measure will always be used for inventory synchronization.

### Expected behavior

N/A

### Steps to reproduce

N/A

### Additional context

N/A

aidasberesinevicius:

Quick summary:

We saw that the Sales Unit of Measure should be used for inventory synchronization but there is a bug that prevents it from doing it. In line 61, Item.???Sales Unit Of Measure??? will always be empty, because in the function in line 51, it is cleared. It means that if the statement in line 63 never is true, the Base Unit of Measure will always be used for inventory synchronization.

Fixes #26719
Fixes [AB#539103](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/539103)

### DIFF
--- a/Apps/W1/Shopify/app/src/Inventory/Codeunits/ShpfyInventoryAPI.Codeunit.al
+++ b/Apps/W1/Shopify/app/src/Inventory/Codeunits/ShpfyInventoryAPI.Codeunit.al
@@ -33,6 +33,7 @@ codeunit 30195 "Shpfy Inventory API"
         ShopifyVariant: Record "Shpfy Variant";
         StockCalculation: Interface "Shpfy Stock Calculation";
         UOM: Code[10];
+        SalesUOM: Code[10];
     begin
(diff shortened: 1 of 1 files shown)


JesperSchulz (approved):

JesperSchulz:

Build ID: <None>.

PR #26712, opened 2024-06-20
[Shopify] Add items as product variants

tinestaric:

Added an action to Variants subpage to add BC Items as additional Variants to an existing Shopify Product.
If the parent product only had a default variant before the addition, it is deleted from both Shopify and BC.
The variants are added under the existing the Product Option (ex. Color, Material, or Title, if product only had the default variant)

Items cannot be added if the shop has "UOM as Variant" enabled, or if the parent product has more than one option defined.

Fixes #26819
Fixes [AB#468218](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/468218)


Apps/W1/Shopify/app/src/Products/Codeunits/ShpfyCreateProduct.Codeunit.al
```diff
@@ -87,22 +87,7 @@ codeunit 30174 "Shpfy Create Product"
                             ProductPriceCalc.CalcPrice(Item, ItemVariant.Code, ItemUnitofMeasure.Code, TempShopifyVariant."Unit Cost", TempShopifyVariant.Price, TempShopifyVariant."Compare at Price");
                             TempShopifyVariant.Title := ItemVariant.Description;
                             TempShopifyVariant."Inventory Policy" := Shop."Default Inventory Policy";
-                            case Shop."SKU Mapping" of
```
tinestaric, line 90: More could be refactored in this codeunit, but I'll only change things that I touch/need. And as I wanted to extract the create variant from item, I extracted this SKU mapping, as it's the same for every procedure.

Apps/W1/Shopify/app/src/Products/Codeunits/ShpfyCreateProduct.Codeunit.al
```diff
@@ -157,6 +127,7 @@ codeunit 30174 "Shpfy Create Product"
                 ItemUnitofMeasure.SetRange("Item No.", Item."No.");
                 if ItemUnitofMeasure.FindSet(false) then
                     repeat
+                        TempShopifyProduct."Has Variants" := true;
```
tinestaric, line 130: This was a Bug. Product was created on Shopify with variants (due to UOM), yet in BC it was marked as if it doesn't have any.

Apps/W1/Shopify/app/src/Products/Codeunits/ShpfyCreateProduct.Codeunit.al
```diff
@@ -256,6 +184,48 @@ codeunit 30174 "Shpfy Create Product"
...
+    /// </summary>
+    /// <param name="Item">The item to create the variant from.</param>
+    /// <param name="TempShopifyVariant">The temporary Shopify variant record set where the variant will be inserted.</param>
+    internal procedure CreateTempShopifyVariantFromItem(Item: Record Item; var TempShopifyVariant: Record "Shpfy Variant" temporary)
+    begin
+        Clear(TempShopifyVariant);
+        TempShopifyVariant."Available For Sales" := true;
+        TempShopifyVariant.Barcode := CopyStr(GetBarcode(Item."No.", '', Item."Sales Unit of Measure"), 1, MaxStrLen(TempShopifyVariant.Barcode));
+        ProductPriceCalc.CalcPrice(Item, '', Item."Sales Unit of Measure", TempShopifyVariant."Unit Cost", TempShopifyVariant.Price, TempShopifyVariant."Compare at Price");
+        TempShopifyVariant.Title := ''; // Title will be assigned to "Default Title" in Shopify as no Options are set.
```
tinestaric, line 218: Previously, this was set to ItemVariant.Description, but ItemVariant was never retrieved/Initialized in this code path, so it was effectivelly always empty. For clarity, this could also be "Default Title", as that's what the title is going to be in Shopify.

Apps/W1/Shopify/app/src/Products/Codeunits/ShpfyVariantAPI.Codeunit.al
```diff
@@ -140,8 +141,12 @@ codeunit 30189 "Shpfy Variant API"
         JResponse := CommunicationMgt.ExecuteGraphQL(GraphQuery.ToText());
         NewShopifyVariant := ShopifyVariant;
         NewShopifyVariant.Id := JsonHelper.GetValueAsBigInteger(JResponse, 'data.productVariantCreate.productVariant.legacyResourceId');
-        if NewShopifyVariant.Id > 0 then
+        if NewShopifyVariant.Id > 0 then begin
             NewShopifyVariant.Insert();
+            exit(true);
```
tinestaric, line 146: Added a return value, as I want to mark the parent product with "Has Variants" only if a variant was successfully created.

Apps/W1/Shopify/app/src/Products/Codeunits/ShpfyCreateItemAsVariant.Codeunit.al
```diff
@@ -0,0 +1,109 @@
+namespace Microsoft.Integration.Shopify;
+
+using Microsoft.Inventory.Item;
+
+codeunit 30213 "Shpfy Create Item As Variant"
+{
+    TableNo = Item;
+
```
pri-kise, line 8: I think this codeunit should have Access= Internal.

Apps/W1/Shopify/app/src/Products/Reports/ShpfyAddItemAsVariant.Report.al
```diff
@@ -0,0 +1,116 @@
+namespace Microsoft.Integration.Shopify;
+
+using Microsoft.Inventory.Item;
+
+/// <summary>
+/// A report to add items as variants to a parent product.
+/// </summary>
+report 30117 "Shpfy Add Item As Variant"
```
pri-kise, line 8: I know that the other reports don't have any AccessByPermission property set, but I think it would be useful to hide this report in the UI  from Users with missing permissions.

I would try to set the AccessByPermission to e.g. `tabledata "Shpfy Variant" = RIMD` (this was the table that I found in the code)

tinestaric: Good idea. I'll run it by Andrei.

Apps/W1/Shopify/app/src/Products/Codeunits/ShpfyCreateProduct.Codeunit.al
```diff
@@ -247,7 +175,7 @@ codeunit 30174 "Shpfy Create Product"
     /// <param name="VariantCode">Parameter of type Code[10].</param>
     /// <param name="UoM">Parameter of type Code[10].</param>
     /// <returns>Return value of type Text.</returns>
```
pri-kise, line 177: XML Summary is outdated after the change of the return type

tinestaric: Updated.

Apps/W1/Shopify/app/src/Products/Codeunits/ShpfyCreateItemAsVariant.Codeunit.al
```diff
@@ -0,0 +1,110 @@
+namespace Microsoft.Integration.Shopify;
+
+using Microsoft.Inventory.Item;
+
+codeunit 30343 "Shpfy Create Item As Variant"
```
onbuyuka, line 5: I think this could use a little tests. We are already well covered in codeunit 139601 "Shpfy Create Product Test". That would be a good place to start.

onbuyuka: @tinestaric fyi

onbuyuka (dismissed):

onbuyuka (approved):

AndreiPanko (approved):

PR #26728, opened 2024-06-21
[Shopify] Shopify return location on refunds

tinestaric:

Added a setting to the shopify shop to use original return location (From shopify) or a default return location set in the Shopify shop when creating Credit Memos for Refunds.

Modified the return and refund creation to pull original location from shopify refund or return orders. 
When an item on a single return order in shopify is restocked to multiple locations, we cannot determine a single location to put on a return line. 

If there is any reason that Original location couldn't be determined then the Default Return Location will be used during the creation of a Credit Memo.

Fixes #26819

Fixes [AB#540965](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/540965)


JesperSchulz:
@tinestaric, I'll try to process as many of the Shopify PRs as I can this week. Any chance you could resolve the merge issue before I do?

JesperSchulz:
Also, it seems like ShpfyGraphQLType.Enum.al is modified in a lot of PRs. Maybe it would be an idea to already take care of the IDs to prevent overlap between all PRs?

tinestaric:
@JesperSchulz Fixed the merge conflict.

As for the enum, we'll have the same issue with object IDs. We weren't sure how these PRs were going to be processed. Would they go through the normal contribution workflow, or was someone from MS team taking them over.

In any case, I have no problem aligning the enum and object ids. Is there an order of PRs you'll follow? So I can sequence the ids in the same way.

JesperSchulz:
> @JesperSchulz Fixed the merge conflict.
> 
> As for the enum, we'll have the same issue with object IDs. We weren't sure how these PRs were going to be processed. Would they go through the normal contribution workflow, or was someone from MS team taking them over.
> 
> In any case, I have no problem aligning the enum and object ids. Is there an order of PRs you'll follow? So I can sequence the ids in the same way.

Let's just say I'm going bottom up 😋 Oldest first.
I too wasn't sure how these will get processed, but I'm tired of having them open, so now I decided to take action 😎

tinestaric:
@JesperSchulz I'll go through them. Quick question, what's the [OnHold] tag for?

tinestaric:
@JesperSchulz 
I went through all the Shopify PRs and fixed the CI/CD errors. I've also tried to renumber object and enum IDs so that they wouldn't clash if only these PRs would be merged oldest to newest. 

GraphQL Enum will most likely still throw a merge conflict, just because every PR is adding values at the end after value 78.

I didn't fix object IDs for the newest PR #26748, primarily because I ran out of IDs for Codeunits, so the object range will have to be addressed, before that one can be merged in.

onbuyuka (approved):

PR #26730, opened 2024-06-22
[Fix] URL maximum length in E-Document Integration

### LINKED ISSUE #26672: [Bug]: URL maximum length in E-Document Integration
### Describe the issue

When implementing the interface Microsoft.eServices.EDocument."E-Document Integration" to Send E-Documents we came across the issue that request URIs which are longer than 250 characters lead to an error in procedure InsertIntegrationLog of Codeunit 6132 "E-Document Log" due to the validation of the URL record field 7 of table 6127 "E-Document Integration Log" which is limited to 250 characters.

### Expected behavior

The supported maximum length of the URL should be longer.

### Steps to reproduce

1. Create a Codeunit which implements "E-Document Integration"
2. Implement procedure "Send" using a request URI with more than 250 characters in parameter HttpRequest
3. Error will be raised by the field validation in the subsequent call to the procedure InsertLog in codeunit 6132 "E-Document Log" due to the length of the RequestUri which is longer than 250 characters.

image

### Additional context

The .NET System.Uri class limits the maximum length of a URI for the System.Net.Http.HttpClient to [65519 characters](https://learn.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=net-8.0#system-uri-ctor(system-string)). Per [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#section-4.1-5) (published in June 2022) there is no fixed limit for URIs, but it is recommended to support, "at a minimum, URIs with lengths of 8000 octets in protocol elements". Given the Business Central context, we would argue for URIs limited to not more than 2048 characters as an expected limit.

As a workaround we currently update the request URI of the HttpRequest parameter before exiting the procedure as follows:

image

pluethi:
#### Summary 
Added a new field "Request URL" to table 6127 "E-Document Integration Log" with a maximum string length of 2048 to allow for longer URLs when using the "E-Document Integration" interface. The new filed replaces the existing field URL. I marked the existing field as obsolete and updated three (all) references from the old URL field to the new "Request URL" field.

#### Work Item(s) 
Fixes #26672

Fixes [AB#540448](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/540448)

### DIFF
--- a/Apps/W1/EDocument/app/src/Log/EDocumentIntegrationLogs.Page.al
+++ b/Apps/W1/EDocument/app/src/Log/EDocumentIntegrationLogs.Page.al
@@ -31,7 +31,7 @@ page 6128 "E-Document Integration Logs"
                     {
                         ToolTip = 'Specifies the service code for the document.';
                     }
-                    field(URL; Rec.URL)
+                    field(URL; Rec."Request URL")
                     {
                         ToolTip = 'Specifies the integration url used to send the document.';
                     }
--- a/Apps/W1/EDocument/app/src/Log/EDocumentIntegrationLog.Table.al
+++ b/Apps/W1/EDocument/app/src/Log/EDocumentIntegrationLog.Table.al
@@ -41,10 +41,22 @@ table 6127 "E-Document Integration Log"
         field(7; URL; Text[250])
         {
             Caption = 'URL';
+            ObsoleteReason = 'Replaced with Request URL field';
+#if not CLEAN25
+            ObsoleteState = Pending;
+            ObsoleteTag = '25.0';
+#else
+            ObsoleteState = Removed;
+            ObsoleteTag = '28.0';
+#endif
         }
         field(8; Method; Text[10])
         {
             Caption = 'Method';
+        }
+        field(9; "Request URL"; Text[2048])
+        {
+            Caption = 'URL';
         }
     }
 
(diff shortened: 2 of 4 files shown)


pluethi:

Apps/W1/EDocument/app/src/Log/EDocumentIntegrationLogs.Page.al
```diff
@@ -31,7 +31,7 @@ page 6128 "E-Document Integration Logs"
                     {
                         ToolTip = 'Specifies the service code for the document.';
                     }
-                    field(URL; Rec.URL)
+                    field(URL; Rec."Request URL")
```
JesperSchulz, line 34: Given we now reference the new field, shouldn't we have upgrade code then? Using DataTransfer that should be a swift procedure.

pluethi: That's true, of course. I had not considered this, as it is "only" a log. However, it is visible to the user, so as a user I would also expect the existing data to be transferred. I will add this.

JesperSchulz:
Upgrade code added internally. Will merge once it has passed the build system.

JesperSchulz (approved):
I've added upgrade code during merge to local repo.

Apps/W1/EDocument/app/src/Log/EDocumentIntegrationLogs.Page.al
JesperSchulz: I've added the upgrade code internally. Case closed.

JesperSchulz:

Build ID: 22963.

PR #26799, opened 2024-07-05
Move BCPT Sample Tests.

JesperSchulz:
Move the BCPT sample tests to https://github.com/microsoft/BusinessCentralApps/tree/main/App/Layers/W1/Tests/BCPT-SampleTests.


BardurKnudsen (approved):
Fedt!

vjekob:
@JesperSchulz is this why this error is now received in BCPT extension in VS Code:
image

(happens when you try to set up a new BCPT project and choose GitHub Latest sample tests from GitHub option)

thomasguldstrand:
When I try to setup a new BCPT project in VS Code I get the error. Could this be because I do not have access to the private repo?

image

image

JesperSchulz:
@ChethanT, could you look into this?

ChethanT:
BCPT vsix points to the old repo to download the sample tests. We need to fix the vsix to point to the new location. I will open a workitem for us to fix that.

PR #27350, opened 2024-09-30
[Intrastat Core] Changes to "Validate Codeunit ID" and "Content Codeunit ID" behavior on "Intrastat Report" page

### LINKED ISSUE #27114: [Bug]: UpdateErrors() on Intrastat Report page is not called when executing Validation Codeunit ID set in VAT Reports Configuration
### Describe the issue

We are using custom CU for exporting Intrastat Report. To be able to make file, we also need to use custom CU for validation because of this condition in CreateFile action:
image

For our validation we want to use standard code, therefore our validation CU calls the same procedure as standard:
image

When we execute ChecklistReport action, check is performed normally but UpdateErrors() is not called:
image

If any error is logged, line is colored red but Error Messages part is empty:
image

Refreshing with F5 does not help. Only after you close and reopen the Intrastat Report page, Error Messages part is populated correctly:
image

### Expected behavior

In ChecklistReport action, UpdateErrors() should be called after validation codeunit is executed. Calling this function from validation CU is impossible, because, a) the function is local, and b) validation CU has no access to CurrentPage object.

Furthermore, CreateFile action also needs to be refactured, because it doesn't call UpdateErrors() after calling validation CU.

Actually, I would reconsider the choice to only allow usage of Content CU and Validation CU in tandem. As our case shows, sometimes standard validation is enough and you only need custom export logic. So further refactoring of CreateFile action might be in order, separating these two steps and enabling us to choose, where to use standard and where to go custom.

### Steps to reproduce

Create Validation CU with copy of standard code as shown above and Content CU (can be dummy).
Set both codeunits to be used in VAT Reports Configuration.
Configure Intrastat Report Checklist. Example:
image
Prepare Intrastat Report Lines with errors (value missing in fields that are set up in Intrastat Report Checklist). Example:
image
Execute Checklist Report action. Line is colored red but Error Messages is empty:
image

### Additional context

I found another open issue regarding UpdateErrors() function. That issue requests that the procedure is made protected, but that only enables a workaround (replacing standard actions with custom ones) and does not fix the bug.

matjaz-sega:
#### Summary

There is a problem with current implementation of "Validate Codeunit ID" and "Content Codeunit ID" in several actions on "Intrastat Report" page. I have rewritten the code so that Validate and Content codeunits can be used independently of each other. Whenever validation is performed, if "Validate Codeunit ID" is not set in "VAT Reports Configuration", standard validation code is executed. Whenever file creation is performed, if "Content Codeunit ID" is not set in "VAT Reports Configuration", standard export code is executed.

#### Work Item(s) 
fixes #27114 

Fixes [AB#554367](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/554367)

### DIFF
--- a/Apps/W1/Intrastat/app/src/IntrastatReport.Page.al
+++ b/Apps/W1/Intrastat/app/src/IntrastatReport.Page.al
@@ -138,11 +138,9 @@ page 4812 "Intrastat Report"
                     then begin
                         Commit();
                         Codeunit.Run(VATReportsConfiguration."Validate Codeunit ID", Rec);
-                        CurrPage.Update();
-                        exit;
-                    end;
+                    end else
+                        IntrastatReportMgt.ValidateReportWithAdvancedChecklist(Rec, false);
 
-                    IntrastatReportMgt.ValidateReportWithAdvancedChecklist(Rec, false);
                     UpdateErrors();
                     CurrPage.Update();
                 end;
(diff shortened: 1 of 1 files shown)


matjaz-sega:
@microsoft-github-policy-service agree company="BE-terna"

RadoArvay (approved):
Great! As we discussed.

JesperSchulz:
The PR was merged internally. In order to merge here on GitHub, we need a succeeding build. Not sure why it failed... will re-run.

JesperSchulz (approved):

JesperSchulz:

Build ID: 25897.

PR #27402, opened 2024-10-11
Allow debugging all first party apps

### LINKED ISSUE #27385: [BC Idea]: Allow debugging any first level app
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=4b814ed7-0986-ef11-9443-6045bdb37955

### Description

Some Apps from Microsoft set resource exposure policies to not allow debugging. from a developer experience perspective, this is not good and makes troubleshooting and traceability very difficult.

apps that do not allow debugging: https://github.com/search?q=repo%3Amicrosoft%2FALAppExtensions%20%22allowDebugging%22%3A%20false&type=code

- [X] I will provide the implementation for this BC Idea

ChrisBlankDe:
#### Summary 
Set all apps allowDebugging to true
#### Work Item(s) 
Fixes #27385

### DIFF
--- a/Apps/AT/HybridBCLast_AT/app/app.json
+++ b/Apps/AT/HybridBCLast_AT/app/app.json
@@ -28,7 +28,7 @@
   "screenshots": [],
   "platform": "26.0.0.0",
   "resourceExposurePolicy": {
-    "allowDebugging": false,
+    "allowDebugging": true,
     "allowDownloadingSource": true,
     "includeSourceInSymbolFile": true
   },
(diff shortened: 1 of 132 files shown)


JesperSchulz:
@ChrisBlankDe, we've had some internal discussions about what to do with this PR, as every single app needs to go through security review for this change to go in - which will set us back days.

Given we anyway have an ongoing project of migrating the apps from this repo to BCApps, where the apps must be "clean", this work could get bundled with that effort. We will remove any .NET from the apps, make target "cloud" and make sure they are debuggable.

That is going to take a little while (probably first half of 2025), but we'd do things the right way.

Would that work for you?

ChrisBlankDe:
That sounds very good.
It's better to do everything calmly and in a structured way than to rush into something.

\>first half of 2025< is not that far away now :)

PR #27539, opened 2024-10-25
[Shopify] Log Skipped Records on export processes

tinestaric:
This pull request does not have a related issue as it's part of delivery for development agreed directly with @AndreiPanko 
Fixes #26819 

### Created new structure to log records skipped on export

Solution has new structure possible to open by typing "Shopify Skipped Records". On page user can see all the skipped records with reasons and date and time when the record was skipped. User can also open the related record using "Show record" action or clicking the "Description" field value.

### Log cases:

Customer:
- Customer has empty email
- Customer with same email or phone exists

Posted Sales Invoice:
- Customer does not exist in Shopify
- Payment terms do not exist in Shopify
- Customer No. is Default Customer No. for Shopify Shop
- Customer No. is used in Shopify Customer Template
- No lines existing in sales invoice
- Invalid Quantity
- Empty No. value

Product:
-  Item is blocked/sales blocked (Item Variant)
-  Item is blocked

Shipments:
 - Related Shopify Order does not exist
 - No lines in Posted Sales Shipment applicable for fulfillment 
 - No corresponding fulfillment found in Shopify.

If the condition met log record is being created.

### Retention policy
Retention policy is inserted on extension install. It can be turned on "Retention Policies" page.

Fixes [AB#473306](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/473306)

PR #27544, opened 2024-10-25
[Shopify] Shopify return location on refunds tests

petemchlk:
### Shopify return location on refunds tests

Automated tests for PR https://github.com/microsoft/ALAppExtensions/pull/27539


petemchlk:
Created by mistake

PR #27772, opened 2024-12-05
GP - Remove the CLEAN25 preprocessor checks from the US app

jaymckinney-enavate:
This changes corrects an issue when migrating GP Vendor 1099 data on version v25, but the new IRS Forms feature is disabled.
Fixes [AB#559817](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/559817)


nikolakukrika (changes requested):
The preprocessor symbols are not the root of the issue I believe. We need preprocessor symbols to automatically cleanup the code.

PR #27773, opened 2024-12-06
Exclude or Include Tables from Cloud Migration

### LINKED ISSUE #27655: [BC Idea]:
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=03befaf8-22a7-ef11-95f5-0022485288a6

### Description

We have moved our first customer from a legacy OnPrem solution to a AppSource Solution in the Cloud. We would like to pursue this goal further. In order to achieve that we have following requirements:

- We have multiple packages of tables for cloud migration that we want to run in chunks
- We do not want to migrate all the tables suggested in the default package - costs involved
- We do want to have the setup of those tables replicated for multiple customers and the manual selection/deselection for many customers for many tables is prone to errors.

Suggestion is to have a publicly available methods so that we could ship an extension that would enables us to:

- deselect the tables enabled in sync programmatically.
- select the tables in the AppSource solution and enable them programmatically.

First step of migration to a cloud is abandoning the OnPrem legacy solution and migrating to the AppSource tables but using OnPrem installation and then later syncing it to the Cloud.

Image

@nikolakukrika

- [x] I will provide the implementation for this BC Idea

miljance:
#### Summary
Enables a developer:
- To prepare packages of tables for cloud sync
- Disable some tables for cloud sync- those that are set as default during initial setup - in order to save database space.
- Select custom tables programmatically

#### Work Item(s)
Fixes #27655

@nikolakukrika There are some tests in the Codeunit 139656 "Hybrid Cloud Management Tests". I am not sure if new tests are needed for this change.

### DIFF
--- a/Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
+++ b/Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
@@ -71,6 +71,46 @@ codeunit 40021 "Cloud Mig. Replicate Data Mgt."
         exit(CanBeIncluded);
     end;
 
+    procedure CheckAndIncludeExcludeTableListFromCloudMigration(TableIDList: List of [Integer]; NewReplicateData: Boolean)
+    var
+        Company: Record Company;
+    begin
+        if Company.FindSet() then
+            repeat
+                CheckAndIncludeExcludeTableListFromCloudMigration(TableIDList, Company.Name, NewReplicateData);
+            until Company.Next() = 0;
+        CheckAndIncludeExcludeTableListFromCloudMigration(TableIDList, '', NewReplicateData);
+    end;
+
+    procedure CheckAndIncludeExcludeTableListFromCloudMigration(TableIDList: List of [Integer]; SelectedCompany: Text[30]; NewReplicateData: Boolean)
+    var
+        IntelligentCloudStatus: Record "Intelligent Cloud Status";
+        TableIDFilter: Text;
+        TableID: Integer;
+        i: Integer;
+        SeparatorChar: Char;
+    begin
+        if TableIDList.Count = 0 then
+            exit;
+
+        SeparatorChar := '|';
+        for i := 1 to TableIDList.Count do begin
+            TableIDList.Get(i, TableID);
+            TableIDFilter += Format(TableID) + SeparatorChar;
+        end;
+        TableIDFilter := TableIDFilter.TrimEnd(SeparatorChar);
+        IntelligentCloudStatus.SetRange("Company Name", SelectedCompany);
+        IntelligentCloudStatus.SetFilter("Table Id", TableIDFilter);
+        if IntelligentCloudStatus.FindSet() then
+            repeat
+                IntelligentCloudStatus.Mark(true);
+            until IntelligentCloudStatus.Next() = 0;
+        IntelligentCloudStatus.MarkedOnly();
+
+        CheckCanChangeTheTable(IntelligentCloudStatus);
+        IncludeExcludeTablesFromCloudMigration(IntelligentCloudStatus, NewReplicateData);
+    end;
+
     internal procedure IncludeExcludeTablesFromCloudMigration(var IntelligentCloudStatus: Record "Intelligent Cloud Status"; NewReplicateData: Boolean)
     var
         HybridCloudManagement: Codeunit "Hybrid Cloud Management";


Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
```diff
@@ -71,6 +71,46 @@ codeunit 40021 "Cloud Mig. Replicate Data Mgt."
         exit(CanBeIncluded);
     end;
 
+    procedure CheckAndIncludeExcludeTableListFromCloudMigration(TableIDList: List of [Integer]; NewReplicateData: Boolean)
```
nikolakukrika, line 74: I do not like this method. There is no strong reason to tie ourselves to a public interface that forces everyone to provide a list of integers. It can also cause a lot of issues when having thousands of records to update. 

Can we get a method that enables a single table, then you can decide if you want to write like this
IncludeTableToReplication(Database::"Change Log");
IncludeTableToReplication(Database::"Change Log Entry");
IncludeTableToReplication(Database::"My Change Log");
ExcludeTableFromReplciation(Database::"Customer);

We need 2 method types for replication:
IncludeTableToReplication(TableID)
ExcludeTableFromReplication(TableID)

Alternatevelly we need the following methods too, to specify how the delta sync should work:
SetReplaceCloudData(TableID)
SetPreserveCloudData(TableID) 

IncludeTableToReplication/ExcludeTableFromReplication could get an overload with preserve cloud data so you can set it in one go.

Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
```diff
@@ -71,6 +71,46 @@ codeunit 40021 "Cloud Mig. Replicate Data Mgt."
...
+    var
+        Company: Record Company;
```
nikolakukrika, line 76: This is incorrect as we will be updating all companies and not only the companies that are included in the migration. 
Maybe company name should become a part of the signature to add the table?

nikolakukrika (changes requested):

Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
miljance: I would actually prefer a function for a single table only. In this case you would probably get hundreds of telemetry entries. If you prefer that path anyway, I will change or better said simplify the function. My goal was to be aligned with how the current function works in the context of sending telemetry data.
image

Replace and Preserve - agreed. I can do that.

Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
miljance: We can either drop this function or do it in a different way. The idea was to be able to say - I want same packages of tables for all companies included in the migration.

Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
nikolakukrika: The problem with the list is that there are multiple properties that control how the table is replicated.

Currently there is:
Table ID and Company name identify a table
Replicate, Replace or delta sync control how the data is replicated.

There could be more in the future, I'm not sure how the table extensions are handled, so we may need an extra property. 

For these reasons a list of integers is not ideal. I would recommend that we implement dedicated methods to update a single table first, then we can always add overloads if needed.

Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
nikolakukrika: same as above, lets start with basic functions and add overloads where we feel that they are needed.

Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
miljance: Sounds even better as far as I am concerned.

miljance: "Per table" functions added.

Apps/W1/HybridBaseDeployment/app/src/codeunits/CloudMigReplicateDataMgt.Codeunit.al
miljance: "Per table" functions added.

JesperSchulz:
@nikolakukrika, you're up! Let's get this one reeled in!

PR #27876, opened 2025-01-03
[E-Document Core] - Show Purchase attachments in Peppol

GMatuleviciute:
This pull request does not have a related issue as it's part of delivery for development agreed directly with @altotovi @Groenbech96 

Implementation

This update ensures that when a Peppol file with Purchase document which contains an embedded PDF is received, the system extracts the PDF and displays it in assigned Factbox as attachment. This feature allows users to access embedded PDF before creating Purchase documents like invoices, orders or credit memos.

A new Factbox 'Doc. Attachment List' is added to the following pages:
'E-Document' (Page 6121)
'E-Documents' (Page 6122)

Fixes #


Apps/W1/EDocument/app/src/Processing/EDocAttachmentProcessor.Codeunit.al
```diff
@@ -126,6 +126,21 @@ codeunit 6169 "E-Doc. Attachment Processor"
...
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Document Attachment Mgmt", OnAfterGetRefTable, '', false, false)]
+    local procedure OnAfterGetRefTableForEDocs(var RecRef: RecordRef; DocumentAttachment: Record "Document Attachment")
+    var
+        EDocument: Record "E-Document";
+    begin
+        case DocumentAttachment."Table ID" of
+            Database::"E-Document":
+                begin
+                    RecRef.Open(Database::"E-Document");
+                    if EDocument.Get(DocumentAttachment."No.") then
```
pri-kise, line 138: Is the `DocumentAttachment."No."` really correct here?
Shouldn't it be `DocumentAttachment."E-Document Entry No."`?

GMatuleviciute: If e-document has no linked document in the attachments table Table ID is equal to E-Document table id and No. is equal to Entry No.
E-Document:
image
Attachments table:
image
If e-document has linked document then attachment record is renamed to have a link with related document.

miljance:
Side note: This change partially addresses the issue that the attachments should be visible on an E-Document. If the lines were not mapped during processing, the processing will fail, and the attachments will not appear in the Fact box for received e-doc. Ideally a user should be able to see the attachments in order to create mapping to GL for example.

Currently:
- User must create the mapping based on error logs
- Only after the document has been processed and purchase document has been created, he will see the attachments.

altotovi:
> Side note: This change partially addresses the issue that the attachments should be visible on an E-Document. If the lines were not mapped during processing, the processing will fail, and the attachments will not appear in the Fact box for received e-doc. Ideally a user should be able to see the attachments in order to create mapping to GL for example.
> 
> Currently:
> 
> * User must create the mapping based on error logs
> * Only after the document has been processed and purchase document has been created, he will see the attachments.

That's right. Attachment must be presented even if the processing is not finished. This is the main reason why we want to show attachment in the e-document.

miljance:
Thanks, does that mean that we agree that this change might be just a first step towards it?

altotovi:
> Thanks, does that mean that we agree that this change might be just a first step towards it?

Yes

Groenbech96:
@GMatuleviciute i merged the code in without the test. Had some issues with it. 
 Wait for https://github.com/microsoft/ALAppExtensions/pull/28124 to complete, then update the PR and verify the test. Then we will merge the test in.

Groenbech96:
Merged internally. Closing

PR #27952, opened 2025-01-17
Syncing with version 26.0.29027.0

aholstrup1:
Fixes [AB#560539](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/560539)


JesperSchulz (dismissed):

Apps/DE/EDocumentFormatXRechnung/app/src/ExportXRechnungDocument.Codeunit.al
```diff
@@ -0,0 +1,1117 @@
...
+        SalesInvLine.SetFilter(Quantity, '<>0');
+        if SalesInvLine.FindSet() then
+            exit(true);
+        exit(false);
+    end;
+
+    local procedure DocumentLinesExist(SalesCrMemoHeader: Record "Sales Cr.Memo Header"; var SalesCrMemoLine: Record "Sales Cr.Memo Line"): Boolean
+    begin
+        SalesCrMemoLine.SetRange("Document No.", SalesCrMemoHeader."No.");
+        SalesCrMemoLine.SetFilter(Type, '>%1', 0);
```
pri-kise, line 1103: Those filters here aren't the best practice.

We shouldn't use integer values for Enums.
Therefore I would suggest to use something like 
`SalesCrMemoLine.SetFilter(Type, '<>%1', SalesCrMemoLine.Type::" ");` instead of `SalesCrMemoLine.SetFilter(Type, '>%1', 0);`.


Furthemore the filter for an empty `No. `field is wrong.
it should be `SalesCrMemoLine.SetFilter("No.", '<>%1', '');` instead of `SalesCrMemoLine.SetFilter("No.", '<>%1', ' ');`
(I removed the space in the filter):
Have you planned to add there any events?

grobyns: apart from using enum values and not integers, please also use the SetFilter syntax consistently. This code uses both the 2 and 3 param overloads inconsistently. 
If you use the 2 param syntax, you get no type checking. It is preferred to use the 3 param one for that reason, so:
SetFilter(Quantity, '<>%1', 0) will check that 0 matches the type of Quantity...

ex:
SetFilter(Quantity, '<>O'); => this would compile
SetFilter(Quantity, '<>%1', O); => this would not

Apps/AT/ContosoCoffeeDemoDatasetAT/app/Demo Data/Finance/1. Setup Data/CreateVatPostingGroupAT.Codeunit.al
```diff
@@ -63,6 +63,27 @@ codeunit 11186 "Create VAT Posting Group AT"
...
+        CreatePostingGroup: Codeunit "Create Posting Groups";
+    begin
+        UpdateGenProdPostingGrp(CreatePostingGroup.FreightPostingGroup(), VAT20());
+        UpdateGenProdPostingGrp(CreatePostingGroup.RawMatPostingGroup(), VAT20());
+        UpdateGenProdPostingGrp(CreatePostingGroup.RetailPostingGroup(), VAT20());
+        UpdateGenProdPostingGrp(CreatePostingGroup.MiscPostingGroup(), VAT20());
+        UpdateGenProdPostingGrp(CreatePostingGroup.ServicesPostingGroup(), VAT10());
+    end;
+
+    local procedure UpdateGenProdPostingGrp(ProdPostingGroup: COde[20]; DefaultVATProdPostingGroup: Code[20])
```
IceOnly, line 77: nit: Typo here.
```suggestion
    local procedure UpdateGenProdPostingGrp(ProdPostingGroup: Code[20]; DefaultVATProdPostingGroup: Code[20])
```

IceOnly (changes requested):

Apps/DE/EDocumentFormatXRechnung/app/src/ExportXRechnungDocument.Codeunit.al
JesperSchulz: This is a sync between our internal repo and GitHub, and not a developer PS as such. That of course doesn't mean that the comments aren't valid, but it goes mean that the code already is checked in. I will create a bug for these comments.

PR #27972, opened 2025-01-22
[On Hold][E-Document Core] - Extending the Core app with payments

GMatuleviciute:

Implementation

The new functionality improves E-Document feature by introducing payment tracking. Users can now track both Incoming and Outgoing payments, with ability to monitor payment statuses by showing whether they are fully paid, partially paid or unpaid.

- New field "Payment Integration" was added to E-Document Services page together with "Calculate Payment VAT" boolean. For the "Payment Integration" a user can set a particular payment handler/service for sending and receiving payments. By enabling "Calculate Payment VAT" boolean, payment amount is split into Base Amount and VAT Amount.

- Additionally, the "Paid Amount" and "Payment Status" fields were added on E-Document page. On "Paid Amount" drilldown, user can access "E-Documents Payments" page, where payment date, paid amount without VAT and VAT amount are displayed, together with payment status and whether payment is incoming or outgoing. The payment date and amount fields can be edited manually, while the other fields are automatically populated by the system.

- New actions "Receive Payments" and "Send Payment" have also been added under the "Actions" menu on the E-Document page. "Receive Payments" actions allows user to receive payments via service and all data related to payment appears on "E-Document Payment" page, as well as updated payment status on E-Document. Meanwhile, "Send Payment" action can be used to filter unsent payments based on status and direction. The system adds E-Document and its values, identifies the right service for sending and updates the payment status of E-Document. This occurs even when the payment is created manually, regardless of whether it has been sent yet.

- Additional fields were added on "E-Documents Service" page, related to background job: "Auto Sync Payments", "Sync Start Time" and "Minutes between runs". The job queue is created when "Auto Sync Payments" boolean is enabled and same background job runs both sending and receiving actions.

Default integration is a placeholder and does not actually send or receive payments. New integrations can be added by extending enum "Payment Integration" and implementing interface IDocumentPaymentHandler.

Fixes #


Groenbech96:
On hold for now. We need more requirements from first country. 
Also we need to reconsider if we can create payments in journals, and use the existing interfaces to do so.

PR #27974, opened 2025-01-22
[E-Documents Core] - Enable deleting incorrect E-Documents

GMatuleviciute:

Implementation

A new functionality allows deletion of duplicate received E-Documents, defined as two documents having the same:

- Vendor
- External Document Number
- Date

Only users with "E-Document Core - Edit" permission set can delete received E-Documents, while the delete action will remain inactive for those without the required permissions.

- The E-Document cannot be deleted if its status is Processed.
- Deletion is not allowed for E-Documents linked to a purchase or sales document.
- A confirmation prompt will appear before deletion to prevent accidental removal.

This feature helps keep data accurate by allowing the removal of incorrect or extra E-Documents while ensuring proper control and security in the system.

Fixes #


Groenbech96:
What about all the logs, and all other associated data to the E-Document?

Groenbech96 (changes requested):
Added some comments.

Seems like this needs more thought. 
We get a lot of orphaned data this way.

AdeleStankeviciute:
Added procedure which deletes related logs and attachments when deleting E-Document.

Groenbech96 (approved):
LGTM

Groenbech96:
Running in our internal build right now.

Groenbech96:
Closing. Merged internally.

PR #28359, opened 2025-03-15
E-Document: Change accessibility level for EDoc WorkFlow

### LINKED ISSUE #28326: [Request-For-External] Codeunit 6139 "E-Document Workflow Setup" - Workflow codes from Internal to Global
### Describe the issue

I cannot access any functions on codeunit 6139 "E-Document Workflow Setup", since the `Access`-property is set to `Internal`.
I would like to have access to the following functions:
https://github.com/microsoft/ALAppExtensions/blob/329828d334aa7b1877e0bda1c6e3efb4d5a42a7b/Apps/W1/EDocument/app/src/Workflow/EDocumentWorkFlowSetup.Codeunit.al#L39-L62

### Expected behavior

To be able to access these functions.

### Steps to reproduce

1. Create a function with E-Document Workflow Setup as a variable.

```al
procedure EDocSendEDocResponseCode(): Code[128];
var
  EDocumentWorkflowSetup: Codeunit "E-Document Workflow Setup";
begin
end;
```
2. Observe [AL0161](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/diagnostics/diagnostic-al161): `'Codeunit Microsoft.eServices.EDocument."E-Document Workflow Setup"' is inaccessible due to its protection level`

### Additional context

I would like to avoid hardcoding the same codes in my extension (Where I depend on E-Document).

joandrsn:
#### Summary 
Enables access to hardcoded WorkFlow-names.
The changed code does not change the actual functions I want exposed, since the whole object is now Public. The functions I want are `EDocReceived`, `EDocImport`, `EDocSendEDocResponseCode`, `EDocCreated` & `EDocStatusChanged`

#### Work Item(s) 
Fixes #28326

### DIFF
--- a/Apps/W1/EDocument/app/src/Workflow/EDocumentWorkFlowSetup.Codeunit.al
+++ b/Apps/W1/EDocument/app/src/Workflow/EDocumentWorkFlowSetup.Codeunit.al
@@ -8,9 +8,9 @@ using System.Automation;
 
 codeunit 6139 "E-Document Workflow Setup"
 {
-    Access = Internal;
+    Access = Public;
 
-    procedure InsertSendToSingleServiceTemplate()
+    internal procedure InsertSendToSingleServiceTemplate()
     var
         Workflow: Record Workflow;
         WorkflowSetup: Codeunit "Workflow Setup";
@@ -23,7 +23,7 @@ codeunit 6139 "E-Document Workflow Setup"
         Workflow.Modify();
     end;
 
-    procedure InsertSendToMultiServiceTemplate()
+    internal procedure InsertSendToMultiServiceTemplate()
     var
         Workflow: Record Workflow;
         WorkflowSetup: Codeunit "Workflow Setup";
--- a/Apps/W1/EDocument/app/src/Workflow/EDocumentWorkFlowSetup.Codeunit.al
+++ b/Apps/W1/EDocument/app/src/Workflow/EDocumentWorkFlowSetup.Codeunit.al
@@ -36,12 +36,12 @@ codeunit 6139 "E-Document Workflow Setup"
         Workflow.Modify();
     end;
 
-    procedure EDocReceived(): Code[128]
+    internal procedure EDocReceived(): Code[128]
     begin
         exit('EDOCRECEIVED');
     end;
 
-    procedure EDocImport(): Code[128]
+    internal procedure EDocImport(): Code[128]
     begin
         exit('EDOCIMPORT');
     end;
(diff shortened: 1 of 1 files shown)


Apps/W1/EDocument/app/src/Workflow/EDocumentWorkFlowSetup.Codeunit.al
```diff
@@ -8,9 +8,9 @@ using System.Automation;
 
 codeunit 6139 "E-Document Workflow Setup"
 {
-    Access = Internal;
```
Groenbech96, line 11: Keep the EDocReceived, EDocImport and EDocStatusChanged internal.

Groenbech96 (changes requested):
Minor change then we can merge it in.

Groenbech96:
> Hi @Groenbech96.
> 
> That makes no sense.
> 
> I will try to explain the issue further:
> 
> The codeunit has the following internal procedures (Which can ONLY be called in this app):
> 
> ```
> InsertSendToSingleServiceTemplate
> InsertSendToMultiServiceTemplate
> EDocReceived
> EDocImport
> EDocSendEDocResponseCode
> EDocCreated
> EDocStatusChanged
> ```
> 
> Of these, I want to be able to have the 5 procedures public (For me to be able to avoid hardcoding the vales in my app):
> 
> ```
> EDocReceived
> EDocImport
> EDocSendEDocResponseCode
> EDocCreated
> EDocStatusChanged
> ```
> 
> In order to do this, I have to change the access of the whole object to `Public`, while marking the procedures I do NOT want to be public to be `internal`.
> 
> This PR does exactly this. Another way could have been to create a new codeunit and then expose those 5 procedures.
> 
> Hope to hear from you. Thank you!

Let me try and explain my reasoning:
The import and received procedures has been added, but are not integrated into the framework. So there is no reason to make it public, if we end up not wanting it. So I am curious why you need them?

joandrsn:
Thank you for the explaination @Groenbech96, that helped.
We are using the functions to push a default setup for our customers (Using Assisted Setup). We currently only use the workflow codes `EDOCSendEDOCRESPONSE` and `EDOCCREATEDEVENT`, so I have changed the visibility of the functions.
I just thought other customers/partners might have the same issue as us and therefore extended the visiblity to the other codes in the codeunit.

Groenbech96 (approved):
Approved.

Groenbech96:
merged internally.

PR #28433, opened 2025-03-28
Fix for GP 1099 issue

jaymckinney-enavate:
This PR fixes issues with the GP 1099 migration.
- If the new IRS Forms Feature is disabled, the original IRS form boxes are cleared and recreated, ensuring that they all get created.
- If the new IRS Forms Feature is enabled, don't set the IRS 1099 Code on the GenJournalLine since that is a deprecated field.

### DIFF
--- a/Apps/US/HybridGP_US/app/src/Codeunits/GPCloudMigrationUS.Codeunit.al
+++ b/Apps/US/HybridGP_US/app/src/Codeunits/GPCloudMigrationUS.Codeunit.al
@@ -107,6 +107,10 @@ codeunit 42004 "GP Cloud Migration US"
         if IsIRSFormsFeatureEnabled() then
             exit;
 
+        // Clear the table to ensure all available IRS 1099 Form-Box records are created
+        if not IRS1099FormBox.IsEmpty() then
+            IRS1099FormBox.DeleteAll();
+
         IRS1099FormBox.InitIRS1099FormBoxes();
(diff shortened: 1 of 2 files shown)


nikolakukrika (approved):

Apps/US/HybridGP_US/app/src/Codeunits/GPCloudMigrationUS.Codeunit.al
```diff
@@ -107,6 +107,10 @@ codeunit 42004 "GP Cloud Migration US"
         if IsIRSFormsFeatureEnabled() then
             exit;
 
+        // Clear the table to ensure all available IRS 1099 Form-Box records are created
+        if not IRS1099FormBox.IsEmpty() then
+            IRS1099FormBox.DeleteAll();
```
ventselartur, line 112: This is not right. In this case we delete all the codes, then insert only ones we are aware of and then migrate. In this case we will not migrate the custom codes the certain tenant has. Make sure that you insert missing codes, but do not remove the existing ones.

jaymckinney-enavate: This gets executed against a newly created company, created during the migration. No custom codes would exist at this point.

JesperSchulz:

Build ID: 32545.

PR #28450, opened 2025-04-01
Audit File Export. Remove the loop by export documents when restart task

### LINKED ISSUE #28339: [Bug]: Audit File Export Restart Task action restarts lines from other documents
### Describe the issue

The issue bases on next:
When we call  CurrPage.SetSelectionFilter(AuditFileExportLine)
System will filter record by primary key if we select only one record.
But in procedure RestartTaskOnExportLine. I don't know why remove filter for ID field. So, only filter by Line No. and status was applies.

I suggest remove logic from RestartTaskOnExportLine when we run loop for other documents. 
Or add in the onAction trigger filter by ID in second filtergroup

### Expected behavior

Only selected line will we scheduled to process.

### Steps to reproduce

Create two Audit Export Documents with parallel processing and generate lines.
Press restart task action on first line.
The system will schedule task also for lines in other documents

### Additional context

_No response_

vlkov:
#### Summary 
Remove the loop by export documents when restart task
#### Work Item(s) 
Fixes #28339 

Fixes [AB#573841](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/573841)

### DIFF
--- a/Apps/W1/AuditFileExport/app/src/ExportEngine/AuditFileExportMgt.Codeunit.al
+++ b/Apps/W1/AuditFileExport/app/src/ExportEngine/AuditFileExportMgt.Codeunit.al
@@ -717,10 +717,13 @@ codeunit 5261 "Audit File Export Mgt."
         exit(true);
     end;
 
-    local procedure CheckLineStatusForRestart(var AuditFileExportLine: Record "Audit File Export Line"): Boolean;
+    local procedure CheckLineStatusForRestart(var AuditFileExportLine: Record "Audit File Export Line"): Boolean
+    var
+        AuditFileExportLineCopy: Record "Audit File Export Line";
     begin
-        AuditFileExportLine.SetFilter(Status, '%1|%2', AuditFileExportLine.Status::"In Progress", AuditFileExportLine.Status::Completed);
-        if not AuditFileExportLine.IsEmpty() then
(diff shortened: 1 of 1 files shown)


JesperSchulz:
@ventselartur, could you review this code?

AleksanderGladkov:
@vlkov Looks good, thanks for the fix.
But to fix it fully, I would also update the CheckLineStatusForRestart() function, so it does not apply additional filter for the Status field.
My suggestion:

```
    local procedure CheckLineStatusForRestart(var AuditFileExportLine: Record "Audit File Export Line"): Boolean;
    var
        AuditExportLineCopy: Record "Audit File Export Line";
    begin
        AuditExportLineCopy.Copy(AuditFileExportLine);
        AuditExportLineCopy.SetFilter(Status, '%1|%2', AuditFileExportLine.Status::"In Progress", AuditFileExportLine.Status::Completed);
        if not AuditExportLineCopy.IsEmpty() then
            exit(HandleConfirm(StrSubstNo(TwoStringsTxt, LinesInProgressOrCompletedMsg, RestartExportLineQst)));
        exit(true);
    end;
```

AleksanderGladkov (changes requested):
Please check my previous comment

AleksanderGladkov (approved):

JesperSchulz (approved):

PR #28464, opened 2025-04-03
Avoid inconsistensies when posting auto acc in alternative currency

HederGit:
The Auto Acc funtionality sometimes create inconsistencies in the database when an alternative currency is used. To avoid this the code is updated with a validate to also set the GenJnlLine.Source Currency Amount.

#### Work Item(s) 

Inconsistencies when posting Auto Acc and Alternative Currency

Fixes [AB#573902](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/573902)

### DIFF
--- a/Apps/W1/AutomaticAccountCodes/app/src/Codeunits/AACodesPostingHelper.Codeunit.al
+++ b/Apps/W1/AutomaticAccountCodes/app/src/Codeunits/AACodesPostingHelper.Codeunit.al
@@ -71,6 +71,7 @@ codeunit 4850 "AA Codes Posting Helper"
         GenJnlCheckLine: Codeunit "Gen. Jnl.-Check Line";
         NoOfAutoAccounts: Decimal;
         TotalAmount: Decimal;
+        TotalAltAmount: Decimal;
         SourceCurrBaseAmount: Decimal;
         AccLine: Integer;
     begin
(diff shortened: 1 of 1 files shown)


JesperSchulz:
In order to process this pull request, we would need two things:
1) A GitHub issue which explains the issue, incl. repro steps. Based on that issue, the product group will determine if this is a suitable candidate to be fixed through a code contribution here on GitHub.
2) Once the issue is approved, a PR can be created (you jumped the gun there a bit 😉). The changes should incl. a test which demonstrates how the previous issue now has been addressed.

BrightComChristian:
Hi,
We've tried to reproduce this in CRONUS but it's hard unfortunately. There is a lot of sales order lines like 200+ and each with some auto accounts. This customer has been onprem first on BC17 and now on BC19. We've solved it by creating a new base version.
Maybe we could show you in the testmigration? I think the code is similar to the code in cu12 where there is a rounding but that's missing in this case.

JesperSchulz:
OK, we need a subject matter expert to look at this code then, as I cannot assess the validity of this code without repro or test cases.
@PredragMaricic, any chance you could look at this PR?

PredragMaricic (approved):

JesperSchulz (approved):

JesperSchulz:

Build ID: 32905.

PR #28517, opened 2025-04-10
[E-Document Connector] Logiq tests uptake

petemchlk:
This PR contains tests uptake to use new HTTP response handler.

Fixes #


petemchlk:
Created PR to early by mistake

PR #28522, opened 2025-04-14
email reminder/fin charge memo e-document

GMatuleviciute:

Implementation

With this update E-Documents are automatically created and emailed, when Issuing Reminders and Finance Charge Memos.

When the user clicks "Issue" action on the Reminder:

If they click OK without selecting an option in the "Print" field, the Reminder is issued, an E-Document is created and sent via the integrated service.
If they choose Print-> Email, the Reminder is issued, an E-Document is created, it is sent via the integrated service and email is sent with the E-Document attached.
When the user clicks "Issue" action on the Finance Charge Memo, exactly the same process happens as described above:

If they click OK without selecting an option in the "Print" field, the Finance Charge Memo is issued, an E-Document is created and sent via the integrated service.
If they choose Print-> Email, the Finance Charge Memo is issued, an E-Document is created, it is sent via the integrated service and email is sent with the E-Document attached.
Two new actions were added to "E-Document" Action Group in the Issued Reminder:

Create and Send E-Document: Creates an E-Document from the issued Reminder and sends it via integrated service.
Create and E-mail E-Document: Executes the same action as above, with an additional step to send an email containing the E-Document as an attachment.
The same new actions were added to "E-Document" Action Group in Issued Finance Charge:

Create and Send E-Document: Creates an E-Document from the issued Finance Charge Memo and sends it via integrated service.
Create and E-mail E-Document: Executes the same action as above, with an additional step to send an email with attached E-Document.
This update adds full E-Document support for Issued Reminders and Issued Finance Charge Memos, ensuring they meet Peppol standards while keeping financial document processing consistent.

Fixes #

### DIFF
--- a/Apps/W1/EDocument/app/src/Extensions/EDocIssuedFinChMemoHdr.TableExt.al
+++ b/Apps/W1/EDocument/app/src/Extensions/EDocIssuedFinChMemoHdr.TableExt.al
@@ -1,0 +1,78 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.eServices.EDocument;
+
+using Microsoft.Sales.FinanceCharge;
+using Microsoft.Sales.Customer;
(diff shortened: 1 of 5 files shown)


Groenbech96:
Internally Processed.

PR #28560, opened 2025-04-29
[E-Documents Core] - Export Sales Shipments as E-Document

GMatuleviciute:

**Implementation**

With this update E-Documents can be automatically created and sent to service when posting Sales Shipments.

This functionality works when user selects either "Ship" or "Ship and Invoice" option on posting Sales Order with correct customer setup. When using the second option, both invoice and shipment are exported as E-Documents.

New supported document type "Sales Shipment" was added to Enum "E-Document Type". It needs to be included in E-Document service configuration if user wants to export shipments.

New action group "E-Document" was added to "Posted Sales Shipment" page with action "Open E-Document" which opens related E-Document record.

Fixes #


Groenbech96:
Completed.

JesperSchulz:

Build ID: 37094.

PR #28583, opened 2025-05-05
[E-Documents Core] - Export Transfer Shipments as E-Document

GMatuleviciute:

**Implementation**

This update enables automatic creation and sending of E-Documents to the service when posting Transfer Shipments from Transfer Orders.
The functionality is triggered when the user selects the "Ship" option during posting, provided that a valid Document Sending Profile for Transfer Orders is set.

- A new action group "E-Document" has been added to the "Posted Sales Shipment" page, including the action "Open E-Document", which opens the related E-Document record.
- A new supported document type "Transfer Shipment" has been added to the "E-Document Type" Enum. To enable export functionality it must be included in the E-Document service configuration.
- Additionally, a new field "Transfer Doc. Sending Profile" has been introduced under the E-Documents section in the Warehouse FastTab of the Location card. This field specifies the document sending profile used specifically for Transfer Shipment documents.

Fixes #


JesperSchulz:

Build ID: 37094.

PR #28618, opened 2025-05-14
[E-Document Connector] SignUp Uptake HTTP Testing

GMatuleviciute:

Implementation

This update includes test adjustments for the SignUp connector, refactoring the tests to use HttpRequestHandlers.

Fixes [AB#557131](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/557131)


petemchlk:
@Groenbech96 we tried to make access token dummy here as well but unfortunately Logiq Connector App validates the bearer value:

image

image

Groenbech96:
@petemchlk any idea why we dont have the same number of files for mocking http? Seems that some files are not in this pr
image

petemchlk:
> @petemchlk any idea why we dont have the same number of files for mocking http? Seems that some files are not in this pr image

@Groenbech96 - I just used the files which were necessary to run the tests, though there is no point to store files which are not used

PR #28768, opened 2025-06-12
[Subscription Billing] - Account Payable Administrator extension with Vendor Contract action

petemchlk:
#### Summary 
These changes extend new Account Payable Administrator Role Center page with action 'Vendor Contract' which enables user to create new Vendor Contract directly from role center page.

#### Work Item(s) 
Fixes #


petemchlk:
created by mistake

PR #28770, opened 2025-06-15
Skip updating prices if "Subscription Option" has not changed

### LINKED ISSUE #28764: [Bug]: Subscription Billing - Item Template cannot create new item with "Allow Invoice Disc." = false
### Describe the issue

There are many fields in Item Template table that can be set on order to define the defaults for newly created item. A field "Allow Invoice Disc." is a field that exist in a template table since a long time but more recently the field "Subscription Option" has been introduced. The field "Subscription Option" has a default value of "No Subscription". Even if the value in this field should not have an impact on the process when using item template this field gets validated anyway. When this field is validated, it sets the value in the field "Allow Invoice Disc." thus overwriting the default from Item Template. That leads to the wrong result that regardless of the field "Allow Invoice Disc." in Item Template the value of field "Allow Invoice Disc." on Item is always set to true.

The s

### Expected behavior

The Item Template can have flag "Allow Invoice Disc." set to false and newly created item should take over this setting from Item Template.

### Steps to reproduce

- Create an item template with Allow Invoice Discount set to false
- Create new item based on this template
- Note that the Allow Invoice will remain true

### Additional context

_No response_

miljance:
#### Summary 
When validating the **"Subscription Option"** field, the code performs several checks. One of them is within the `UpdateItemPriceList` function, which enforces specific logic for **"Subscription Items"**:

- If the **"Subscription Option"** is set to **"Subscription Item",** the function ensures that the **"Allow Invoice Disc."** flag is **disabled** both on the Item and its related sales prices. This is because Subscription Items are excluded from document totals in Sales Quotes and Orders.
- Conversely, if the **"Subscription Option"** is _changed_ from **"Subscription Item"** to another value, the function resets **"Allow Invoice Disc."** to its default value (**true**).

This logic should execute **only when the "Subscription Option" value changes**.

However, when an Item is created from an **Item Template** where:
- **"Allow Invoice Disc."** is already set to **false**, and
- **"Subscription Option"** is **"No Subscription"**,

then this validation should **not run** and should **not modify sales prices** ΓÇö because no change to the **"Subscription Option"** has occurred.

#### Work Item(s) 
Fixes #28764

Fixes [AB#581599](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/581599)

### DIFF
--- a/Apps/W1/SubscriptionBilling/App/Service Objects/Table Extensions/Item.TableExt.al
+++ b/Apps/W1/SubscriptionBilling/App/Service Objects/Table Extensions/Item.TableExt.al
@@ -127,6 +127,9 @@ tableextension 8052 Item extends Item
         PriceCalculationMgt: Codeunit "Price Calculation Mgt.";
 #endif
     begin
+        if Rec."Subscription Option" = xRec."Subscription Option" then
+            exit;
+
         if IsServiceCommitmentItem() then begin
             Rec.Validate("Allow Invoice Disc.", false);
 #if not CLEAN25
--- a/Apps/W1/SubscriptionBilling/App/Service Objects/Table Extensions/ItemTempl.TableExt.al
+++ b/Apps/W1/SubscriptionBilling/App/Service Objects/Table Extensions/ItemTempl.TableExt.al
@@ -18,6 +18,13 @@ tableextension 8095 "Item Templ." extends "Item Templ."
                     DeleteItemTemplateServiceCommitmentPackages();
             end;
         }
+        modify("Allow Invoice Disc.")
+        {
+            trigger OnAfterValidate()
+            begin
+                Rec.ValidateItemField(FieldNo("Allow Invoice Disc."));
+            end;
+        }
     }
 
     local procedure DeleteItemTemplateServiceCommitmentPackages()


miljance:
@JesperSchulz I guess this is a Prio 1 issue as it affects the customers that do not have anything to do with Subscription Billing but want to default "Allow Invoice Disc." properly from Item Template. Needed for BC26.x I guess ASAP.

pri-kise (changes requested):
The pull request describes the issue and the fix very well, but shouldn't we have a test for this bugfix?

I guess you could check the `codeunit 138008 "Cust/Vend/Item/Empl Templates"` in the SMB Tests for the BaseApp.

And wouldn't we need a validation on the item template to prevent a wrong setup?
e.g. it shouldn't be possible to change the "Allow Inv. Disc." if a subscription option is set?
https://github.com/microsoft/ALAppExtensions/blob/main/Apps/W1/SubscriptionBilling/App/Service%20Objects/Table%20Extensions/ItemTempl.TableExt.al

miljance:
> The pull request describes the issue and the fix very well, but shouldn't we have a test for this bugfix?
> 
> I guess you could check the `codeunit 138008 "Cust/Vend/Item/Empl Templates"` in the SMB Tests for the BaseApp.
> 
> And wouldn't we need a validation on the item template to prevent a wrong setup? e.g. it shouldn't be possible to change the "Allow Inv. Disc." if a subscription option is set? https://github.com/microsoft/ALAppExtensions/blob/main/Apps/W1/SubscriptionBilling/App/Service%20Objects/Table%20Extensions/ItemTempl.TableExt.al

related PR on the BusinessCentralApps repo:
https://github.com/microsoft/BusinessCentralApps/pull/1421

pri-kise (approved):

JesperSchulz:

Build ID: 36172.

JesperSchulz:
Availability update: We will publish a fix for this issue in the next update for release 26.

Build ID to track: 36022.

PR #28883, opened 2025-07-13
Excel Trial Balance Starting Balance

### LINKED ISSUE #27561: [BC Idea]:
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=ec62231e-c43d-ef11-a295-000d3a059b20

### Description

It seems that since ever, the Trial Balance printed out the Net Change for given period and the balance on the upper limit of the date filter provided. However, in praxis it is needed that the balance on the lower limit of date filter provided is printed out, net change and the balance on the upper limit of the date filter provided.

Old:
image

Desired:
image

Currently there is an RDL Report but also an Excel Report. This change has been approved for Excel Report.

Reference: https://github.com/microsoft/BusinessCentralApps/issues/833

- [x] I will provide the implementation for this BC Idea

miljance:
#### Summary 
Added Starting Balance for Excel Trial Balance.
image

#### Work Item(s) 
Fixes #27561 

Fixes [AB#540995](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/540995)


miljance:
There is still an error in the Excel Trial Balance compared to the RDL Trial Balance. Therefore, a question arises if this change should be implemented just yet.

Attached there are printouts of an Excel Trial Balance and RDL Trial Balance. For a single G/L Account in this report there can only be Credit or Debit side for Net Change or Balance, not both. The reason for this error is wrong data model that is used for the first sheet. While the Dataset generated is just fine for the second sheet which is per dimension - "Trial Balance LCY (Analysis)" - this Dataset is cannot be used for the sheet "Trial Balance List (Print)" as it sums app credit and debit amounts per dimension. This makes the Excel Trial Balance unusable at the moment.
image
image

Apps/W1/ExcelReports/app/src/Financials/TrialBalance.Codeunit.al
```diff
@@ -124,8 +125,23 @@ codeunit 4410 "Trial Balance"
...
     local procedure InsertTrialBalanceDataForGLAccountWithFilters(var GLAccount: Record "G/L Account"; Dimension1ValueCode: Code[20]; Dimension2ValueCode: Code[20]; BusinessUnitCode: Code[20]; var TrialBalanceData: Record "EXR Trial Balance Buffer"; var Dimension1Values: Record "Dimension Value" temporary; var Dimension2Values: Record "Dimension Value" temporary)
+    var
+        GLAccount2: Record "G/L Account";
+        DateRec: Record Date;
     begin
         Clear(TrialBalanceData);
+        if GLAccount.GetFilter("Date Filter") <> '' then begin
+            DateRec.SetRange("Period Type", DateRec."Period Type"::Date);
+            DateRec.SetFilter("Period Start", GLAccount.GetFilter("Date Filter"));
+            DateRec.SetFilter("Period End", '%1..', 17530101D);
```
mynjj: why is this filter on "Period End" needed? also, maybe I'm missing something, but I guess it's not enough with `GetRangeMin`?

mynjj: Consider using it instead, it's a bit easier to read, if there are cases not covered that this approach covers then it's ok :)

mynjj (changes requested):
reviewing changes to xlsx is hard and could be a security vuln, so I propose the following:
- remove the xlsx from this PR
- attach separately the xlsx somewhere here
- provide a description of the changes to the xlsx

Then when this one get in in our builds I'll make sure to edit the xlsx appropriately.

Apps/W1/ExcelReports/app/src/Financials/TrialBalance.Codeunit.al
miljance: In the first draft I actually used GetRangeMin. However, this produces an error when there is no Min Date, Date Filter can be anything actually. After the first draft with GetRangeMin many Tests in "TrialBalanceExcelReports.Codeunit.al" started failing because "Date Filter" is not mandatory field prior to starting a report. The tests were very useful in that aspect as they prevented me to break current functionality. In a second try I did not have the line:
`DateRec.SetFilter("Period End", '%1..', 17530101D);`
This produced a runtime error as BC does not know about 00000000D and throws a runtime error.

So interested in exploring other options but for the moment I do not have a better idea.

miljance: I’ve reworked the logic to avoid using a hardcoded date, since I wasn’t happy with that approach either.

miljance:
> reviewing changes to xlsx is hard and could be a security vuln, so I propose the following:
> 
> * remove the xlsx from this PR
> * attach separately the xlsx somewhere here
> * provide a description of the changes to the xlsx (which I assume is adding the columns to the pivot table)
> 
> Then when this one get in in our builds I'll make sure to edit the xlsx appropriately.

I have removed changes for Excel in this PR.
The Excel file is attached to this message.
[TrialBalanceExcel.xlsx](https://github.com/user-attachments/files/21705918/TrialBalanceExcel.xlsx)

The changes inside Excel are:

1. Sheet $TrialBalancePrint$
image

2. Sheet $TrialBalanceLCYAnalysis$
image

3. Sheet $TrialBalanceACYSheetAnalysis$
image

4. Hidden sheet TrialBalanceData
In order to add colums in TrialBalanceData sheet a bit of magic and tinkering with Power Query Editor is needed. But I guess as Excel is not accepted because of security concerns and there is no official way how to add new columns in Excel Layout - Microsoft will handle this.
image

mynjj (approved):

Apps/W1/ExcelReports/app/src/Financials/TrialBalance.Codeunit.al
mynjj: It was a clever workaround :)

PR #28884, opened 2025-07-14
Syncing with version 27.0.36667.0

aholstrup1:
Fixes AB#560539


Apps/W1/Sustainability/app/src/Reports/SustItemEmissionBuffer.Table.al
```diff
@@ -0,0 +1,104 @@
...
+using Microsoft.Inventory.Ledger;
+using Microsoft.Purchases.Vendor;
+using Microsoft.Sales.Customer;
+using Microsoft.Sustainability.Setup;
+
+table 6250 "Sust. Item Emission Buffer"
+{
+    Caption = 'Sustainability Item Emission Buffer';
+    DataClassification = CustomerContent;
+    ReplicateData = false;
```
pri-kise, line 18: Is there any reason why this Buffer Table doesn't have the TableType temporary set?

(I know that this is only a sync PR, but maybe this could be addressed internally.)

mazhelez:
New syn available: https://github.com/microsoft/ALAppExtensions/pull/28920/files

PR #28920, opened 2025-07-22
Syncing with version 27.0.37019.0

Groenbech96:
Fixes AB#560539


Apps/CH/HybridBCLast_CH/app/app.json
```diff
@@ -34,4 +34,4 @@
   },
   "target": "OnPrem",
   "application": "27.0.0.0"
-}
\ No newline at end of file
+}
```
ankosiorek, line 37: Any reason we have empty changes?

PR #29070, opened 2025-08-17
Two step processing for E-Documents on sales side

miljance:
### Description

Sending an e-document using E-Documents module in sales means that as soon as a sales document is posted it will be sent to the service provider. This is fine in most of the cases. However, the cancelling process in some countries might be cumbersome. Therefore, it is desired that an E-document is sent out to the service provider only after it is assured that the invoice being sent is correct. With a new Workflow response, it is possible to create an E-Document during posting of a sales document without sending the document to the customer. If the posted invoice is ok - the action **"Send Document"** on E-Document card can be called manually.

#### Summary 

- New Status for E-Document: Exported.
- New Workflow response added: **'Init E-Document using setup'** which should be manually selected instead of the current one named **'Send E-Document using setup'**
- No new tests added as there was no test on sending side checking statuses - glad to add new ones if needed.
- Waiting for the iniital feedback.

#### Work Item(s) 
Fixes microsoft/BCApps#5004

Fixes [AB#580686](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/580686)

### DIFF
--- a/Apps/W1/EDocument/app/src/Document/EDocument.Page.al
+++ b/Apps/W1/EDocument/app/src/Document/EDocument.Page.al
@@ -544,6 +544,8 @@ page 6121 "E-Document"
                 StyleStatusTxt := 'Unfavorable';
             Rec.Status::Processed:
                 StyleStatusTxt := 'Favorable';
+            Rec.Status::Exported:
+                StyleStatusTxt := 'Ambiguous';
             else
                 StyleStatusTxt := 'None';
         end;
--- a/Apps/W1/EDocument/app/src/Document/Status/EDocExportedStatus.Codeunit.al
+++ b/Apps/W1/EDocument/app/src/Document/Status/EDocExportedStatus.Codeunit.al
@@ -1,0 +1,16 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.eServices.EDocument;
+
+/// <summary>
+/// E-Document Exported Status
+/// </summary>
+codeunit 6105 "E-Doc Exported Status" implements IEDocumentStatus
+{
+    procedure GetEDocumentStatus(): Enum "E-Document Status"
+    begin
+        exit(Enum::"E-Document Status"::Exported);
+    end;
+}
--- a/Apps/W1/EDocument/app/src/Document/Status/EDocumentServiceStatus.Enum.al
+++ b/Apps/W1/EDocument/app/src/Document/Status/EDocumentServiceStatus.Enum.al
@@ -14,7 +14,7 @@ enum 6106 "E-Document Service Status" implements IEDocumentStatus
     value(1; "Exported")
     {
         Caption = 'Exported';
-        Implementation = IEDocumentStatus = "E-Doc Processed Status";
+        Implementation = IEDocumentStatus = "E-Doc Exported Status";
     }
     value(2; "Sending Error")
     {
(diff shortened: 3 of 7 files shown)


JesperSchulz:
@Groenbech96, could you take a look?

Groenbech96:
@miljance sounds like a clearance model scenario?

Is it some manual check that is done outside BC ?

miljance:
> @miljance sounds like a clearance model scenario?
> 
> Is it some manual check that is done outside BC ?

No clearance outside BC. Just a manual check prior to sending it to the government portal. But if you are planning the clearance model with manual clearance in the BC that could also do the trick.

Groenbech96:
Ok. We just add some changes to the workflow bit and should be synced here https://github.com/microsoft/ALAppExtensions/pull/29098

Can you take a look at those and see how they make sense with your changes? :D

miljance:
Ah ok, that makes things a bit more different. There are two conflicting files, and I see that you have already introduced a similar Workflow response. I would need to test it with the new code. Also, I see you have not foreseen a new status for E-Document. 
Which status do you set when waiting for Clearance? In-progress or processed? For me it fehlt we need a new status indicating that the E-Documents is waiting for something, but In-progress felt wrong as it is also used for immediate sending.

image

miljance:
I have synced the changes with the new clearance model. As the clearance model provided all the needed functions, the change can be done much easier. 

- New Workflow Template
- Setting the status to Exported for E-Document.

@Groenbech96 Can you check if the new changes fit? Please evaluate the naming of the workflow step.

Groenbech96:
> I have synced the changes with the new clearance model. As the clearance model provided all the needed functions, the change can be done much easier.
> 
> * New Workflow Template
> * Setting the status to Exported for E-Document.
> 
> @Groenbech96 Can you check if the new changes fit? Please evaluate the naming of the workflow step.

Can you use the new response E-Document exported using X? We added that with support for clearance model?

Groenbech96:
> I have synced the changes with the new clearance model. As the clearance model provided all the needed functions, the change can be done much easier.
> 
> * New Workflow Template
> * Setting the status to Exported for E-Document.
> 
> @Groenbech96 Can you check if the new changes fit? Please evaluate the naming of the workflow step.

Do the new workflow step to export e-document 
Also @miljance we moved e-docs to bcapps :)

miljance:
> > I have synced the changes with the new clearance model. As the clearance model provided all the needed functions, the change can be done much easier.
> > 
> > * New Workflow Template
> > * Setting the status to Exported for E-Document.
> > 
> > @Groenbech96 Can you check if the new changes fit? Please evaluate the naming of the workflow step.
> 
> Do the new workflow step to export e-document Also @miljance we moved e-docs to bcapps :)

Hi @Groenbech96 - I can add another Workflow response but, in this case, it will do the same thing as for clearance model. So, if I create a new workflow response, do you expect it to do something else compared to clearance model? I mean the clearance model does exactly what I need - postpones the sending until resent manually. I am wondering, what is trigger for clearance model to actually send the document afterwards?

I also wondering if you see the changes I have made in regard to E-Document Status? The new Status called "Exported" might affect the clearance model as well.

My question, the clearance model exports the document first - and then - what is the trigger for actually sending the document?
image

miljance:
@JesperSchulz Can the PR and the Issue be moved to BCApps as the E-Docs module has been moved to BCApps?

JesperSchulz:
> @JesperSchulz Can the PR and the Issue be moved to BCApps as the E-Docs module has been moved to BCApps?

I'm afraid I can only move issues. The PR will have to get recreated.

miljance:
@JesperSchulz Thanks, I will create another PR over there.

PR #29087, opened 2025-08-21
External Storage - Document Attachments

StefanSosic:
#### Summary 
# Document Attachments External Storage for Microsoft Dynamics 365 Business Central

## Overview

The External Storage extension provides seamless integration between Microsoft Dynamics 365 Business Central and external storage systems such as Azure Blob Storage, SharePoint, and File Shares. This extension automatically manages document attachments by storing them in external storage systems while maintaining full functionality within Business Central.

### **Automatic Upload**
- Automatically uploads new document attachments to configured external storage
- Supports multiple storage connectors via the File Account framework
- Generates unique file names to prevent collisions
- Maintains original file metadata and associations

### **Flexible Deletion Policies**
- **Immediately**: Delete from internal storage right after external upload
- **1 Day**: Keep internally for 1 day before deletion
- **7 Days**: Keep internally for 7 days before deletion (default)
- **14 Days**: Keep internally for 14 days before deletion

### **Bulk Operations**
- Synchronize multiple files between internal and external storage
- Bulk upload to external storage
- Bulk download from external storage
- Progress tracking with detailed reporting

### Prerequisites
- Microsoft Dynamics 365 Business Central version 27.0 or later
- File Account module configured with external storage connector
- Appropriate permissions for file operations

### Installation Steps

1. **Configure File Account**
   - Open **File Accounts** page
   - Create a new File Account with your preferred connector:
     - Azure Blob Storage
     - SharePoint
     - File Share
   - Assign the account to **External Storage** scenario

2. **Configure External Storage**
   - Open **File Accounts** page
   - Select assigned **External Storage** scenario
   - Open **Additional Scenario Setup**
   - Configure settings:
     - **Auto Upload**: Enable automatic upload of new attachments
     - **Delete After**: Set retention policy for internal storage

#### Auto Upload Settings
- **Enabled**: New document attachments are automatically uploaded to external storage
- **Disabled**: Manual upload required via actions

### Automatic Mode
When Auto Upload is enabled:
1. User attaches a document to any Business Central record
2. System automatically uploads to external storage
3. File remains accessible through standard attachment functionality
4. Internal file is deleted based on configured retention policy

#### Individual File Operations
From **Document Attachment - External** page:
- **Upload to External Storage**: Upload selected file
- **Download from External Storage**: Download file for viewing
- **Download to Internal Storage**: Restore file to internal storage
- **Delete from External Storage**: Remove file from external storage
- **Delete from Internal Storage**: Remove file from internal storage

#### Bulk Operations
From **External Storage Synchronize** report:
- **To External Storage**: Upload multiple files to external storage
- **From External Storage**: Download multiple files from external storage
- **Delete Expired Files**: Clean up files based on retention policy

### File Access
- Files uploaded to external storage remain fully accessible through standard Business Central functionality
- Document preview, download, and management work seamlessly
- No change to end-user experience

**© 2025 Microsoft Corporation. All rights reserved.**

#### Work Item(s) 
Fixes [AB#592686](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/592686)


Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExtStor.TableExt.al
```diff
@@ -0,0 +1,74 @@
...
+{
+    fields
+    {
+        field(8750; "Uploaded Externally"; Boolean)
+        {
+            Caption = 'Uploaded Externally';
+            DataClassification = CustomerContent;
+            Editable = false;
+        }
+        field(8751; "External Upload Date"; Date)
```
pri-kise: I would suggest to use a DateTime here,

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExtStor.TableExt.al
```diff
@@ -0,0 +1,74 @@
...
+        {
+            Caption = 'External Upload Date';
+            DataClassification = CustomerContent;
+            Editable = false;
+        }
+        field(8752; "External File Path"; Text[250])
```
pri-kise: Are we sure that Text[250] is sufficent here?

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
```diff
@@ -0,0 +1,225 @@
...
+                {
+                    ApplicationArea = Basic, Suite;
+                    ToolTip = 'Specifies when the file was attached.';
+                }
+                field("Attached By"; Rec."Attached By")
+                {
+                    ApplicationArea = Basic, Suite;
+                    ToolTip = 'Specifies who attached the file.';
+                }
+                field("BCY Deleted Internally"; Rec."Deleted Internally")
```
pri-kise: I guess that the prefix `BCY` can be removed?

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
```diff
@@ -0,0 +1,225 @@
...
+                field("BCY External Upload Date"; Rec."External Upload Date")
+                {
+                    ApplicationArea = Basic, Suite;
+                    Caption = 'Upload Date';
+                    ToolTip = 'Specifies when the file was uploaded to external storage.';
+                }
+                field("BCY External File Path"; Rec."External File Path")
+                {
+                    ApplicationArea = Basic, Suite;
+                    Caption = 'External File Path';
```
pri-kise, line 72: The Caption an the ToolTips here of the own fields could be moved to the table level.

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
```diff
@@ -0,0 +1,225 @@
...
+            {
+                field("Table ID"; Rec."Table ID")
+                {
+                    ApplicationArea = Basic, Suite;
+                    ToolTip = 'Specifies the table ID the attachment belongs to.';
+                }
+                field("No."; Rec."No.")
+                {
+                    ApplicationArea = Basic, Suite;
+                    ToolTip = 'Specifies the record number the attachment belongs to.';
```
pri-kise, line 33: nit: if the ToolTips are moved to the table level in the BaseApp in a sparate PR then would could avoid introducing the same ToolTips here.

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
```diff
@@ -0,0 +1,225 @@
...
+    {
+        area(Processing)
+        {
+            action("Upload to External Storage")
+            {
+                ApplicationArea = Basic, Suite;
+                Promoted = true;
+                PromotedCategory = Process;
+                PromotedIsBig = true;
+                PromotedOnly = true;
```
pri-kise: I would suggest to use the new Syntax for Promoted Actions

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageSubs.Codeunit.al
```diff
@@ -0,0 +1,149 @@
...
+        // Check if it should be immediately deleted
+        if ExternalStorageProcessor.ShouldBeDeleted() then
+            ExternalStorageProcessor.DeleteFromInternalStorage(Rec);
+    end;
+
+    /// <summary>
+    /// Handles cleanup of external storage when document attachments are deleted.
+    /// Triggers on delete of Document Attachment records.
+    /// </summary>
+    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", 'OnAfterDeleteEvent', '', true, true)]
```
pri-kise, line 53: I'm not sure that we always want that.
We have a solution to store a reference to an external file, too.
Most of our customers don't want the file to be deleted if the record in BC is deleted.

What happens when an Document Attachment is transferred from one Document to the other one.
Will we have the file there two times with the same name...

Apps/W1/External Storage - Document Attachments/app/src/Permissions/DAExternalStorage.PermissionSet.al
```diff
@@ -0,0 +1,20 @@
...
+// ------------------------------------------------------------------------------------------------
+/// <summary>
+/// Permission set for External Storage functionality.
+/// Grants necessary permissions to use external storage features.
+/// </summary>
+permissionset 8750 "DA External Storage"
+{
+    Assignable = true;
+    Caption = 'Document Attachments - External Storage';
+    Permissions = tabledata "DA External Storage Setup" = RIMD,
```
pri-kise: We maybe should have separate permissions for View and for Setup/Admin.

Apps/W1/External Storage - Document Attachments/app/src/DAExtStorageDeleteAfter.Enum.al
```diff
@@ -0,0 +1,29 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+/// <summary>
+/// Defines when attachments should be deleted from internal storage after upload to external storage.
+/// </summary>
+enum 8750 "DA Ext. Storage - Delete After"
+{
+    Extensible = true;
```
pri-kise: I don't think that the enum is extensible in it's current form.

Currently you expect that the enum ordinal value is the number of days.
Therefore PTE could e.g. only create a new value for 50000 days or more.

For starters I would suggest to set this enum as extensible false.

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
StefanSosic: Yeah... leftover. Removed

Apps/W1/External Storage - Document Attachments/app/src/DAExtStorageDeleteAfter.Enum.al
StefanSosic: I would agree here, changed

Apps/W1/External Storage - Document Attachments/app/src/Permissions/DAExternalStorage.PermissionSet.al
StefanSosic: We can, added

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageSubs.Codeunit.al
StefanSosic: I could agree with Delete part here also, added setup Auto Delete, since we have Auto Upload. We can give this option.

When it get's transferred, nothing happens, due same name. When it's already existing it doesn't upload it by default. But adding one line more for this check.

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
StefanSosic: Yes, when the time comes, we would need to adjust everywhere and here also, true. Will make a note somewhere

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
StefanSosic: Agree, moved

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExtStor.TableExt.al
StefanSosic: I was in doubt, initially I put DateTime, but later changed. Makes still sense though... changed

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExtStor.TableExt.al
StefanSosic: It's not :D Changed to max of 2048, file name has limit of more then 1k chars, so including path, it can be more. But I would still put a limit of 2048 at this point, for now

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
StefanSosic: Definitely! Did for other actions, but this left... nice catch

StefanSosic:
> I couldn't find event subscribers for
> 
> * `OnBeforeExportToStream`
> * `OnBeforeGetAsTempBlob`
> * `OnBeforeGetContentType`
> * `OnBeforeHasContent` (not sur if we need this one)
> * etc. (I might have missed events)
> 
> if the file is exported then an app should still be able to call those procedures to get a file without knowing that this was exported to an external storage.

@pri-kise Good point. If you think about other events and places, let me know. Thank you. For this change won't push immediately, probably during the weekend, after good testing.

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageSubs.Codeunit.al
```diff
@@ -88,6 +88,110 @@ codeunit 8752 "DA External Storage Subs."
...
+    /// <summary>
+    /// Handles getting content as TempBlob for externally stored document attachments.
+    /// Downloads from external storage when internal content is not available.
+    /// </summary>
+    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", 'OnBeforeGetAsTempBlob', '', false, false)]
+    local procedure DocumentAttachment_OnBeforeGetAsTempBlob(var DocumentAttachment: Record "Document Attachment"; var TempBlob: Codeunit "Temp Blob"; var IsHandled: Boolean)
+    var
+        ExternalStorageProcessor: Codeunit "DA External Storage Processor";
+    begin
+        // Only handle if file is uploaded externally and not available internally
```
pri-kise, line 104: This code could be moved into one general helper procedure to check if the Document Attachment ist moved to the External Storage.
This applies to all the places.

pri-kise (changes requested):
We should have some test coverage for this new functionality.

I don't know if we already have a mocking Storage in the System Application, but we could add this here.

For starters I'm thinking about the following test cases

```al
// [GIVEN] Mocked External Storage (Maybe via a internal table or single instance codeunit)
// [GIVEN] Valid Setup for Document Attachments Exteranl Storage provided
// [GIVEN] Document Attachment is created
// [WHEN] Document Attachment sync runs
// [THEN] Verify Document Attachment is updated correctly
```

The following test case is an example for one procedure of the document attachment.
```al
// [GIVEN] Mocked External Storage
// [GIVEN] Valid Setup for Document Attachments Exteranl Storage provided
// [GIVEN] Document Attachment is created
// [GIVEN] Document Attachment synced
// {WHEN] TemplBlob is retrieved via procedure from Document Attachment 
// [THEN] Verify TempBlob
```
We should have tests for all the procedures that we handle.

Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageSubs.Codeunit.al
StefanSosic: Yes! Thought about that one also, not sure why I didn't go that way at the end. Implemented.

PR #29306, opened 2025-10-20
Mandatory Customer Data in Sales for xRechnung and ZUGFeRD

### LINKED ISSUE #29264: [Bug]: XRechnung – Invalid XML Generated When Customer is missing data.
### Describe the issue

When a Sales Invoice is posted for a customer record that is missing mandatory master data (e.g., address fields, email), the generated xRechnung XML file is invalid.  
The XML export does not enforce validation of required customer fields before posting, which results in missing elements in the XML.  
As a consequence, the file fails validation in the official xRechnung validator with multiple errors related to incomplete customer address information.

### Expected behavior

- A posted Sales Invoice should always produce a **valid xRechnung XML file**.  
- If mandatory customer data is missing, the system should **prevent posting** until the required information is completed.  
- Users should not discover XML errors only after posting; the export must remain compliant with posted document values.

## Repro
1. Use **Business Central for Germany**.  
2. Enable **E-Document Service** for XRechnung.  
3. In setup:  
   - Set **Buyer Reference Mandatory** = *Yes*  
   - Set **Buyer Reference** = *Your Reference*  
4. Configure a **Workflow** for XRechnung (per [docs](https://learn.microsoft.com/en-us/dynamics365/business-central/finance-how-setup-edocuments#set-up-the-workflow)).  
5. Configure a **Document Sending Profile** for XRechnung (per [docs](https://learn.microsoft.com/en-us/dynamics365/business-central/finance-how-setup-edocuments#set-up-a-document-sending-profile)) and mark it as **default**.  
6. Create a new domestic customer **without using customer templates**.  
7. Enter only a **name** for the customer (no address, no email).  
8. Assign posting groups: *Gen. Bus. Posting Group*, *VAT Bus. Posting Group*, and *Customer Posting Group* (e.g., INLAND).  
9. Create a new **Sales Invoice** for this customer and enter any value in **Your Reference**.  
10. Add a Sales Line with any value (e.g., **4.000,00 EUR**).  
11. Post the Sales Invoice.  
12. Open the **Posted Sales Invoice**.  
13. Navigate to **Related → E-Document → Open E-Document**.  
14. From **E-Document Logs**, export the XML file.  
15. Upload the file to the [xRechnung Validator](https://erechnungsvalidator.service-bw.de/).  
16. Observe the errors: validator reports multiple missing address fields.

Image

### Additional context

Fixes
[AB#609594](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609594)

miljance:
#### Summary 
* **Pre-posting validation:** Run Peppol validation before posting a sales document to ensure all required data is present (customer and Company Information), guaranteeing the exported E-Document is valid.
* **Customer email requirement:** For xRechnung, the customer’s email is mandatory and included in the XML. For ZUGFeRD, email is not mandatory in the schema, but operationally required to deliver the document to the customer.
* **ZUGFeRD due date export:** Export “Due Date” when no “Payment Terms” are defined for the customer, ensuring a valid ZUGFeRD format.
* **ZUGFeRD XML tag corrections:** Correct the XML mapping so IBAN is exported under PayeePartyCreditorFinancialAccount and SWIFT Code under PayeeSpecifiedCreditorFinancialInstitution, instead of both under PayeePartyCreditorFinancialAccount.

#### Work Item(s) 
Fixes #29264

Fixes [AB#609594](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609594)

### DIFF
--- a/Apps/DE/EDocumentDE/app/src/XRechnung/ExportXRechnungDocument.Codeunit.al
+++ b/Apps/DE/EDocumentDE/app/src/XRechnung/ExportXRechnungDocument.Codeunit.al
@@ -644,7 +644,7 @@ codeunit 13916 "Export XRechnung Document"
         AllowanceChargeElement := XmlElement.Create('AllowanceCharge', XmlNamespaceCAC);
         AllowanceChargeElement.Add(XmlElement.Create('ChargeIndicator', XmlNamespaceCBC, 'false'));
         AllowanceChargeElement.Add(XmlElement.Create('AllowanceChargeReason', XmlNamespaceCBC, AllowanceChargeReason));
-        AllowanceChargeElement.Add(XmlElement.Create('MultiplierFactorNumeric', XmlNamespaceCBC, FormatDecimal(MultiplierFactorNumeric)));
+        AllowanceChargeElement.Add(XmlElement.Create('MultiplierFactorNumeric', XmlNamespaceCBC, FormatFiveDecimal(MultiplierFactorNumeric)));
         AllowanceChargeElement.Add(XmlElement.Create('Amount', XmlNamespaceCBC, XmlAttribute.Create('currencyID', CurrencyCode), FormatDecimal(Amount)));
         AllowanceChargeElement.Add(XmlElement.Create('BaseAmount', XmlNamespaceCBC, XmlAttribute.Create('currencyID', CurrencyCode), FormatDecimal(BaseAmount)));
         if InsertTaxCat then
@@ -1025,6 +1025,11 @@ codeunit 13916 "Export XRechnung Document"
     procedure FormatFourDecimal(VarDecimal: Decimal): Text[30];
     begin
         exit(Format(Round(VarDecimal, 0.0001), 0, 9));
+    end;
+
+    procedure FormatFiveDecimal(VarDecimal: Decimal): Text[30];
+    begin
+        exit(Format(Round(VarDecimal, 0.00001), 0, 9));
     end;
 
     procedure GetUoMCode(UoMCode: Code[10]): Text;
--- a/Apps/DE/EDocumentDE/app/src/XRechnung/XRechnungFormat.Codeunit.al
+++ b/Apps/DE/EDocumentDE/app/src/XRechnung/XRechnungFormat.Codeunit.al
@@ -12,6 +12,7 @@ using Microsoft.Sales.Customer;
 using Microsoft.Foundation.Company;
 using Microsoft.eServices.EDocument;
 using Microsoft.eServices.EDocument.IO.Peppol;
+using Microsoft.Service.History;
 
 codeunit 13914 "XRechnung Format" implements "E-Document"
 {
@@ -91,6 +92,14 @@ codeunit 13914 "XRechnung Format" implements "E-Document"
             exit;
 
         if not EDocumentService."Buyer Reference Mandatory" then
+            exit;
+
+        if not (SourceDocumentHeader.Number in
+            [Database::"Sales Invoice Header",
+            Database::"Sales Cr.Memo Header",
+            Database::"Service Invoice Header",
+            Database::"Service Cr.Memo Header"])
+        then
             exit;
 
         case EDocumentService."Buyer Reference" of
--- a/Apps/DE/EDocumentDE/test/src/XRechnungXMLDocumentTests.Codeunit.al
+++ b/Apps/DE/EDocumentDE/test/src/XRechnungXMLDocumentTests.Codeunit.al
@@ -6,6 +6,8 @@ namespace Microsoft.eServices.EDocument.Formats;
 using Microsoft.Foundation.Company;
 using Microsoft.Sales.Customer;
 using Microsoft.Sales.History;
+using Microsoft.Purchases.Vendor;
+using Microsoft.Purchases.Document;
 using System.Utilities;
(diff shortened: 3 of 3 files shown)


Apps/DE/EDocumentDE/app/src/PEPPOL/EDocPEPPOLBIS30DE.Codeunit.al
```diff
@@ -237,6 +237,12 @@ codeunit 11035 "EDoc PEPPOL BIS 3.0 DE" implements "E-Document"
         IsHandled := true;
     end;
 
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"PEPPOL Validation", 'OnAfterCheckSalesDocument', '', false, false)]
```
pri-kise: Shouldn't this check be only manually binded? Otherwise this check will happen for any document. Even when another format is used.

The other integrations (event subscription) into the peppol validation should maybe be manuall binded, too.

This is at least my opinion.

Those additional checks should only happen when the specific formats are used.

If you want this additional check for XRechnung or ZUGFeRD then you need to add a manual event binding in the `Check` method.

I can understand that one can argue that you would only install the E-Document Extension for Germany when you intend to use those formats. (I can follow this argumnent, but I still wanted to tell you my concerns.)

Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
```diff
@@ -644,14 +644,12 @@ codeunit 13917 "Export ZUGFeRD Document"
...
     begin
-        if PaymentTermsCode = '' then
-            exit;
-        if not PaymentTerms.Get(PaymentTermsCode) then
-            exit;
-
         PaymentTermsElement := XmlElement.Create('SpecifiedTradePaymentTerms', XmlNamespaceRAM);
-        PaymentTermsDescriptionElement := XmlElement.Create('Description', XmlNamespaceRAM, PaymentTerms.Description);
-        PaymentTermsElement.Add(PaymentTermsDescriptionElement);
+        if PaymentTermsCode <> '' then
```
pri-kise, line 648: Shouldn't we ensure this bugfix via an addtional test case.
```al
// [SCENARIO] Export “Due Date” when no “Payment Terms” are defined for the customer, ensuring a valid ZUGFeRD format.
// [GIVEN] Sales Invoice without Payment Terms
// [WHEN] ZUGFerD is generated
// [THEN] Verify XML Data contains the Due Date
```

pri-kise (changes requested):

Apps/DE/EDocumentDE/app/src/PEPPOL/EDocPEPPOLBIS30DE.Codeunit.al
miljance: You’re absolutely right to raise this concern. I initially tried to keep it simple, since there’s already another subscriber behaving in a similar way (`SkipCheckOnCheckSalesDocumentOnBeforeCheckYourReference`).  

That said, I agree with your point — I’ll go ahead and manually bind both checks. At first, I didn’t want to introduce such a significant change, I just wanted to make sure the happy path worked. But your feedback makes sense, and this way the checks will only run when the specific formats (like XRechnung or ZUGFeRD or PEPPOL DE) are actually in use.  

Btw. I would not recommend mixing different connectors as I already know they generally interfere with each other.

Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
miljance: done

Apps/DE/EDocumentDE/test/src/ZUGFeRDXMLDocumentTests.Codeunit.al
```diff
@@ -187,6 +187,27 @@ codeunit 13922 "ZUGFeRD XML Document Tests"
...
+    var
+        SalesHeader: Record "Sales Header";
+        SalesInvoiceHeader: Record "Sales Invoice Header";
+        TempXMLBuffer: Record "XML Buffer" temporary;
+    begin
+        // [SCENARIO] Export "Due Date" when no "Payment Terms" are defined for the customer, ensuring a valid ZUGFeRD format.
+        Initialize();
+
+        // [GIVEN] Create and Post Sales Invoice without Payment Terms.
+        SalesHeader.Get("Sales Document Type"::Invoice, CreateSalesDocumentWithLine("Sales Document Type"::Invoice, Enum::"Sales Line Type"::Item, false));
```
pri-kise, line 201: @miljance the `CreateSalesDocumentWithLine` uses the local procedure `CreateSalesHeader` that manually assigns a Payment Term to the SalesHeader.
Do I miss something or shouldn't add something to clear the payment terms before posting the invoice?

miljance: you missed one commit 😛

PR #29395, opened 2025-11-10
GP - Post migration validation feature

jaymckinney-enavate:
Fixes [AB#579005](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/579005)
Fixes #579005

### DIFF
--- a/Apps/W1/HybridBaseDeployment/test/src/MockMigrationValidator.Codeunit.al
+++ b/Apps/W1/HybridBaseDeployment/test/src/MockMigrationValidator.Codeunit.al
@@ -1,4 +1,4 @@
-codeunit 139501 "Mock Migration Validator"
+codeunit 139633 "Mock Migration Validator"
 {
     trigger OnRun()
     begin


Apps/US/HybridGP_US/app/src/Codeunits/GPUSMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,188 @@
+namespace Microsoft.DataMigration.GP;
+
+using Microsoft.DataMigration;
+using Microsoft.Purchases.Vendor;
+using Microsoft.Purchases.Payables;
+using Microsoft.Finance.VAT.Reporting;
+
+codeunit 42006 "GP US Migration Validator"
```
nikolakukrika: It would be better to have IRS1099 validator. To keep the validations specific. US is a bit too broad.

Apps/US/HybridGP_US/app/src/Codeunits/GPUSMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,188 @@
...
+        GPFormAndBoxNo: Text[50];
+    begin
+        if CompanyValidationProgress.Get(CompanyNameTxt, ValidatorCodeLbl, ValidationStepVendor1099Lbl) then
+            exit;
+
+        EntityType := Vendor1099EntityCaptionLbl;
+        TaxPeriod := Format(GPCompanyAdditionalSettings.Get1099TaxYear());
+
+        // GP
+        if GPCompanyAdditionalSettings.GetMigrateVendor1099Enabled() then begin
```
nikolakukrika: Should we exit if it is not enabled?

Apps/US/HybridGP_US/app/src/Codeunits/GPUSMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,188 @@
...
+            GPPM00200.SetFilter(VENDORID, '<>%1', '');
+            if GPPM00200.FindSet() then
+                repeat
+                    if Vendor.Get(GPPM00200.VENDORID) then begin
+                        GPIRS1099Code := GPVendor1099MappingHelpers.GetIRS1099BoxCode(System.Date2DMY(System.Today(), 3), GPPM00200.TEN99TYPE, GPPM00200.TEN99BOXNUMBER);
+
+                        if not TempGPVendor.Get(Vendor."No.") then begin
+                            TempGPVendor."No." := Vendor."No.";
+                            TempGPVendor."Federal ID No." := CopyStr(GPPM00200.TXIDNMBR.TrimEnd(), 1, MaxStrLen(TempGPVendor."Federal ID No."));
+                            TempGPVendor.Insert();
```
nikolakukrika: This is dangerous. Even though the record is empty, we will fire triggers, we can persist the data in the other tables. 
It is better to keep a dictionary or a dedicated record to keep the data.

nikolakukrika: Second feedback on the temp records - they are taking a lot of memory. We should not store many records, max. 20-30.000 records. Dictionaries or smaller dedicated tables are much better. Vendor is a rather large table.

Apps/US/HybridGP_US/app/src/Codeunits/GPUSMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,188 @@
...
+                            AssignIRS1099CodeToVendor(GPIRS1099Code, Vendor, GPCompanyAdditionalSettings, TempIRS1099VendorFormBoxSetup);
+                        end;
+
+                        Clear(VendorYear1099AmountDictionary);
+                        BuildVendor1099Entries(Vendor."No.", VendorYear1099AmountDictionary);
+                        foreach GPIRS1099Code in VendorYear1099AmountDictionary.Keys() do begin
+                            TaxAmount := VendorYear1099AmountDictionary.Get(GPIRS1099Code);
+
+                            if TaxAmount > 0 then begin
+                                TempGPVendorLedgerEntry."Vendor No." := Vendor."No.";
```
nikolakukrika: Ledger entries will hurt, see the comments above

Apps/US/HybridGP_US/app/src/Codeunits/GPUSMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,188 @@
...
+                    continue;
+
+                if IRS1099VendorFormBoxSetup.Get(TaxPeriod, Vendor."No.") then
+                    BCFormAndBoxNo := StrSubstNo(ExpectedFormAndBoxNoTok, TaxPeriod, IRS1099VendorFormBoxSetup."Form No.", IRS1099VendorFormBoxSetup."Form Box No.");
+
+                if TempIRS1099VendorFormBoxSetup.Get(TaxPeriod, TempGPVendor."No.") then
+                    GPFormAndBoxNo := StrSubstNo(ExpectedFormAndBoxNoTok, TaxPeriod, TempIRS1099VendorFormBoxSetup."Form No.", TempIRS1099VendorFormBoxSetup."Form Box No.");
+
+                // Vendor: "Federal ID No.", "IRS 1099 Code"
+                MigrationValidationMgmt.ValidateAreEqual('VEND1099FEDIDNO', TempGPVendor."Federal ID No.", Vendor."Federal ID No.", FederalIdNoLbl);
```
nikolakukrika: We should not hardcode inline, we should use labels, they should be defined in a single place.

Apps/US/HybridGP_US/app/src/Codeunits/GPUSMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,188 @@
...
+        if not GPCompanyAdditionalSettings.Get(CompanyName()) then
+            exit;
+
+        ValidatorCodeLbl := GetValidatorCode();
+        CompanyNameTxt := CompanyName();
+
+        RunVendor1099MigrationValidation(GPCompanyAdditionalSettings);
+    end;
+
+    local procedure RunVendor1099MigrationValidation(var GPCompanyAdditionalSettings: Record "GP Company Additional Settings")
```
nikolakukrika: This test needs to be structured differently. 
We must get how far we got during the test, we should update the status every 500-1000 records checked to a table, so we can continue if the test fails. Ideally it should check record by record, not - prepare all data, then check the records. 

Without continuation it can easily be interrupted or fail. We could use the same setup record to show the progress - how far did it get.

Apps/US/HybridGP_US/app/src/Codeunits/HybridGPUSInstall.Codeunit.al
```diff
@@ -0,0 +1,61 @@
...
+    Access = Internal;
+    Subtype = Install;
+
+    trigger OnInstallAppPerDatabase()
+    begin
+        RegisterValidator();
+        RegisterTests();
+    end;
+
+    local procedure RegisterValidator()
```
nikolakukrika: This should not be necessary. The better approach is the interface and an enum. If we have an extensible enum, each of the extension can add its own.

We could also raise an event and populate the list of interfaces, or insert on the fly. The issue with OnInstall is that it is happening once, users cannot reinstall the app.

Apps/US/HybridGP_US/app/src/PageExt/IntelligentCloudExt.PageExt.al
```diff
@@ -17,6 +17,96 @@ pageextension 41101 "Intelligent Cloud Ext." extends "Intelligent Cloud Manageme
         }
     }
 
+    actions
+    {
+        addlast(Processing)
```
nikolakukrika: This code can be in the base solution. We can get which product is from the Intelligent Cloud Setup to be certain on which product we should run the validation.

Apps/US/HybridGP_US/app/src/PageExt/IntelligentCloudExt.PageExt.al
```diff
@@ -17,6 +17,96 @@ pageextension 41101 "Intelligent Cloud Ext." extends "Intelligent Cloud Manageme
...
+        {
+            action(RunAllVal)
```
nikolakukrika: RunAllValidation?

Apps/US/HybridGP_US/app/src/PageExt/IntelligentCloudExt.PageExt.al
```diff
@@ -17,6 +17,96 @@ pageextension 41101 "Intelligent Cloud Ext." extends "Intelligent Cloud Manageme
...
+                Image = Process;
+
+                trigger OnAction()
+                var
+                    MigrationValidationError: Record "Migration Validation Error";
+                    MigrationValidationMgmt: Codeunit "Migration Validation Mgmt.";
+                    HybridGPWizard: Codeunit "Hybrid GP Wizard";
+                begin
+                    MigrationValidationError.SetRange("Company Name", CompanyName());
+                    if not MigrationValidationError.IsEmpty() then
```
nikolakukrika: Minor - it is better to have the history per run, then to delete the entries. We could keep the previous results as part of the list. You could also create a log entry in the list and connect the results to the log entries.

nikolakukrika: Feel free to won't fix.

Apps/US/HybridGP_US/app/src/PageExt/IntelligentCloudExt.PageExt.al
```diff
@@ -17,6 +17,96 @@ pageextension 41101 "Intelligent Cloud Ext." extends "Intelligent Cloud Manageme
...
+                        MigrationValidationError.DeleteAll();
```
nikolakukrika: Now if the validation times out or it is interrupted, how do we continue?

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
```diff
@@ -2223,6 +2223,23 @@ codeunit 4001 "Hybrid Cloud Management"
...
+    var
+        MigrationValidationError: Record "Migration Validation Error";
+        CompanyValidationProgress: Record "Company Validation Progress";
+    begin
+        if MigrationType <> '' then
+            MigrationValidationError.SetRange("Migration Type", MigrationType);
+
+        MigrationValidationError.SetRange("Company Name", CompanyName());
+        if not MigrationValidationError.IsEmpty() then
+            MigrationValidationError.DeleteAll();
```
nikolakukrika: In general it is better to soft delete, e.g. set the property dismissed to true. Then in the UI you can choose if you want to see the old data or not.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidationMgmt.Codeunit.al
```diff
@@ -0,0 +1,345 @@
+namespace Microsoft.DataMigration;
+
+using System.Reflection;
+
+codeunit 40030 "Migration Validation Mgmt."
```
nikolakukrika: We can drom the Mgmt. from the name

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidationMgmt.Codeunit.al
```diff
@@ -0,0 +1,345 @@
...
+
+        if not MigrationValidatorRegistry.FindSet() then
+            exit;
+
+        Commit();
+        repeat
+            if UseSession then
+                Session.StartSession(SessionId, MigrationValidatorRegistry."Codeunit Id", Company)
+            else begin
+                Codeunit.Run(MigrationValidatorRegistry."Codeunit Id");
```
nikolakukrika: There are multiple issues here. 
1. We cannot start many background sessions, it will consume too many resources. We can max start few sessions in parallel. 
2. If the Codeunit.Run throws an error we will not proceed with other tests. We need to be able to generically handle the errors.

Apps/W1/HybridBaseDeployment/app/src/pages/CompanyMigrationStatus.Page.al
```diff
@@ -0,0 +1,82 @@
...
+                {
+                    ToolTip = 'Specifies the value of the Name field.';
+                }
+                field("Upgrade Status"; Rec."Upgrade Status")
+                {
+                    ToolTip = 'Specifies the value of the Upgrade Status field.';
+                }
+                field(Validated; Rec.Validated)
+                {
+                    ToolTip = 'Specifies the value of the Validated field.';
```
nikolakukrika: We could get a better tooltip here :)

Apps/W1/HybridBaseDeployment/app/src/pages/CompanyMigrationStatus.Page.al
```diff
@@ -0,0 +1,82 @@
...
+                var
+                    IntelligentCloudSetup: Record "Intelligent Cloud Setup";
+                    HybridCompanyStatus: Record "Hybrid Company Status";
+                    MigrationValidationMgmt: Codeunit "Migration Validation Mgmt.";
+                    SessionsStarted: Boolean;
+                begin
+                    if not IntelligentCloudSetup.Get() then
+                        exit;
+
+                    HybridCompanyStatus.SetFilter(Name, '<>%1', '');
```
nikolakukrika: I believe we have some code duplication here, it would be rather good if we could consolidate

Apps/W1/HybridBaseDeployment/app/src/pages/CompanyMigrationStatus.Page.al
```diff
@@ -0,0 +1,82 @@
...
+                    HybridCompanyStatus.SetRange(Validated, false);
+                    if not HybridCompanyStatus.FindSet() then begin
+                        Message('No companies need to be validated.');
+                        exit;
+                    end;
+
+                    repeat
+                        MigrationValidationMgmt.StartValidationSession(IntelligentCloudSetup."Product ID", HybridCompanyStatus.Name, true, false);
```
nikolakukrika: Similar comment here, we cannot start multiple sessions. It should be a single session that is rescheduling or maximum few in parallel.

Apps/W1/HybridBaseDeployment/app/src/pages/CompanyMigrationStatus.Page.al
```diff
@@ -0,0 +1,82 @@
...
+                        SessionsStarted := true;
+                    until HybridCompanyStatus.Next() = 0;
+
+                    if SessionsStarted then
+                        Message('Validation has started.');
```
nikolakukrika: We need a label here

Apps/W1/HybridBaseDeployment/app/src/pages/IntelligentCloudStatFactbox.Page.al
```diff
@@ -180,6 +181,37 @@ page 4008 "Intelligent Cloud Stat Factbox"
                             Page.Run(Page::"Hybrid Companies List");
                         end;
                     }
+
+                    field("Validation Errors"; ValidationErrors)
```
nikolakukrika, line 199: As we have talked we should get a status per test (validator) that can be drilled into specific validators.

Apps/W1/HybridGP/app/src/codeunits/GPMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,1620 @@
...
+        if not GPCompanyAdditionalSettings.Get(CompanyName()) then
+            exit;
+
+        ValidatorCodeLbl := GetValidatorCode();
+        CompanyNameTxt := CompanyName();
+        DefaultCurrency.InitRoundingPrecision();
+
+        GetUnpostedBatchCounts();
+
+        RunGLAccountMigrationValidation(GPCompanyAdditionalSettings);
```
nikolakukrika: Ideally this should not be done in a hardcoded way. We should get a list of records that list the available tests and invoke them via the interface.

Apps/W1/HybridGP/app/src/codeunits/GPMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,1620 @@
...
+                MigrationValidationMgmt.ValidateAreEqual('STATACCOUNTNAME', TempGPStatisticalAccount.Name, StatisticalAccount.Name, AccountNameLbl, true);
+                MigrationValidationMgmt.ValidateAreEqual('STATACCOUNTDIM1', TempGPStatisticalAccount."Global Dimension 1 Code", StatisticalAccount."Global Dimension 1 Code", Dimension1Lbl);
+                MigrationValidationMgmt.ValidateAreEqual('STATACCOUNTDIM2', TempGPStatisticalAccount."Global Dimension 2 Code", StatisticalAccount."Global Dimension 2 Code", Dimension2Lbl);
+                MigrationValidationMgmt.ValidateAreEqual('STATACCOUNTBALANCE', TempGPStatisticalAccount.Balance, StatisticalAccount.Balance, BeginningBalanceLbl, BalanceFailureShouldBeWarning);
+            until TempGPStatisticalAccount.Next() = 0;
+
+        LogValidationProgress(ValidationStepStatAccountLbl);
+    end;
+
+    local procedure RunBankAccountMigrationValidation(var GPCompanyAdditionalSettings: Record "GP Company Additional Settings")
```
nikolakukrika: There is too much nesting, these methods should be split in a similar way as IRS validation. 
We need to track progress so it can be continued.

Apps/W1/HybridBaseDeployment/app/src/pages/IntelligentCloudStatFactbox.Page.al
jaymckinney-enavate: This opens a list page that has drill down functionality per test.

Apps/W1/HybridGP/app/src/codeunits/GPMigrationValidator.Codeunit.al
jaymckinney-enavate: Each validator handles its own group of tests. In my opinion, this simpler approach is easier to understand, implement, and debug.

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
jaymckinney-enavate: The current design is to reduce confusion and noise. In my opinion it is the best approach when dealing with the migration partners.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidationMgmt.Codeunit.al
jaymckinney-enavate: 1. Getting rid of sessions.
2. Adding a Cloud Migration Warning implementation for this.

Apps/US/HybridGP_US/app/src/PageExt/IntelligentCloudExt.PageExt.al
jaymckinney-enavate: This action is just for testing

Apps/US/HybridGP_US/app/src/PageExt/IntelligentCloudExt.PageExt.al
jaymckinney-enavate: This action is just for testing

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
```diff
@@ -941,6 +942,14 @@ codeunit 4001 "Hybrid Cloud Management"
         OnInsertDefaultTableMappings(IntelligentCloudSetup."Product ID", DeleteExisting);
     end;
 
+    procedure PrepareMigrationValidation()
+        IntelligentCloudSetup: Record "Intelligent Cloud Setup";
+    begin
+        if IntelligentCloudSetup.Get() then;
```
nikolakukrika: we need a hard error here. We should not raise an event and not know in which data the setup is

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
```diff
@@ -2223,6 +2232,23 @@ codeunit 4001 "Hybrid Cloud Management"
...
+        CompanyValidationProgress: Record "Company Validation Progress";
+    begin
+        if MigrationType <> '' then
+            MigrationValidationError.SetRange("Migration Type", MigrationType);
+
+        MigrationValidationError.SetRange("Company Name", CompanyName());
+        if not MigrationValidationError.IsEmpty() then
+            MigrationValidationError.DeleteAll();
+
+        CompanyValidationProgress.SetRange("Company Name", CompanyName());
```
nikolakukrika: Maybe we could save a table and add addtional fields to hybrid company status?

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationBuffer.Table.al
```diff
@@ -0,0 +1,111 @@
...
+        field(1; "No."; Text[100])
+        {
+            Caption = 'No.';
+            NotBlank = true;
+        }
+        field(2; "Parent No."; Text[100])
+        {
+            Caption = 'Parent No.';
+        }
+        field(3; "Text 1"; Text[250])
```
nikolakukrika: We have 20248 if neeeded

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationBuffer.Table.al
```diff
@@ -0,0 +1,111 @@
+namespace Microsoft.DataMigration;
+
+table 40046 "Migration Validation Buffer"
```
nikolakukrika: This table is not the best approach. We need a different approach to store the data that needs to be validated.

Apps/US/HybridGP_US/app/src/Codeunits/GPUSMigrationValidator.Codeunit.al
jaymckinney-enavate: I redesigned all of the tests from 'load then test', to 'record by record'. I agree that it is much better.

All of the tests now also have continuation at the procedure/step level. These tests are currently very fast. I think adding a more granular record-level continuation would just slow down the tests. 

I'm marking this comment as Resolved, feel free to re-open if you have any concerns.

Apps/W1/HybridGP/app/src/pages/IntelligentCloudExtension.PageExt.al
```diff
@@ -52,8 +52,11 @@ pageextension 4015 "Intelligent Cloud Extension" extends "Intelligent Cloud Mana
 
                 trigger OnAction()
                 var
+                    HybridCloudManagement: Codeunit "Hybrid Cloud Management";
                     GPMigrationConfiguration: Page "GP Migration Configuration";
                 begin
+                    HybridCloudManagement.PrepareMigrationValidation(); // TODO: REMOVE AFTER LOCAL TESTING
```
nikolakukrika: TODO remains in the code

Apps/US/HybridGP_US/app/src/Codeunits/GPIRS1099MigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,203 @@
...
+codeunit 42006 "GP IRS1099 Migration Validator"
+{
+    trigger OnRun()
+    var
+        GPCompanyAdditionalSettings: Record "GP Company Additional Settings";
+    begin
+        if not GPCompanyAdditionalSettings.GetMigrateVendor1099Enabled() then
+            exit;
+
+        ValidatorCodeLbl := GetValidatorCode();
```
nikolakukrika: Nit - i it is not a lbl it is txt. 
Lbl is used in the UI

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
```diff
@@ -941,6 +942,14 @@ codeunit 4001 "Hybrid Cloud Management"
         OnInsertDefaultTableMappings(IntelligentCloudSetup."Product ID", DeleteExisting);
     end;
 
+    procedure PrepareMigrationValidation()
```
nikolakukrika: All of the methods added here should be moved to a new codeunit. The testing code should be moved to a dedicated subfolder in the solution. We should group the validation code together

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,509 @@
...
+    /// Start migration validation for the current company
+    /// </summary>
+    /// <param name="MigrationType">The type of migration</param>
+    /// <param name="RunOnlyAutomatic">Should only run validators that are set to run automatically</param>
+    procedure StartValidation(MigrationType: Text; RunOnlyAutomatic: Boolean)
+    var
+        MigrationValidatorRegistry: Record "Migration Validator Registry";
+        CloudMigrationWarning: Record "Cloud Migration Warning";
+    begin
+        MigrationValidatorRegistry.SetRange("Migration Type", MigrationType);
```
nikolakukrika: Should we check here if the migration supports the validators and if there are any validators available as a first line?

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,509 @@
...
+    begin
+        exit(Left = Right)
+    end;
+
+    local procedure IsNumber(Value: Variant): Boolean
+    begin
+        exit(Value.IsDecimal or Value.IsInteger or Value.IsChar)
+    end;
+
+    local procedure TypeOf(Value: Variant): Integer
```
nikolakukrika: Since these are a copy from the library assert, it would be good to move them to a codeunit so they are extracted from the rest of the validation code.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidatorWarning.Codeunit.al
```diff
@@ -0,0 +1,60 @@
...
+        MigrationValidatorWarningMsg: Label 'Number of migration validators that had an error during the last migration.';
+
+    procedure CheckWarning(): Boolean
+    begin
+        exit(GetWarningCount() > 0);
+    end;
+
+    procedure FixWarning()
+    begin
+        // Not sure what to put here.
```
nikolakukrika, line 15: Comments remaining

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidatorWarning.Codeunit.al
```diff
@@ -0,0 +1,60 @@
+namespace Microsoft.DataMigration;
+
+codeunit 40031 "Migration Validator Warning" implements "Cloud Migration Warning"
```
nikolakukrika, line 3: We already have warnings system, it would be good to align the 2 existing systems.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,509 @@
...
+
+        if RunOnlyAutomatic then
+            MigrationValidatorRegistry.SetRange(Automatic, true);
+
+        if not MigrationValidatorRegistry.FindSet() then
+            exit;
+
+        repeat
+            Commit();
+            if not Codeunit.Run(MigrationValidatorRegistry."Codeunit Id") then begin
```
nikolakukrika: We have talked about continuation.
GP may not need it, however the other products will most likely need the continuation. This is a generic codeunit that can be used by other migrations. We need to remember which tests were done and have a way to continue where we stopped.

Apps/W1/HybridBaseDeployment/app/src/pages/CompanyMigrationStatus.Page.al
```diff
@@ -0,0 +1,97 @@
+namespace Microsoft.DataMigration;
+
+page 40066 "Company Migration Status"
```
nikolakukrika: We have existing pages that are showing the status, can't we reuse any of them?

Apps/W1/HybridBaseDeployment/app/src/tables/CompanyValidationProgress.Table.al
```diff
@@ -0,0 +1,31 @@
...
+        }
+        field(2; "Validator Code"; Code[20])
+        {
+            Caption = 'Validator Code';
+        }
+        field(3; "Validation Step"; Code[20])
+        {
+            Caption = 'Validation Step';
+        }
+    }
```
nikolakukrika: We also need the status on the validator, so the checks can continue if interrupted.
The table is generic so it would be really good to get the support for this

Apps/W1/HybridBaseDeployment/test/src/DummyMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,96 @@
+codeunit 139501 "Dummy Migration Validator"
```
nikolakukrika: Nit - mock migration validator

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationBuffer.Table.al
jaymckinney-enavate: I'll update them to 2048. Can't guarantee it won't be needed.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidatorWarning.Codeunit.al
jaymckinney-enavate: This is in the base. I wanted to align with the Cloud Migration Warning feature there. I think eventually the GP warnings will move toward this new approach.

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
```diff
@@ -2245,6 +2271,49 @@ codeunit 4001 "Hybrid Cloud Management"
...
+        MigrationValidationError: Record "Migration Validation Error";
+        MigrationValidation: Codeunit "Migration Validation";
+    begin
+        if DataCreationFailed then
+            exit;
+
+        if not IntelligentCloudSetup.Get() then
+            exit;
+
+        MigrationValidation.StartValidation(IntelligentCloudSetup."Product ID", true);
```
nikolakukrika: I'm wondering if this process should be async?

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
```diff
@@ -2245,6 +2271,49 @@ codeunit 4001 "Hybrid Cloud Management"
...
         SendSetupWebhooksNotification.Send();
     end;
+    
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Data Migration Mgt.", OnBeforeMigrationStarted, '', false, false)]
+    local procedure BeforeMigrationStarted(var DataMigrationStatus: Record "Data Migration Status"; Retry: Boolean)
+    begin
+        ClearCompanyMigrationValidation(DataMigrationStatus."Migration Type");
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Data Migration Mgt.", OnValidateMigration, '', false, false)]
```
nikolakukrika: This is not a correct event to subscribe to.
This is the code where the event is defined:
  OnValidateMigration(DataMigrationStatus, DataCreationFailed);
    if DataCreationFailed then
        exit;

    OnAfterMigrationFinished(DataMigrationStatus, false, StartTime, Retry);
        
OnAfterMigrationFinished will move additional data. E.g. for GP IRS1099, Historical data, PostGLTransactions...  Subscribing to the event before will run some of the tests too early. 

The issue is that some processes are running Async, so I'm not sure if we can automatically subscribe to the validation. Maybe we will need to trigger these tests manually or via specific subscribers of each implementation engine.

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
```diff
@@ -2245,6 +2271,49 @@ codeunit 4001 "Hybrid Cloud Management"
...
+    begin
+        if DataCreationFailed then
+            exit;
+
+        if not IntelligentCloudSetup.Get() then
+            exit;
+
+        MigrationValidation.StartValidation(IntelligentCloudSetup."Product ID", true);
+
+        MigrationValidationError.SetRange("Migration Type", IntelligentCloudSetup."Product ID");
```
nikolakukrika: This code can impact the other data that is created, as it will stop the transformations.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,581 @@
...
+    /// <param name="RunOnlyAutomatic">Should only run validators that are set to run automatically</param>
+    procedure StartValidation(MigrationType: Text[250]; RunOnlyAutomatic: Boolean)
+    var
+        MigrationValidatorRegistry: Record "Migration Validator Registry";
+        CloudMigrationWarning: Record "Cloud Migration Warning";
+    begin
+        CommitAfterXValidatedRecordCount := GetDefaultCommitAfterXValidatedRecordCount();
+        OverrideCommitAfterXValidatedRecordCount := CommitAfterXValidatedRecordCount;
+
+        OnBeforeStartValidation(MigrationType, OverrideCommitAfterXValidatedRecordCount);
```
nikolakukrika: This is a bit unnecessary, we do not need the 2 variables. We could raise an event with a single variable.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,581 @@
...
+        if not MigrationValidatorRegistry.FindSet() then
+            exit;
+
+        repeat
+            Commit();
+            if not Codeunit.Run(MigrationValidatorRegistry."Codeunit Id") then begin
+                Clear(CloudMigrationWarning);
+                CloudMigrationWarning."Entry No." := 0;
+                CloudMigrationWarning."Warning Type" := CloudMigrationWarning."Warning Type"::"Migration Validator";
+                CloudMigrationWarning.Message := CopyStr(StrSubstNo(CloudMigrationWarningErr, MigrationValidatorRegistry."Validator Code", GetLastErrorText()), 1, MaxStrLen(CloudMigrationWarning.Message));
```
nikolakukrika: Should we list these as errors if the codeunit has failed running?
We may need a callstack too.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,581 @@
...
+    end;
+
+    /// <summary>
+    /// Validate that the actual value matches the expected value.
+    /// </summary>
+    /// <param name="Expected">The expected value for the test.</param>
+    /// <param name="Actual">The actual value being tested.</param>
+    /// <param name="TestDescription">Description of the test.</param>
+    /// <param name="ShouldBeWarning">Should the test be handled as a warning.</param>
+    procedure ValidateAreEqual(TestCode: Code[30]; Expected: Variant; Actual: Variant; TestDescription: Text[250]; ShouldBeWarning: Boolean): Boolean
```
nikolakukrika: Asserts should go to a separate codeunit. This codeunit is scheduling and running the tasks.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,581 @@
...
+        if IsHandled then begin
+            TimeoutDuration := OverrideTimeoutDuration;
+            MaxAttempts := OverrideMaxAttempts;
+            TimeBetweenScheduledJobs := OverrideTimeBetweenScheduledJobs;
+        end;
+
+        FailoverToSession := not CanStartBackgroundJob();
+
+        if not FailoverToSession then begin
+            SendStartValidationResultMessage('', StrSubstNo(TelemetryValidationToBeScheduledMsg, JobQueueLbl), false, false);
```
nikolakukrika: There are many blank values, is this intentional?

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidatorWarning.Codeunit.al
nikolakukrika: Maybe an interface call? The default implementation could repeat the text and say that the issue should be addressed manually.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidatorWarning.Codeunit.al
```diff
@@ -0,0 +1,60 @@
...
+    end;
+
+    procedure FixWarning()
+    begin
+        // Not sure what to put here.
+        // The migration validators should not fail, and if one does it cannot be fixed by the one running the migration.
+        // The issue should be reported.
+    end;
+
+    procedure ShowWarning(var CloudMigrationWarning: Record "Cloud Migration Warning"): Text
```
nikolakukrika, line 20: Nit - The method is not showing the warning it is returning text, the name should be udpated

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidatorWarning.Codeunit.al
```diff
@@ -0,0 +1,60 @@
...
+        SearchCloudMigrationWarning.SetRange("Warning Type", SearchCloudMigrationWarning."Warning Type"::"Migration Validator");
+        HybridReplicationSummary.SetCurrentKey("Start Time");
+        if HybridReplicationSummary.FindLast() then
+            SearchCloudMigrationWarning.SetFilter(SystemCreatedAt, '>%1', HybridReplicationSummary."End Time");
+
+        if not SearchCloudMigrationWarning.FindSet() then
+            exit;
+
+        repeat
+            FilterTxt := FilterTxt + Format(SearchCloudMigrationWarning."Entry No.") + '|'
```
nikolakukrika, line 35: FYI there is a limit on how many or statements we can have for search, it is around 1000 to 4000, I cannot remember exactly. Could we hit this issue?

Apps/W1/HybridBaseDeployment/app/src/pages/MigrationValidationErrors.Page.al
```diff
@@ -0,0 +1,86 @@
...
+                {
+                    ToolTip = 'Specifies the value of the Company Name field.';
+                }
+                field("Entity Type"; Rec."Entity Type")
+                {
+                    ToolTip = 'Specifies the value of the Entity Type field.';
+                }
+                field(Context; Rec.Context)
+                {
+                    ToolTip = 'Specifies the value of the Context field.';
```
nikolakukrika: Tooltips should be ideally on the table.
We should elaborate a bit more on them, not repeat the caption. It would be good to understand what the context is and for what it can be used for.

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationError.Table.al
```diff
@@ -0,0 +1,78 @@
...
+        field(5; "Migration Type"; Text[250])
+        {
+            Caption = 'Migration Type';
+        }
+        field(6; "Entity Type"; Text[50])
+        {
+            Caption = 'Entity Type';
+            NotBlank = true;
+        }
+        field(7; Context; Text[250])
```
nikolakukrika: As stated above, it would really be good to get the tooltips here.
What is context used for?
We need a field to store the additional information about the record that is specific, maybe even a callstack if it fails. These should be blobs. 

250 characters for context is a bit short.

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationError.Table.al
```diff
@@ -0,0 +1,78 @@
...
+                                                                     "Errors should fail migration" = const(true)));
+        }
+    }
+    keys
+    {
+        key(PK; "Entry No.")
+        {
+            Clustered = true;
+        }
+    }
```
nikolakukrika, line 95: We need secondary keys based on the filters that we are using to avoid long running queries

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationTest.Table.al
```diff
@@ -0,0 +1,48 @@
...
+    DataClassification = SystemMetadata;
+    DataPerCompany = false;
+
+    fields
+    {
+        field(1; "Code"; Code[30])
+        {
+            Caption = 'Code';
+        }
+        field(2; "Validator Code"; Code[20])
```
nikolakukrika: Tootlitps are missing here

Apps/W1/HybridBaseDeployment/app/src/tables/ValidationSuiteLine.Table.al
```diff
@@ -0,0 +1,48 @@
...
+    {
+        field(1; "Code"; Code[30])
+        {
+            Caption = 'Code';
+        }
+        field(2; "Validator Code"; Code[20])
+        {
+            Caption = 'Validator Code';
+        }
+        field(3; Entity; Text[50])
```
nikolakukrika, line 21: Wrong caption

Apps/W1/HybridBaseDeployment/app/src/tables/ValidationSuiteLine.Table.al
```diff
@@ -0,0 +1,48 @@
...
+{
+    Caption = 'Migration Validation Test';
+    DataClassification = SystemMetadata;
+    DataPerCompany = false;
+
+    fields
+    {
+        field(1; "Code"; Code[30])
+        {
+            Caption = 'Code';
```
nikolakukrika, line 13: Do we need field(2; "Migration Type"; Text[250]) here too?

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationTest.Table.al
```diff
@@ -0,0 +1,48 @@
+namespace Microsoft.DataMigration;
+
+table 40044 "Migration Validation Test"
```
nikolakukrika: What is the difference between this table and Migration Validator Registry? 
They look similar, for what should each of the tables be used for?

jaymckinney-enavate: Migration Validator Registry is to register a validator Codeunit and to give the user control over if validation is automatic and/or if validation errors should fail the migration.

Migration Validation Test is a detail table for each of the validator's tests.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
jaymckinney-enavate: I added entity level continuation per our discussion.

Apps/W1/HybridBaseDeployment/app/src/tables/CompanyValidationProgress.Table.al
jaymckinney-enavate: This has been implemented. The new table is Validation Progress, which adds better validator entity-level continuation as per our discussion.

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
jaymckinney-enavate: IRS 1099 and others should be using OnCreatePostMigrationData, which is right before OnValidateMigration.
In my view, OnAfterMigrationFinished is after all transforms and validation is complete.

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
jaymckinney-enavate: I don't think this is the case. Repeating my previous response. IRS 1099 and others should be using OnCreatePostMigrationData, which is right before OnValidateMigration.
In my view, OnAfterMigrationFinished is after all transforms and validation is complete.

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
jaymckinney-enavate: The automatic validation is intentionally synchronous to add the ability to fail the migration if it's configured to fail for validation errors.

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationTest.Table.al
jaymckinney-enavate: Added Tooltips

Apps/W1/HybridBaseDeployment/app/src/tables/ValidationSuiteLine.Table.al
jaymckinney-enavate: Corrected

Apps/W1/HybridBaseDeployment/app/src/tables/ValidationSuiteLine.Table.al
jaymckinney-enavate: Maybe not, it's using Validator Code as part of the key.
The Validator Registry table has the Migration Type.

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationError.Table.al
jaymckinney-enavate: Tooltips added.
The Context field is meant to be a text display of the entity Id. The UI isn't meant to drill down to the entity staging record being tested. Issues should most often be resolved in the old system, not in the staging tables.

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
nikolakukrika: We may need a new event after OnAfterMigrationFinished, something OnValidateAfterMigrationFinished.
I simply do not know how this will work. Some processes are triggered as the job queue or a scheduled task. 
Maybe it is the best to have the subscriber to invoke this functionality per implementation for now? 
GP knows when the migration is complete as it knows which tasks are running

Apps/W1/HybridBaseDeployment/app/src/codeunits/HybridCloudManagement.Codeunit.al
jaymckinney-enavate: This has been replaced with a better Validation Progress implementation that allows for entity-level continuation.

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationTest.Table.al
nikolakukrika: Can we merge these in a single table?
Ideally users should be able to pick the tests they want to run. Validator should be grouping of the tests. Having 2 tables seem excessive.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
jaymckinney-enavate: I don't know. I just wanted to reuse the existing Cloud Migration Warning functionality to show a failed validator at least somewhere.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
jaymckinney-enavate: I need a telemetry event Id. I forgot to bring that up.

Apps/US/HybridGP_US/app/src/Codeunits/GPIRS1099MigrationValidator.Codeunit.al
jaymckinney-enavate: Updated

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
jaymckinney-enavate: Refactored per our discussion

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationError.Table.al
jaymckinney-enavate: Agreed and updated

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidatorWarning.Codeunit.al
jaymckinney-enavate: There will be very few active validators and a failure should be very rare. A failed validator will not be the cause of this limit being exceeded.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidatorWarning.Codeunit.al
jaymckinney-enavate: That is part of the interface, I'm conforming to it. Is there a better way to implement it?

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationTest.Table.al
jaymckinney-enavate: Users should not need to select which tests to run. They will be preset by the app. For instance, the GP validators are currently responsible for 116 total tests. The user doesn't need to know that and/or need to select all 116.

jaymckinney-enavate: I changed the naming to Validation Suite and Validation Suite Line as discussed.

Apps/W1/HybridBaseDeployment/app/src/tables/MigrationValidationError.Table.al
jaymckinney-enavate: Updated the field name to Entity Display Name

Apps/W1/HybridBaseDeployment/test/src/MockMigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,103 @@
...
+        MigrationValidationAssert.SetSourceRowValidated(GetValidationSuiteId(), Customer);
+    end;
+
+    internal procedure GetValidationSuiteId(): Code[20]
+    begin
+        exit('TEST');
+    end;
+
+    // Normally initialized by this event, but called directly for testing.
+    //[EventSubscriber(ObjectType::Codeunit, Codeunit::"Hybrid Cloud Management", OnPrepareMigrationValidation, '', false, false)]
```
nikolakukrika: Commented out?

Apps/W1/HybridBaseDeployment/app/src/pages/IntelligentCloudStatFactbox.Page.al
```diff
@@ -194,17 +242,26 @@ page 4008 "Intelligent Cloud Stat Factbox"
         }
     }
 
-    trigger OnOpenPage()
+    trigger OnAfterGetRecord()
```
nikolakukrika: Are we sure that replacing this trigger is safe?
OnOpen page will fire before OnAfterGetRecord?
Can we keep the both triggers.

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,306 @@
...
+            ValidationProgress.DeleteAll(true);
+
+        if not HybridCompanyStatus.Get(Company) then
+            exit;
+
+        HybridCompanyStatus.Validate(Validated, false);
+        HybridCompanyStatus.Modify(true);
+    end;
+
+    procedure PrepareValidation()
```
nikolakukrika, line 167: Var is missing

Apps/US/HybridGP_US/app/src/Codeunits/GPIRS1099MigrationValidator.Codeunit.al
```diff
@@ -0,0 +1,190 @@
...
+                    MigrationValidationAssert.ValidateAreEqual(Test_VEND1099FEDIDNO_Tok, CopyStr(GPPM00200.TXIDNMBR.TrimEnd(), 1, MaxStrLen(Vendor."Federal ID No.")), Vendor."Federal ID No.", FederalIdNoLbl);
+
+                    Clear(VendorYear1099AmountDictionary);
+                    BuildVendor1099Entries(VendorNo, VendorYear1099AmountDictionary);
+                    foreach IRS1099Code in VendorYear1099AmountDictionary.Keys() do begin
+                        TaxAmount := VendorYear1099AmountDictionary.Get(IRS1099Code);
+
+                        if TaxAmount > 0 then begin
+                            Clear(VendorLedgerEntry);
+                            VendorLedgerEntry.SetLoadFields(Description, Amount);
```
nikolakukrika: SetLoadFields specifies Description and Amount, but Amount is a FlowField that should be loaded via CalcFields, not SetLoadFields. SetLoadFields only works with FieldClass = Normal fields.

Recommendation
Remove Amount from SetLoadFields since it's a FlowField. The CalcFields call on line 77 is the correct way to load it. Change to: VendorLedgerEntry.SetLoadFields(Description);

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,306 @@
...
+        if not JobQueueEntry.WritePermission then
+            exit(false);
+
+        exit(true);
+    end;
+
+    local procedure SendStartValidationResultMessage(MessageText: Text; IsError: Boolean; ShouldShowMessage: Boolean)
+    begin
+        if IsError then
+            Session.LogMessage('', MessageText, Verbosity::Error, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', CloudMigrationTok)
```
nikolakukrika: This is the new code 0000SEU

Apps/W1/HybridBaseDeployment/app/src/codeunits/MigrationValidation.Codeunit.al
```diff
@@ -0,0 +1,306 @@
...
+        else
+            Session.LogMessage('', MessageText, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', CloudMigrationTok);
```
nikolakukrika: This is the new code 0000SEV

JesperSchulz:
Nikola confirmed this is done. Closing this PR (as merge doesn't work). Will sync the repo right afterwards to get all lastest changes synced out to this repo.

PR #29466, opened 2025-11-26
xRechnung: Exporting Email and Data from Header

### LINKED ISSUE #29420: [Bug]: Customer email is not exported to ZUGFeRD XML
### Describe the issue

When creating a ZUGFeRD e-document with **Business Central for Germany** using the **E-Document Service**, the customer’s email address is not exported to the generated XML.

Validating the exported ZUGFeRD file shows the following:

- **Port Invoice Validator** (https://www.portinvoice.com/):  
  → error that the **Email is missing**.
- **erechnungs-validator** (https://erechnungs-validator.de/):  
  → the file is technically accepted, but a **warning** is shown that the email is missing.

There is currently no obvious way in Business Central to provide or map the email so that it appears in the ZUGFeRD XML in a way that satisfies these validators.

### Expected behavior

- When a ZUGFeRD e-document is generated, the **customer’s email address** should be exported into the ZUGFeRD XML.
- Commonly used validators (e.g. Port Invoice Validator, erechnungs-validator) should **not** report a missing-email error or warning, assuming the email exists in Business Central.

### Steps to reproduce

1. Use **Business Central for Germany**.  
2. Enable **E-Document Service** for ZUGFeRD.  
3. In setup:  
   - **Buyer Reference Mandatory** = *Yes*  
   - **Buyer Reference** = *Your Reference*  
4. Configure a **Workflow** for ZUGFeRD (per [docs](https://learn.microsoft.com/en-us/dynamics365/business-central/finance-how-setup-edocuments#set-up-the-workflow)).  
5. Configure a **Document Sending Profile** for ZUGFeRD (per [docs](https://learn.microsoft.com/en-us/dynamics365/business-central/finance-how-setup-edocuments#set-up-a-document-sending-profile)).
6. Mark this sending profile as the **default profile**.  
7. Assign the sending profile to a **Customer**.
8. Create a new **Sales Invoice** for the same **Customer**.  
9. Post the Sales Invoice.  
10. Open the **Posted Sales Invoice**.  
11. Navigate to **Related → E-Document → Open E-Document**.  
12. From **E-Document Logs**, export the XML file.  
13. Upload the file to the [Port Invoice Validator](https://www.portinvoice.com/).  
14. Observe the errors: Email is missing.

Some other validators like [erechnungs-validator](https://erechnungs-validator.de/) will recommend acepting the file, but will still give out a warning.

### Additional context

_No response_

miljance:
#### Summary 
**Refactored _InsertApplicableHeaderTradeAgreement_ :**

- Changed the method signature from accepting CustomerNo and YourReference parameters to accepting a RecordVariant parameter
- Added support for both Sales Invoice Header and Sales Cr.Memo Header using RecordRef and case statement
- Extracted all customer/header fields (Customer No., Name, Address, Post Code, City, Country Code, VAT Reg No., Your Reference, Contact, Email, Phone Number) directly from the header record instead of loading the Customer record separately
- Added email addresses for both Seller (Company) and Buyer (Customer) to the XML output using URIUniversalCommunication elements with schemeID='EM'
- Updated method calls to pass the header record instead of individual parameters

**Updated test support:**

- Added E-Mail field validation when creating test customers
- Added verification for Seller email in VerifySellerData
- Added verification for Buyer email in VerifyBuyerData methods
- Set "Buyer Reference Mandatory" to true when configuring the E-Document Service

**Summary**: Corrected the implementation to include email addresses in the ZUGFeRD XML output for both the seller (company) and buyer (customer), making the electronic documents more complete and compliant with ZUGFeRD format requirements.

#### Work Item(s) 
Fixes #29420
[AB#614338](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/614338)

### DIFF
--- a/Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
+++ b/Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
@@ -311,35 +311,37 @@ codeunit 13917 "Export ZUGFeRD Document"
         case HeaderRecordRef.Number of
             Database::"Sales Invoice Header":
                 begin
-                    CustomerNo := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("Bill-to Customer No.")).Value;
-                    CustomerName := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("Bill-to Name")).Value;
-                    Address := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("Bill-to Address")).Value;
-                    Address2 := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("Bill-to Address 2")).Value;
-                    PostCode := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("Bill-to Post Code")).Value;
-                    City := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("Bill-to City")).Value;
-                    CountryCode := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("VAT Country/Region Code")).Value;
-                    VATRegistrationNo := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("VAT Registration No.")).Value;
-                    YourReference := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("Your Reference")).Value;
-                    Contact := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("Sell-to Contact")).Value;
+                    HeaderRecordRef.SetTable(SalesInvoiceHeader);
+                    CustomerNo := SalesInvoiceHeader."Bill-to Customer No.";
+                    CustomerName := SalesInvoiceHeader."Bill-to Name";
+                    Address := SalesInvoiceHeader."Bill-to Address";
+                    Address2 := SalesInvoiceHeader."Bill-to Address 2";
+                    PostCode := SalesInvoiceHeader."Bill-to Post Code";
+                    City := SalesInvoiceHeader."Bill-to City";
+                    CountryCode := SalesInvoiceHeader."VAT Country/Region Code";
+                    VATRegistrationNo := SalesInvoiceHeader."VAT Registration No.";
+                    YourReference := SalesInvoiceHeader."Your Reference";
+                    Contact := SalesInvoiceHeader."Sell-to Contact";
                     ReportSelections.FindEmailBodyUsageForCust("Report Selection Usage"::"S.Invoice", CustomerNo, TempBodyReportSelections);
                     CustomerEmail := ReportSelections.GetEmailAddressExt("Report Selection Usage"::"S.Invoice".AsInteger(), RecordVariant, CustomerNo, TempBodyReportSelections);
-                    PhoneNumber := HeaderRecordRef.Field(SalesInvoiceHeader.FieldNo("Sell-to Phone No.")).Value;
+                    PhoneNumber := SalesInvoiceHeader."Sell-to Phone No.";
                 end;
             Database::"Sales Cr.Memo Header":
                 begin
-                    CustomerNo := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to Customer No.")).Value;
-                    CustomerName := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to Name")).Value;
-                    Address := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to Address")).Value;
-                    Address2 := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to Address 2")).Value;
-                    PostCode := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to Post Code")).Value;
-                    City := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to City")).Value;
-                    CountryCode := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("VAT Country/Region Code")).Value;
-                    VATRegistrationNo := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("VAT Registration No.")).Value;
-                    YourReference := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Your Reference")).Value;
-                    Contact := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Sell-to Contact")).Value;
+                    HeaderRecordRef.SetTable(SalesCrMemoHeader);
+                    CustomerNo := SalesCrMemoHeader."Bill-to Customer No.";
+                    CustomerName := SalesCrMemoHeader."Bill-to Name";
+                    Address := SalesCrMemoHeader."Bill-to Address";
+                    Address2 := SalesCrMemoHeader."Bill-to Address 2";
+                    PostCode := SalesCrMemoHeader."Bill-to Post Code";
+                    City := SalesCrMemoHeader."Bill-to City";
+                    CountryCode := SalesCrMemoHeader."VAT Country/Region Code";
+                    VATRegistrationNo := SalesCrMemoHeader."VAT Registration No.";
+                    YourReference := SalesCrMemoHeader."Your Reference";
+                    Contact := SalesCrMemoHeader."Sell-to Contact";
                     ReportSelections.FindEmailBodyUsageForCust("Report Selection Usage"::"S.Cr.Memo", CustomerNo, TempBodyReportSelections);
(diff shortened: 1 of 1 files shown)


djukicmilica (dismissed):

Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
```diff
@@ -283,13 +283,60 @@ codeunit 13917 "Export ZUGFeRD Document"
...
+                    CustomerNo := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to Customer No.")).Value;
+                    CustomerName := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to Name")).Value;
+                    Address := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to Address")).Value;
+                    PostCode := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to Post Code")).Value;
+                    City := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Bill-to City")).Value;
+                    CountryCode := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("VAT Country/Region Code")).Value;
+                    VATRegistrationNo := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("VAT Registration No.")).Value;
+                    YourReference := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Your Reference")).Value;
+                    Contact := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Sell-to Contact")).Value;
+                    CustomerEmail := HeaderRecordRef.Field(SalesCrMemoHeader.FieldNo("Sell-to E-Mail")).Value;
```
duichwer: Wouldn't it make more sense to use the email address that is found via the get email address procedures in table report selections ?
Then we would use here the same email like the one where we would send the document to.

Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
```diff
@@ -331,33 +384,38 @@ codeunit 13917 "Export ZUGFeRD Document"
...
+
         // Buyer Address
         PostalTradeAddressElement := XmlElement.Create('PostalTradeAddress', XmlNamespaceRAM);
-        PostalTradeAddressElement.Add(XmlElement.Create('PostcodeCode', XmlNamespaceRAM, Customer."Post Code"));
-        PostalTradeAddressElement.Add(XmlElement.Create('LineOne', XmlNamespaceRAM, Customer.Address));
-        PostalTradeAddressElement.Add(XmlElement.Create('CityName', XmlNamespaceRAM, Customer.City));
-        PostalTradeAddressElement.Add(XmlElement.Create('CountryID', XmlNamespaceRAM, GetCountryRegionCode(Customer."Country/Region Code")));
+        PostalTradeAddressElement.Add(XmlElement.Create('PostcodeCode', XmlNamespaceRAM, PostCode));
+        PostalTradeAddressElement.Add(XmlElement.Create('LineOne', XmlNamespaceRAM, Address));
+        PostalTradeAddressElement.Add(XmlElement.Create('CityName', XmlNamespaceRAM, City));
```
duichwer, line 419: This could be a follow up issue, but do I see this right that we currently don't provide the address 2 in the LineTwo?

Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
miljance: Nice point, I would need to checkup on this. If this is to be done, I would do the same for xRechnung and ZUGFeRD and perhaps even PEPPOLDE.

Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
miljance: I will check if there is a placeholder for Address 2 in ZUGFeRD. The same would make sense for xRechnung I guess.

duichwer: I have problems finding official documentation - if you search for _LineTwo_ or BT-51 then you should find some documentation.

miljance: I had problems too. That is why I did not primarily search for it :). I just used AI to suggest me where to place new elements and then I rechecked using online validators.

image

Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
duichwer: Just to make it clear what I meant with those procedures:

Call the methods for email address retrieval similar to GetEmailBodyTextForCust

1. FindEmailBodyUsageForCust
2. Use TempReportSrlections with GetEmailAddressExt

Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
miljance: I can confirm this works and passes the validators.

Apps/DE/EDocumentDE/app/src/ZUGFeRD/ExportZUGFeRDDocument.Codeunit.al
miljance: From what I see, the final e-mail address is determined by GetEmailAddress on table "Report Selections", which resolves the address in this order:

1. EmailParameter table (currently used only in tests)
2. Document Layouts for the customer
3. If none of the above apply, "Sell-to E-Mail" on the document

For xRechnung and ZUGFeRD we already enforce that "Sell-to E-Mail" must be filled before posting a sales document. 

The open questions for me are:
- Should we still enforce "Sell-to E-Mail" even when a Document Layout e-mail exists? My gut feeling is yes, because we still want a general contact address on the customer, not only for E-Documents.
- Should we use exactly the same pattern (i.e. the same lookup logic as GetEmailAddress) when choosing the e-mail that ends up in the XML?

If you think we should fully align with the report-selection logic, I’m happy to adjust the implementation accordingly. What do you think?

duichwer: If we want to avoid that VAR or Is a will create a copy of those objects then we should stick to the pattern that is used in the BaseApp for email retrieval.


I think it's still okay to check the sell-to email field.
One could argue what happens when the document layout is deleted .

miljance: Exactly my thinking. Document Layout seems to be loose from this story. And you could also change it in the meanwhile after the E-Document has already been created. So, it does not appear to be a bulletproof concept but at least you can recreate the E-Document if something like that happens.  Then I will give it a try.

Apps/DE/EDocumentDE/app/src/XRechnung/ExportXRechnungDocument.Codeunit.al
```diff
@@ -597,10 +604,12 @@ codeunit 13916 "Export XRechnung Document"
...
             CustomerGLN := Customer.GLN;
+        ReportSelections.FindEmailBodyUsageForCust("Report Selection Usage"::"S.Cr.Memo", SalesCrMemoHeader."Bill-to Customer No.", TempBodyReportSelections);
         InsertCustomerParty(
-            AccountingCustomerParty, Customer."VAT Registration No.", CustomerGLN,
-            SalesCrMemoHeader."Bill-to Name", BillToAddress, SalesCrMemoHeader."Sell-to Customer Name",
-            SalesCrMemoHeader."Sell-to Contact", SalesCrMemoHeader."Sell-to E-Mail");
+              AccountingCustomerParty, Customer."VAT Registration No.", CustomerGLN,
+              SalesCrMemoHeader."Bill-to Name", BillToAddress, SalesCrMemoHeader."Sell-to Customer Name",
+              SalesCrMemoHeader."Sell-to Contact",
+              ReportSelections.GetEmailAddressExt("Report Selection Usage"::"S.Cr.Memo".AsInteger(), SalesCrMemoHeader, SalesCrMemoHeader."Bill-to Customer No.", TempBodyReportSelections));
```
duichwer, line 612: You should update the other procedures to accept a value for the e mail with more than 80 characters. The procedure GetEmailAddressExt has Text 250 as return value.
This shouldn't be a problem for XRechnung

duichwer: @miljance you need to update InsertContact, too

duichwer (approved):

djukicmilica (approved):

JesperSchulz:
This PR has been merged internally. Trying to merge here on GitHub too.

PR #29653, opened 2026-01-28
[Review only] Multiple VAT Number refactoring

ventselartur:
This pull request contains the changes which are only to show to partners and then abandon the PR.
The idea of the changes it to avoid making the interface implementation public and instead move the methods with integration events out to the facade codeunit.


JesperSchulz:
Cleaning out this repo for archiving. Will close this PR.

PR #29666, opened 2026-02-04
Fix typos and update labels for Northern Ireland and XI in CreateCountryRegion codeunit

### LINKED ISSUE #29496: [Bug]: The demo country/region setup for Northern Ireland uses the wrong ISO country code.
### Describe the issue

In the Contoso Coffee Demo Dataset app, the demo country/region setup for Northern Ireland uses the wrong ISO country code.

Current line (linking to the repo for reference):
https://github.com/microsoft/ALAppExtensions/blob/a37faaa46ad5bc77c939ab70394237505129c88c/Apps/W1/ContosoCoffeeDemoDataset/app/DemoData/Foundation/1.Setup%20data/CreateCountryRegion.Codeunit.al#L63

The ISO code parameter for Northern Ireland currently resolves to GB. (written as GBN, but the Code field is only 2 chars long) 
For VAT / EORI / Intrastat / EU-goods trade, Northern Ireland must use the ISO 3166-1 alpha-2 temporary code XI, not GB.

Using GB for Northern Ireland:
breaks the uniqueness of ISO codes if an app expects one ISO code per country/region, and
is inconsistent with how Northern Ireland is treated in customs, VAT and EU goods trade after Brexit.

Authoritative guidance:
UK government guidance: to trade goods between Northern Ireland and the EU, businesses need a VAT registration number starting with XI. 
[GOV.UK](https://www.gov.uk/register-for-vat/selling-or-moving-goods-between-northern-ireland-and-the-eu?utm_source=chatgpt.com)

ISO 3166-1 alpha-2: the code XI is used by the UK government and the EU as the country code prefix for Northern Ireland in EORI and EU VAT reporting. 
[Wikipedia](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2?utm_source=chatgpt.com)

VAT references: XI is described as the “ISO 3166-1 alpha-2 temporary country code” for Northern Ireland VAT numbers used in EU trade in goods. 
[verifyvat.com](https://verifyvat.com/vat-formats/gb-northern-ireland-uk?utm_source=chatgpt.com)

### Expected behavior

The Northern Ireland demo country/region should use XI as ISO code (or the corresponding label/lookup that resolves to XI) so that demo data matches the current ISO- and VAT-related practice for NI.

### Steps to reproduce

1. Create a new demo company in Business Central using the Contoso Coffee Demo Data app (Contoso Coffee Demo Dataset).
2. Let the initialization run so that demo data is created, including countries/regions
3. Open the Country/Region table in the new demo company.
4. Locate the entry for Northern Ireland (NothernIrelandLbl in the code).
5. Observe that the ISO Code (and related token) corresponds to GB instead of XI.

Result:
Northern Ireland demo data is created with ISO/VAT code GB.

### Additional context

We found this while implementing and validating uniqueness of ISO codes for countries/regions in our own Business Central app. Northern Ireland and the United Kingdom currently end up sharing the same ISO code (GB) in the demo data, which breaks assumptions about uniqueness and leads to incorrect behaviour in ISO-based logic (especially VAT/EU reporting logic).

SchulzOli:
#### Summary 
This pull request updates the demo data setup for country and region creation, specifically correcting the label and token for Northern Ireland. The changes ensure proper naming and coding conventions for country/region identifiers and labels.

**Corrections for Northern Ireland:**

* Renamed the label from `NothernIrelandLbl` to `NorthernIrelandLbl` to fix a spelling error.
* Replaced the country token from `GBNTok` (GBN) to `XITok` (XI) for Northern Ireland, aligning with standard country/region codes.
* Updated the country/region insertion logic to use the corrected label and token for Northern Ireland.
#### Work Item(s) 
Fixes #29496

Fixes [AB#621128](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/621128)

### DIFF
--- a/Apps/W1/ContosoCoffeeDemoDataset/app/DemoData/Foundation/1.Setup data/CreateCountryRegion.Codeunit.al
+++ b/Apps/W1/ContosoCoffeeDemoDataset/app/DemoData/Foundation/1.Setup data/CreateCountryRegion.Codeunit.al
@@ -60,7 +60,7 @@ codeunit 5205 "Create Country/Region"
         ContosoCountryOrRegion.InsertCountryOrRegion(MY(), MalaysiaLbl, '458', '', '', Enum::"Country/Region Address Format"::"Post Code+City", 1, '', '');
         ContosoCountryOrRegion.InsertCountryOrRegion(MZ(), MozambiqueLbl, '508', '', '', Enum::"Country/Region Address Format"::"Post Code+City", 1, '', '');
         ContosoCountryOrRegion.InsertCountryOrRegion(NG(), NigeriaLbl, '566', '', '', Enum::"Country/Region Address Format"::"City+Post Code", 1, '', '');
-        ContosoCountryOrRegion.InsertCountryOrRegion(NI(), NothernIrelandLbl, CopyStr(GB(), 1, 2), '826', GBNTok, GBNTok, Enum::"Country/Region Address Format"::"City+County+Post Code", 1, '9932', '');
+        ContosoCountryOrRegion.InsertCountryOrRegion(NI(), NorthernIrelandLbl, CopyStr(GB(), 1, 2), '826', XITok, XITok, Enum::"Country/Region Address Format"::"City+County+Post Code", 1, '9932', '');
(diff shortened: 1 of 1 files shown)


JesperSchulz:

Build ID: 45992.

PR #29820, opened 2026-03-10
[Bug]: Infinite retry loop in Audit File Export parallel processing

### LINKED ISSUE #29797: [Bug]: Infinite retry loop in Audit File Export parallel processing
### Describe the issue

`Audit Line Export Runner` resets `"No. Of Attempts"` to 3 on every execution (line 22). When a task fails, `Audit File Export Error Handl.`  decrements the counter by 1 and unconditionally calls `StartExportLinesNotStartedYet`, which reschedules the task. Since the runner resets the counter back to 3 on the next run, the retry counter never reaches 0, creating an infinite loop of failing tasks.

### Expected behavior

The task should fail permanently after a fixed number of retry attempts

### Steps to reproduce

1. Set up an Audit File Export with `"Parallel Processing"` enabled
2. Start the export where at least one line will consistently fail
3. Observe that the failing task is retried indefinitely instead of stopping after the intended number of attempts

### Additional context

I suggest the initial attempt count only be set when the task is first scheduled in `RunGenerateAuditFileOnSingleLine` (Codeunit "Audit File Export Mgt."), not in the runner itself. 
And remove the line from `Audit Line Export Runner`.

vlkov:
#### Summary 
Set initial attempt count only be set when the task is first scheduled in RunGenerateAuditFileOnSingleLine and remove from runner.
#### Work Item(s) 
Fixes #29797 

Fixes [AB#625904](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/625904)

### DIFF
--- a/Apps/W1/AuditFileExport/app/src/ExportEngine/AuditFileExportMgt.Codeunit.al
+++ b/Apps/W1/AuditFileExport/app/src/ExportEngine/AuditFileExportMgt.Codeunit.al
@@ -283,6 +283,7 @@ codeunit 5261 "Audit File Export Mgt."
         AuditFileExportLine.Validate(Status, AuditFileExportLine.Status::"In Progress");
         Clear(AuditFileExportLine."Audit File Content");
         AuditFileExportLine.Validate(Progress, 0);
+        AuditFileExportLine.Validate("No. Of Attempts", 3);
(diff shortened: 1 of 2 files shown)


JesperSchulz:

Build ID: 47826.

PR #29821, opened 2026-03-10
[Bug]: Two sessions can process the same Audit File Export Line simul…

### LINKED ISSUE #29798: [Bug]: Two sessions can process the same Audit File Export Line simultaneously
### Describe the issue

When parallel export processing is enabled, it is possible for two background sessions to execute `Audit Line Export Runner` on the same `Audit File Export Line` record simultaneously.   SAF-T export is an extremely resource-intensive operation — it reads and processes large volumes of financial data, generates XML.

In `Audit Line Export Runner` I suggest to add after acquiring the lock and before writing session info, check if another active session is already processing this record.
Additionally, I think that the `StartExportLinesNotStartedYet` should cancel the existing scheduled task before creating a new one for the same line.

### Expected behavior

Each `Audit File Export Line` should be processed by exactly one background session at a time. If a task is already running for a given line, no additional task should be scheduled or started for that same line until the current one completes or fails.

#### Scenario 1: Error handler reschedules while a new task is already running

`Audit File Export Error Handl.` unconditionally calls `StartExportLinesNotStartedYet` after a failure. Combined with the infinite retry issue (where the runner resets `"No. Of Attempts"` to 3 on every run), this creates overlapping executions:

1. **Task A** (Session 1) is running on Export Line X
2. Task A fails → Error Handler sets `Status := Failed`, calls `StartExportLinesNotStartedYet`
3. `StartExportLinesNotStartedYet` sees Line X with `Status = Failed` and `"No. Of Attempts" = 2` → schedules **Task B**
4. **Task B** (Session 2) starts executing on Line X
5. Meanwhile, another failing line's error handler also calls `StartExportLinesNotStartedYet`, which may find Line X again (if Task B hasn't committed its Session ID yet) and schedule **Task C**
6. Now **Task B** and **Task C** are both processing Line X concurrently

#### Scenario 2: Race window between task scheduling and session registration

There is a time window between when a task is scheduled via `TaskScheduler.CreateTask()` and when the runner writes its `Session ID` to the record via `Rec.Modify()` + `Commit()`:

1. `StartExportLinesNotStartedYet` is called from two different error handler sessions nearly simultaneously
2. Both read Line X with `Status = Failed` (set by a previous error handler)
3. Both see `"No. Of Attempts" <> 0` and `Status <> Completed`
4. For `Status = Failed` lines, the code does NOT check `IsExportSessionActive` — it goes directly to `RunThisLine := true`
5. Both call `RunGenerateAuditFileOnSingleLine`, each scheduling a separate `TaskScheduler.CreateTask()` for the same record
6. The second `CreateTask` overwrites the first `"Task ID"` on the record, but the first task is already scheduled in TaskScheduler and will also run
7. Two sessions now execute the runner on the same Export Line

### Additional context

In environments with multiple BC server instances, the problem becomes even harder to control. The `Active Session` table only reliably returns sessions for the **current** server instance. 

vlkov:
#### Summary 
Ensure that previously scheduled tasks are properly cancelled before scheduling new ones and that export runner codeunits avoid duplicate processing by checking for active sessions.

In `AuditLineExportRunner.Codeunit.al`, before proceeding, the code now checks if the record's session is active in the current server instance and exits early if so. This avoids duplicate processing by ensuring only one active session handles the export line.
#### Work Item(s) 
Fixes #29798 

Fixes [AB#625902](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/625902)

### DIFF
--- a/Apps/W1/AuditFileExport/app/src/ExportEngine/AuditFileExportMgt.Codeunit.al
+++ b/Apps/W1/AuditFileExport/app/src/ExportEngine/AuditFileExportMgt.Codeunit.al
@@ -289,11 +289,16 @@ codeunit 5261 "Audit File Export Mgt."
             OnBeforeScheduleTask(DoNotScheduleTask, TaskID);
             if DoNotScheduleTask then
                 AuditFileExportLine."Task ID" := TaskID
-            else
+            else begin
+                if not IsNullGuid(AuditFileExportLine."Task ID") then
+                    if TaskScheduler.TaskExists(AuditFileExportLine."Task ID") then
+                        TaskScheduler.CancelTask(AuditFileExportLine."Task ID");
+
                 AuditFileExportLine."Task ID" :=
                     TaskScheduler.CreateTask(
                         Codeunit::"Audit Line Export Runner", Codeunit::"Audit File Export Error Handl.", true, CompanyName(),
                         NotBefore, AuditFileExportLine.RecordId());
+            end;
             AuditFileExportLine.Modify(true);
             Commit();
             NoOfJobs += 1;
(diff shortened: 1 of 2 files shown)


JesperSchulz:

Build ID: 47819.

PR #29839, opened 2026-03-16
[E-Document Formats][PINT A-NZ] - Supports prefixed XML roots in PINT A-NZ import

GMatuleviciute:

**Implementation**

Enhances the PINT A-NZ E-document import process to correctly parse XML documents where the root element includes a namespace prefix (e.g., ``). Previously, the parser assumed a non-prefixed root, leading to failures with such documents.

The solution modifies the PINT A-NZ Import codeunit to:

- Dynamically capture the full XML root path, including any namespace prefixes.
- Utilize this dynamic root path for all subsequent XML node lookups, ensuring accurate data extraction regardless of the root element's prefix.

Adds dedicated test cases to validate the parsing of both standard and namespace-prefixed root elements, confirming the improved robustness of the import logic.

Fixes #


Groenbech96:
Moved https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV/pullrequest/247049

PR #29867, opened 2026-03-24
Syncing with version 29.0.48074.0

JesperSchulz:
Fixes [AB#560539](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/560539)


FabioCerlini:
@JesperSchulz Sorry in advance if this is not the correct place to ask, but: is there any chance the changes to the "Intrastat Report Exp. Ext. IT" codeunit contained in this PR will be backported to BC 27? Just today we noticed the behavior fixed with the aforementioned changes is causing issues for one of our customers, but no app update is available for BC version 27.5.

JesperSchulz:
> Intrastat Report Exp. Ext. IT

I just checked and it looks like the fix was backported to 28.x only.

PR #29954, opened 2026-04-21
Sales Order API: restore Completely Shipped after redistributing invoice discounts

### LINKED ISSUE #29071: [Bug]: SalesOrder "Completely Shipped" is always true in "Sales Order Entity Buffer"
### Describe the issue

in Sales Orders API page whenever you call retrieve a sales order using API, the value of Rec."Completely Shipped" is always set to true.

the issue start with GraphMgtSalesOrderBuffer.RedistributeInvoiceDiscounts(Rec); 
in trigger OnAfterGetRecord() 
of page 30028 "APIV2 - Sales Orders"

this leads to procedure ResetRecalculateInvoiceDisc(SalesHeader: Record "Sales Header") 
in codeunit 56 "Sales - Calc Discount By Type" 
where SalesLine.ModifyAll("Recalculate Invoice Disc.", false); is called

and then in codeunit 5496 "Graph Mgt - Sales Order Buffer"
local procedure OnAfterModifySalesLine(var Rec: Record "Sales Line"; var xRec: Record "Sales Line"; RunTrigger: Boolean)
 UpdateCompletelyShipped(Rec); is called,
here the Rec contains a filter on: Document Type: Order, Document No.: 108765, **Recalculate Invoice Disc.: Yes**

and in UpdateCompletelyShipped theres 
SearchSalesLine.CopyFilters(SalesLine);

so the filter of Recalculate Invoice Disc: true is being set, therefore SearchSalesLine.IsEmpty() would be true in most cases. so the 
`CompletelyShipped := SearchSalesLine.IsEmpty(); `will return True! while in fact the lines are not shipped, 

i believe this line should be removed, no need to copy the filters from the Salesline, as we are putting all the lines filter in the procedure
```
SearchSalesLine.SetRange("Document Type", SalesLine."Document Type"::Order);
        SearchSalesLine.SetRange("Document No.", SalesLine."Document No.");
        SearchSalesLine.SetFilter(Type, '<>%1', SalesLine.Type::" ");
        SearchSalesLine.SetRange("Location Code", SalesLine."Location Code");
        SearchSalesLine.SetRange("Completely Shipped", false);
```

### Expected behavior

the Rec."Completely Shipped" value should be based on the lines of the sales order "Completely Shipped" value

### Steps to reproduce

create a sales order and sales lines, 
call get salesOrder API, check "Completely Shipped" , it will always be true even if the lines are not shipped

### Additional context

_No response_

jeffreybulanadi:
## Summary

Fixes microsoft/BCApps#8716

## Root Cause

In OnAfterGetRecord of both APIV1 - Sales Orders (page 20000) and APIV2 - Sales Orders (page 30028), calling GraphMgtSalesOrderBuffer.RedistributeInvoiceDiscounts(Rec) has an unintended side effect on Rec.\\Completely Shipped\\`.

The call chain is:

1. RedistributeInvoiceDiscounts calls SalesCalcDiscountByType.ResetRecalculateInvoiceDisc(SalesHeader)
2. Which calls SalesLine.ModifyAll(\\Recalculate Invoice Disc.\\, false) with a filter on Recalculate Invoice Disc. = Yes
3. This fires OnAfterModifySalesLine on codeunit 5496, which calls UpdateCompletelyShipped(Rec) where Rec still carries the Recalculate Invoice Disc. = Yes filter
4. UpdateCompletelyShipped calls SearchSalesLine.CopyFilters(SalesLine), inheriting the spurious filter
5. SearchSalesLine.IsEmpty() returns 	rue because no lines have Recalculate Invoice Disc. = Yes after the ModifyAll
6. Rec.\\Completely Shipped\\ := SearchSalesLine.IsEmpty() is set to 	rue incorrectly

## Fix

Save the value of Rec.\\Completely Shipped\\` from the database before invoking RedistributeInvoiceDiscounts and restore it afterward. The invoice discount redistribution function has no business modifying shipping status; this ensures the API returns the actual shipment state of the sales lines.

## Files Changed

- Apps/W1/APIV2/app/src/pages/APIV2SalesOrders.Page.al
- Apps/W1/APIV1/app/src/pages/APIV1SalesOrders.Page.al

## How to Test

1. Create a Sales Order with one or more sales lines (do not ship any lines).
2. Call GET /api/v2.0/companies({id})/salesOrders({id}).
3. Verify ullyShipped is alse.
4. Before this fix, ullyShipped would incorrectly return 	rue.

Fixes [AB#632884](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/632884)

### DIFF
--- a/Apps/W1/APIV1/app/src/pages/APIV1SalesOrders.Page.al
+++ b/Apps/W1/APIV1/app/src/pages/APIV1SalesOrders.Page.al
@@ -451,10 +451,15 @@ page 20028 "APIV1 - Sales Orders"
     }
 
     trigger OnAfterGetRecord()
+    var
+        IsCompletelyShipped: Boolean;
     begin
         SetCalculatedFields();
-        if HasWritePermission then
+        if HasWritePermission then begin
+            IsCompletelyShipped := Rec."Completely Shipped";
             GraphMgtSalesOrderBuffer.RedistributeInvoiceDiscounts(Rec);
+            Rec."Completely Shipped" := IsCompletelyShipped;
+        end;
     end;
 
     trigger OnDeleteRecord(): Boolean
--- a/Apps/W1/APIV1/app/src/pages/APIV1SalesOrders.Page.al
+++ b/Apps/W1/APIV1/app/src/pages/APIV1SalesOrders.Page.al
@@ -451,14 +451,11 @@ page 20028 "APIV1 - Sales Orders"
     }
 
     trigger OnAfterGetRecord()
-    var
-        IsCompletelyShipped: Boolean;
     begin
         SetCalculatedFields();
         if HasWritePermission then begin
-            IsCompletelyShipped := Rec."Completely Shipped";
             GraphMgtSalesOrderBuffer.RedistributeInvoiceDiscounts(Rec);
-            Rec."Completely Shipped" := IsCompletelyShipped;
+            Rec.Find();
         end;
     end;
 
--- a/Apps/W1/APIV2/app/src/pages/APIV2SalesOrders.Page.al
+++ b/Apps/W1/APIV2/app/src/pages/APIV2SalesOrders.Page.al
@@ -613,10 +613,15 @@ page 30028 "APIV2 - Sales Orders"
     }
 
     trigger OnAfterGetRecord()
+    var
+        IsCompletelyShipped: Boolean;
     begin
         SetCalculatedFields();
-        if HasWritePermission then
+        if HasWritePermission then begin
+            IsCompletelyShipped := Rec."Completely Shipped";
             GraphMgtSalesOrderBuffer.RedistributeInvoiceDiscounts(Rec);
+            Rec."Completely Shipped" := IsCompletelyShipped;
+        end;
     end;
 
     trigger OnDeleteRecord(): Boolean
--- a/Apps/W1/APIV2/app/src/pages/APIV2SalesOrders.Page.al
+++ b/Apps/W1/APIV2/app/src/pages/APIV2SalesOrders.Page.al
@@ -613,14 +613,11 @@ page 30028 "APIV2 - Sales Orders"
     }
 
     trigger OnAfterGetRecord()
-    var
-        IsCompletelyShipped: Boolean;
     begin
         SetCalculatedFields();
         if HasWritePermission then begin
-            IsCompletelyShipped := Rec."Completely Shipped";
             GraphMgtSalesOrderBuffer.RedistributeInvoiceDiscounts(Rec);
-            Rec."Completely Shipped" := IsCompletelyShipped;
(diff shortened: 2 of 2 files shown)


nikolakukrika (dismissed):

Apps/W1/APIV1/app/src/pages/APIV1SalesOrders.Page.al
```diff
@@ -451,10 +451,15 @@ page 20028 "APIV1 - Sales Orders"
     }
 
     trigger OnAfterGetRecord()
+    var
+        IsCompletelyShipped: Boolean;
     begin
         SetCalculatedFields();
-        if HasWritePermission then
+        if HasWritePermission then begin
+            IsCompletelyShipped := Rec."Completely Shipped";
```
nikolakukrika: I'm worried about this change as future logic or even maybe current logic could update the Rec."Completely Shipped" and we would set the old value back. 
Second issue is which other fields are we missing?

Since this issue will happen rarely does it make sense introducing this fix?

nikolakukrika: You could also try to use Rec.Find() to read the record again. I'm not sure if this would break anything but it could load the other fields.

jeffreybulanadi (commented):
Updated the fix to use Rec.Find() after RedistributeInvoiceDiscounts, as suggested. This re-reads the entire Sales Header from the database, which addresses both concerns raised:

1. No risk of overwriting legitimate future updates - the database is the source of truth, and Rec.Find() always reflects the actual persisted state.
2. All fields are restored, not just Completely Shipped - any other header field that may have been corrupted in-memory by the call chain is also corrected.

SetCalculatedFields() sets only page-level Text variables (BillingPostalAddressJSONText, ShippingPostalAddressJSONText, CurrencyCodeTxt, PartialOrderShipping) - not Rec fields - so Rec.Find() does not undo that work.

JesperSchulz:
Thank you for your contribution and for taking the time to submit this pull request.

At this stage, we’re not able to accept the change. We’d like to revisit this proposal once the migration to [BCApps](https://github.com/microsoft/BCApps) is complete though. At that point, we’ll be in a better position to properly evaluate and validate changes like this.

We appreciate your contribution and encourage you to resubmit or revisit this after the transition has been finalized.

nikolakukrika:
Please open a BC idea for this and lets get votes. One of the issues here is that we are re-reading the record from the OnAfterGetRecord trigger which is not something we do and can lead to the issues down the line. If we are refactoring this code, it would be better to introduce a more stabile solution. The best would be if we could redistribute discounts always, or maybe from a different trigger/background job or a different system. Patching from the OnAfterGetRecord is a workaround.

PR #29955, opened 2026-04-21
APIV2: replace LoadFields with AddLoadFields

### LINKED ISSUE #29634: [Bug]: APIV2 LoadFields at OnAfterGetRecord leads to JIT Load Error in Telemetry
### Describe the issue

We see on the endpoint APIV2/SalesInvoices sometimes the JIT Load error for the fields "Currency Code" or "Amount". 
In comparison to the number of successfull requests they are less than a tenth of a percent.

After Looking into the page 30012 "APIV2 - Sales Invoices" I discovered that these fields are loaded with Rec.LoadFields in the trigger OnAfterGetRecord. Which I assume leads to this race condition I see in my customer's temeletry.

https://github.com/microsoft/ALAppExtensions/blob/c56d2785c2ee11c4c1ae35a9d67da08da757e509/Apps/W1/APIV2/app/src/pages/APIV2SalesInvoices.Page.al#L775

Wouldn't it be better to change this loading to reading the data immediately? So I propose to change the LoadField which is called from trigger OnAfterGetRecord to an AddLoadFields in the OnOpenPage trigger where it should be added to the visible fields and thus directly loaded. Since there is no condition why the OnAfterGetRecord is not executing the LoadField command I assume this would be better than loading the fields afterwards.

There are simular implementations other pages as well but with less fields.
page 30051 "APIV2 - Bank Accounts"
page 30038 "APIV2 - Sales Credit Memos"
page 30037 "APIV2 - Sales Quotes"
page 30042 "APIV2 - Purchase Invoices"
page 30083 "APIV2 - Purchase Credit Memos"

### Expected behavior

Eliminating the JIT Load error in race conditions.

### Steps to reproduce

Create a large number of sales invoice records. Write a routine to update these records with different currency codes and shuffle them in the records.

Read the data with the endpoint salesinvoices and hope to fall into the race condition gap.

### Additional context

It should even be possible just to add the AddLoadFields commend because the LoadFields will not JIT load fields which are already loaded.

jeffreybulanadi:
## Summary

APIV2 pages call Rec.LoadFields(...) inside SetCalculatedFields and LoadCurrencyInformation procedures, which are themselves called from OnAfterGetRecord. This causes a JIT (Just-In-Time) read from the database for every record fetched, resulting in telemetry errors under concurrent API load as reported in issue microsoft/BCApps#8709.

## Root Cause

Rec.LoadFields in OnAfterGetRecord forces a supplemental DB read per record. When many records are fetched concurrently the platform logs JIT load warnings and the performance degrades unnecessarily.

## Fix

Replace Rec.LoadFields with Rec.AddLoadFields placed in OnOpenPage. AddLoadFields declares which fields must be included in the initial record-set query for the entire page session. The fields are fetched once alongside the primary key columns, removing the per-record JIT reads entirely.

Changes per page:
- APIV2SalesInvoices - fields No., Currency Code, Amount Including VAT, Posted, Status
- APIV2SalesCreditMemos - fields Applies-to Doc. Type, Currency Code
- APIV2SalesQuotes - field Currency Code
- APIV2PurchaseInvoices - field Currency Code
- APIV2PurchaseCreditMemos - fields Applies-to Doc. Type, Currency Code
- APIV2BankAccounts - field Currency Code (new OnOpenPage trigger added)

There is no behavioral change. The same fields are loaded; only the loading strategy changes from JIT per-record to eager per-session.

Fixes microsoft/BCApps#8709

Fixes [AB#632881](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/632881)

### DIFF
--- a/Apps/W1/APIV2/app/src/pages/APIV2BankAccounts.Page.al
+++ b/Apps/W1/APIV2/app/src/pages/APIV2BankAccounts.Page.al
@@ -108,6 +108,11 @@ page 30051 "APIV2 - Bank Accounts"
         CurrencyCodeDoesNotMatchACurrencyErr: Label 'The "currencyCode" does not match to a Currency.', Comment = 'currencyCode is a field name and should not be translated.';
         CurrencyValuesDontMatchErr: Label 'The currency values do not match to a specific Currency.';
 
+    trigger OnOpenPage()
+    begin
+        Rec.AddLoadFields("Currency Code");
+    end;
+
     trigger OnAfterGetRecord()
     begin
         LoadCurrencyInformation();
@@ -127,7 +132,6 @@ page 30051 "APIV2 - Bank Accounts"
     var
         GraphMgtGeneralTools: Codeunit "Graph Mgt - General Tools";
     begin
-        Rec.LoadFields("Currency Code");
         CurrencyCodeTxt := GraphMgtGeneralTools.TranslateNAVCurrencyCodeToCurrencyCode(LCYCurrencyCode, Rec."Currency Code");
     end;
 }
--- a/Apps/W1/APIV2/app/src/pages/APIV2BankAccounts.Page.al
+++ b/Apps/W1/APIV2/app/src/pages/APIV2BankAccounts.Page.al
@@ -108,7 +108,7 @@ page 30051 "APIV2 - Bank Accounts"
         CurrencyCodeDoesNotMatchACurrencyErr: Label 'The "currencyCode" does not match to a Currency.', Comment = 'currencyCode is a field name and should not be translated.';
         CurrencyValuesDontMatchErr: Label 'The currency values do not match to a specific Currency.';
 
-    trigger OnOpenPage()
+    trigger OnInit()
     begin
         Rec.AddLoadFields("Currency Code");
     end;
--- a/Apps/W1/APIV2/app/src/pages/APIV2PurchaseCreditMemos.Page.al
+++ b/Apps/W1/APIV2/app/src/pages/APIV2PurchaseCreditMemos.Page.al
@@ -616,6 +616,7 @@ page 30083 "APIV2 - Purchase Credit Memos"
     begin
         CheckDataUpgrade();
         SetPermissionsFilters();
+        Rec.AddLoadFields("Applies-to Doc. Type", "Currency Code");
     end;
 
     var
@@ -670,7 +671,6 @@ page 30083 "APIV2 - Purchase Credit Memos"
 
     local procedure SetCalculatedFields()
     begin
-        Rec.LoadFields("Applies-to Doc. Type", "Currency Code");
         SetInvoiceId();
         CurrencyCodeTxt := GraphMgtGeneralTools.TranslateNAVCurrencyCodeToCurrencyCode(LCYCurrencyCode, Rec."Currency Code");
     end;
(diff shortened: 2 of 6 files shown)


Apps/W1/APIV2/app/src/pages/APIV2BankAccounts.Page.al
```diff
@@ -108,6 +108,11 @@ page 30051 "APIV2 - Bank Accounts"
         CurrencyCodeDoesNotMatchACurrencyErr: Label 'The "currencyCode" does not match to a Currency.', Comment = 'currencyCode is a field name and should not be translated.';
         CurrencyValuesDontMatchErr: Label 'The currency values do not match to a specific Currency.';
 
+    trigger OnOpenPage()
+    begin
+        Rec.AddLoadFields("Currency Code");
```
nikolakukrika, line 113: We had issues with this kind of approach in the past. API page does not behave 100% as the actual page. In some cases the platform would open and close the page. Many properties that would be set from the OnOpen page would be lost. I'm not sure if we still have this kind of behavior on e.g. delete, POST, get $batch and other operations that can be done.

nikolakukrika: We would need an extensive testing before we uptake this change to address the risk.

nikolakukrika (commented):
left comments

Apps/W1/APIV2/app/src/pages/APIV2BankAccounts.Page.al
jeffreybulanadi: Good point. The BankAccounts page is different from the others in this PR because OnOpenPage was newly added just for AddLoadFields - there was no pre-existing OnOpenPage. This means that for API operations that may not trigger OnOpenPage (POST via OnInsertRecord, PATCH via OnModifyRecord, batch requests), AddLoadFields would not fire and LoadCurrencyInformation would fall back to a JIT read.

Switched OnOpenPage to OnInit for the BankAccounts page. OnInit fires when the page object is instantiated, before any trigger or data operation runs, making it the correct placement for AddLoadFields when there is no pre-existing OnOpenPage.

For the other pages in this PR (SalesInvoices, SalesCreditMemos, SalesQuotes, PurchaseInvoices, PurchaseCreditMemos), OnOpenPage already existed for permission checks. Adding AddLoadFields alongside those existing calls carries the same risk profile as the permission checks themselves, which are already relied on in production.

Regarding extensive testing - agreed. This change can be tested with API requests for GET (list and single), POST, PATCH, and DELETE on the Bank Accounts endpoint to confirm Currency Code is always correctly returned.

JesperSchulz:
Thank you for your contribution and for taking the time to submit this pull request.

At this stage, we’re not able to accept the change. We’d like to revisit this proposal once the migration to [BCApps](https://github.com/microsoft/BCApps) is complete though. At that point, we’ll be in a better position to properly evaluate and validate changes like this.

We appreciate your contribution and encourage you to resubmit or revisit this after the transition has been finalized.

Apps/W1/APIV2/app/src/pages/APIV2BankAccounts.Page.al
nikolakukrika: We decided to skip this optimization for the time being to limit the risk.

PR #29959, opened 2026-04-21
NO SAF-T: skip currency export when currency matches LCY code

### LINKED ISSUE #26554: [Bug]: Norwegian SAF-T file repeats the same amount when Entry has a Currency Code
### Describe the issue

When exporting a SAF-T file with where multiple General Ledger Entries has the same Document No. and Posting Date and the first entry has a Source Type different from blank and the related entry - e.g. Bank Account Ledger Entry - and this entry has a Currency Code the converted amount is repeated on all following entries in the exported file in stead of the actual amount of the General Ledger Entry.

image

### Expected behavior

Should convert amounts on all entries or export without currency code/converted amounts if currency code is the LCY currency

### Steps to reproduce

Post multiple general ledger entries with the same Document No. and Posting No.
First entry shoud have a related Bank Account Ledger Entry with a Currency Code.

### Additional context

image

jeffreybulanadi:
## Summary

When exporting SAF-T (both v1.0 and v1.3), if a Bank Account Ledger Entry (or Customer/Vendor Ledger Entry) has a Currency Code that equals the company LCY code, the export was incorrectly writing CurrencyCode, CurrencyAmount, and ExchangeRate XML nodes for what is effectively a domestic currency entry.

The root cause is in GetFCYData: after checking for a blank Currency Code, the function did not check whether the currency equals the LCY code. A bank account explicitly configured with LCY code (e.g. NOK on a Norwegian company) would pass the blank-check and inject spurious currency XML on all G/L entries in the same transaction.

## Changes

- Added GeneralLedgerSetup.Get() in GetFCYData, called once after the Export Currency Information guard.
- Added an LCY code guard in each ledger-entry branch (Customer, Vendor, Bank Account): if the retrieved Currency Code equals GeneralLedgerSetup.LCY Code, exit immediately.
- The LCY guard is placed before CalcFields to avoid an unnecessary database round-trip when it triggers.
- Applied identically to both GenerateSAFTFile.Codeunit.al (codeunit 10673, SAF-T v1.0) and GenerateSAFT13File.Codeunit.al (codeunit 10692, SAF-T v1.3).

## Related Issue

Fixes #26554

Fixes [AB#632882](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/632882)

### DIFF
--- a/Apps/NO/NorwegianSAFT/app/src/Export Engine/GenerateSAFT13File.Codeunit.al
+++ b/Apps/NO/NorwegianSAFT/app/src/Export Engine/GenerateSAFT13File.Codeunit.al
@@ -1076,6 +1076,7 @@ codeunit 10692 "Generate SAF-T 1.3 File"
         CustLedgEntry: Record "Cust. Ledger Entry";
         VendLedgEntry: Record "Vendor Ledger Entry";
         BankAccLedgEntry: Record "Bank Account Ledger Entry";
+        GeneralLedgerSetup: Record "General Ledger Setup";
     begin
         CurrencyCode := '';
         ExchangeRate := 0;
@@ -1083,6 +1084,7 @@ codeunit 10692 "Generate SAF-T 1.3 File"
         EntryAmountLCY := 0;
         if not SAFTExportHeader."Export Currency Information" then
             exit;
+        GeneralLedgerSetup.Get();
 
         if GLEntry."Source Type" in [GLEntry."Source Type"::Customer, GLEntry."Source Type"::" "] then begin
             CustLedgEntry.SetRange("Transaction No.", GLEntry."Transaction No.");
@@ -1090,6 +1092,8 @@ codeunit 10692 "Generate SAF-T 1.3 File"
             if not CustLedgEntry.FindFirst() then
                 exit;
             if CustLedgEntry."Currency Code" = '' then
+                exit;
+            if CustLedgEntry."Currency Code" = GeneralLedgerSetup."LCY Code" then
                 exit;
             CustLedgEntry.CalcFields(Amount, "Amount (LCY)");
             if CustLedgEntry.Amount = 0 then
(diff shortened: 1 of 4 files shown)


JesperSchulz:
@jeffreybulanadi, could you add some tests? Then I'll process 😊

jeffreybulanadi:
Hi @JesperSchulz, added automated tests for the LCY currency code fix.

Six new test methods added across both SAF-T versions:

**SAFTXMLTests.Codeunit.al (v1.0):**
- VendLCYCurrencyCodeNotExportedInCurrencyBlock
- CustLCYCurrencyCodeNotExportedInCurrencyBlock
- BankLCYCurrencyCodeNotExportedInCurrencyBlock

**SAFTXMLTests13.Codeunit.al (v1.3):**
- VendLCYCurrencyCodeNotExportedInCurrencyBlock
- CustLCYCurrencyCodeNotExportedInCurrencyBlock
- BankLCYCurrencyCodeNotExportedInCurrencyBlock

Each test mocks a GL entry and the corresponding ledger entry with Currency Code set to the company LCY code (read from GeneralLedgerSetup after SetupSAFT sets it), runs the export with Export Currency Information enabled, and verifies the DebitAmount node contains only n1:Amount with no CurrencyCode, CurrencyAmount, or ExchangeRate child elements.

ventselartur:
@jeffreybulanadi, thanks for the pull request. We usually use a blank currency for LCY. The "LCY Code" is primarily used for other purpose. May i know what is the reason to use the NOK currency code explicitly for this bank account, but not the blank currency?

JesperSchulz:
The solution doesn't seem to be right. We can pick this up again after the repo migration to BCApps, if need be. We're not going to accept this PR here on ALAppExtensions at this point.

PR #30186, opened 2026-06-01
GP - Clean combine tables before populating them

jaymckinney-enavate:
The replication service is replicating at least one 'combine' table, when it should not be. The quick fix is to truncate all of the combine tables before attempting to populate them.


JesperSchulz:
@jaymckinney-enavate, https://github.com/microsoft/BCApps will be ready for contributions before end of month. Please reopen PR there when we're ready.

